PR #398, opened 2026-03-05
Optimize validation with lazy error messages and fix benchmark Gradle commands

theqp:
## Description
This PR introduces lazy evaluation for validation error messages in the `FunctionValidator` class to reduce overhead in non-error paths, and fixes the Gradle commands for running specific JMH benchmarks.

### Changes
- **Validation Optimization**: Modified `FunctionValidator.requireMapper` to use null checking for lazy message construction, avoiding string concatenation and reflection when validation passes.
- **Benchmark Fixes**: Updated benchmark documentation to use `-Pincludes` instead of `--includes` for Gradle property passing, and added property handling in `hkj-benchmarks/build.gradle.kts`.
- **Test Updates**: Adjusted test files to accommodate the new lazy message API.

### Performance Improvements
Based on JMH benchmark results comparing before and after the changes in `EitherBenchmark`:
- **Recovery operations**: Up to 380x speedup (e.g., `leftRecovery`: 3.890 → 1353.534 ops/us, `rightRecovery`: 3.635 → 1382.429 ops/us)
- **Fold operations**: Up to 30x speedup (e.g., `leftFold`: 3.853 → 115.972 ops/us, `rightFold`: 3.674 → 103.922 ops/us)

These improvements stem from avoiding unnecessary string and reflection operations in validation paths that succeed, significantly benefiting operations that frequently validate functions.

### DIFF
--- a/hkj-core/src/main/java/org/higherkindedj/hkt/util/validation/FunctionValidator.java
+++ b/hkj-core/src/main/java/org/higherkindedj/hkt/util/validation/FunctionValidator.java
@@ -10,8 +10,6 @@ import static org.higherkindedj.hkt.util.validation.Operation.TRAVERSE;
 
 import java.util.Objects;
 import java.util.function.Function;
-import java.util.function.Supplier;
-
 import org.higherkindedj.hkt.Applicative;
 import org.higherkindedj.hkt.Kind;
 import org.higherkindedj.hkt.Monoid;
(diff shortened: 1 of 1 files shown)


MagnusSmith:
looks good.  If you could do a rebase and run `gradle spotlessApply ` then i think we should be good to go.   Great to have your contribution - Many Thanks

PR #534, opened 2026-05-17
hkj-openrewrite: harden recipes, fix bugs, expand coverage and docs

### LINKED ISSUE #533: [FEAT] Harden and extend hkj-openrewrite migration recipes
Goal
Audit the hkj-openrewrite recipes for correctness and robustness, broaden their coverage, and keep them current with the library's deprecation cycle so users get reliable automated migrations.

Problem
The recipes have gaps that produce wrong or no migrations:

- AddArityBoundsToTypeParameters always emits TypeArity.Unary, generating bounds that don't compile for binary witnesses (Kind2/Bifunctor/Profunctor), and only inspects method signatures — missing fields, locals, class hierarchy, nested generics and wildcards.
- Detection relies on fragile rendered-string matching (a user type named Free or a static import can mis-fire or be missed); detect-only recipes mutate source with TODO comments.
- AddHandleErrorCase doesn't handle switch expressions (the form Free interpreters use).
- No recipes exist for APIs deprecated for removal in 0.5.0.
- The 0.3.0 migration guide is an orphaned root-level file; recipe test coverage is thin and there's no validation that recipes load from the packaged resources.

Proposed scope
- Fix arity correctness (emit Binary where required) and widen witness-usage detection.
- Replace string matching with type-attributed matching; use search-result markers instead of editing source.
- Handle switch expressions in the Free recipe.
- Add a 0.5.0 deprecation-migration recipe group.
- Consolidate docs into a module README, add @DocumentExample, and add a recipe-catalogue validation test.
- Expand the test suite to cover the above.

Out of scope
Converting the detect-only effect recipes into automated rewrites — their targets are user-specific generated code and can't be transformed safely.

MagnusSmith:
Audit and hardening of the hkj-openrewrite migration recipes:

- Fix AddArityBoundsToTypeParameters: emit TypeArity.Binary for Kind2/Bifunctor/Profunctor (was always Unary), detect witness use across fields, locals, class hierarchy, nested generics and wildcard bounds, and fix malformed `<Fextends  ...>` output and the intersection-bound case.
- AddHandleErrorCase now also handles switch expressions and matches case labels by whole word; the three detect-only effect recipes emit SearchResult markers instead of mutating source with TODO comments, and ConvertRawFreeToFreePath/DetectInjectBoilerplate use type-attributed MethodMatcher instead of rendered-string matching.
- New MigrateDeprecationsTo0_5_0 recipe group (StateTKind.narrowK -> narrow, KindValidator.narrowWithPattern -> narrowHolder).
- Remove the invalid, redundant AddWitnessArityImports recipe (it listed a visitor as a recipe, breaking the AddArityBounds composite's validation in every shipped release).
- Consolidate the orphaned root MIGRATION-0.3.0.md into a comprehensive hkj-openrewrite/README.md; refresh stale plugin versions; add @DocumentExample and a recipe-catalogue validation test; note the work in the 0.4.5 release history. Tests 9 -> 34.
 Fixes #533


MagnusSmith:
Addressed the latest review in `118c4ea`:

**Shadowing in `scanArity` (high) — fixed.** The inner scan now stops descending into a nested class/method that redeclares a type parameter with the same name, so an inner witness usage can no longer leak a `WitnessArity` bound onto an unrelated outer parameter. Locked in by a new regression test (`doesNotBindOuterTypeParameterShadowedByAnInnerWitness`).

**`toString()`/`indexOf` fallback in `AddHandleErrorCaseRecipe` (medium) — fixed.** The brittle rendered-text fallback is removed; label matching is now purely structured (`getCaseLabels()` / `getExpressions()`). If a label can't be read structurally we now return empty (no match) rather than scanning printed text. All tests, including the switch-expression and word-boundary cases, still pass.

**Double-spacing in the type bounds (medium ×2) — not applied, with evidence.** The suggested change (setting the bound prefix to `Space.EMPTY`) would re-introduce the exact `<Fextends  WitnessArity<…>>` malformation this PR fixes. For rewrite 8.75.5's `JavaPrinter`, the `TYPE_BOUNDS` container's `before` is the space *before* `extends` and the element prefix is the space *after* it (and after `&`); the printer does not insert its own. The current spacing is asserted exactly by passing tests:
- `addsBoundToMethodTypeParameterUsedWithKind` → `<F extends WitnessArity<TypeArity.Unary>>`
- `appendsWitnessArityAsIntersectionWhenBoundAlreadyExists` → `<F extends Comparable<F> & WitnessArity<TypeArity.Unary>>`

Both produce single spaces; applying the suggestion makes them fail. Leaving the code as-is.

Suite: 36 tests, 0 failures; Spotless clean. Branch is rebased onto latest `main` (`3847c6e`).

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #583, opened 2026-06-27
feat(eitherorboth): add EitherOrBoth (inclusive-or) core type with EitherOrBothPath integration

### LINKED ISSUE #551: [FEAT] Add `Ior` (inclusive-or / `These`) core type with `IorPath` integration
## The gap

HKJ models two failure shapes today, and both are *exclusive*:

- **`Either<E, A>`** — `Left(e)` **or** `Right(a)`. A result is a failure or a success, never both.
- **`Validated<E, A>`** — accumulates errors, but is still exclusive: `Invalid` carries errors and **no value**; `Valid` carries a value and **no errors**.

Neither models a genuinely common outcome: **a successful value that also carries accumulated, non-fatal problems** — "success with warnings", or "partial success". Config parsing that succeeds but reports deprecations; a document that renders but logs recoverable issues; an import that produces records *and* a list of rows it had to skip.

Today users hack around this with `Either<E, Pair<List<Warning>, A>>`, a bespoke result record, or by dropping warnings entirely. None of these compose with the railway combinators (`map`/`via`/`recover`), so the warning channel becomes manual plumbing.

`Ior<A, B>` (inclusive-or; also known as `These`) is the principled type: `Left(a)` | `Right(b)` | `Both(a, b)`. It's exactly a sealed interface over three records — extremely idiomatic Java 25 — and it slots naturally beside `Either` in the Effect Path family. It also composes cleanly with issue #549 : the warning channel is a `NonEmptyList` (a `Both` always has ≥1 warning).

A value-with-warnings result that flows through the same combinators as everything else, accumulating problems on the left while the value travels on the right:

```java
Ior<NonEmptyList<Warning>, Config> result =
    parseConfig(raw)                          // Right(cfg), or Both(warnings, cfg), or Left(fatal)
        .flatMap(cfg -> validateConfig(cfg)); // warnings from both stages accumulate
```

- **`flatMap` semantics**: `Left` short-circuits (fatal); `Right`/`Both` continue; warnings carried by `Both` accumulate down the chain (requires a `Semigroup` on the left — the same mechanism `Validated`/`ValidationPath` already use).
- **`map`/`bimap`/`leftMap`** — `map` transforms only the value (right); `leftMap`/`bimap` reach the warnings.
- **Total accessors** — `getRight(): Maybe<B>`, `getLeft(): Maybe<A>`, plus `isLeft`/`isRight`/`isBoth`; no throwing getters.
- **Pattern-matchable** — `switch` over `Left`/`Right`/`Both` for the imperative shell.
- **Conversions** — `toEither` (drop or fail on warnings), `toValidated`, `fromEither`, `Ior.both(w, a)` / `Ior.right(a)` / `Ior.left(e)`.
- **`IorPath<E, A>`** — a railway variant: a success track that can carry accumulated warnings, with `recover`/fold at the boundary, plus the `.focus()` optics bridge, so it's not core-only.

## Things to explore

- **Shape** — sealed `Ior<A, B>` = `Left<A>` | `Right<B>` | `Both<A, B>` records.
- **Type-class instances** — Functor (right-biased), Bifunctor, Monad, Foldable, Traverse. The accumulating `flatMap`/`ap` needs a `Semigroup<A>` for the `Both + Both` case — **mirror the existing `Validated.ap` / `ValidationPath` approach** rather than inventing a new mechanism.
- **`IorPath<E, A>`** — integrate into the Path/effect family; `Path.ior…` factories; `.focus()` bridge; fold/recover at the boundary.
- **Conversions / interop** — to/from `Either`, `Validated`, `Maybe`; HKT `Kind` plumbing + `instances`/`Witnesses` registration following existing conventions.
- **Tie-in with #549** — default the warning channel to `NonEmptyList` (`Both` is non-empty by construction) and reuse `NonEmptyList.semigroup()`.
- **Naming** — `Ior` (Cats) vs `These` (Haskell): pick one primary name, document the other as the known alias.
- **Serialization** — Jackson support for `hkj-spring`, since a three-case sealed type carrying warnings may cross HTTP boundaries; decide how `Both` is represented.
- **Semantics matrix** — specify and test `flatMap`/`ap` behaviour for every `Left`/`Right`/`Both` combination; this is the subtle part and must be unambiguous.
- **Scope** — ship `Ior` + `IorPath`; do not attempt to also generalize `Validated`.

## Ergonomic constraints

- **Right-biased by default** — success on the right, problems on the left, matching `Either`/`Validated`/Path convention, so `map`/`flatMap`/`via` feel identical to the rest of the railway.
- **Accumulation via the established `Semigroup`-on-the-left mechanism** — supplied exactly as `ValidationPath` does it; pairs naturally with #549 's `NonEmptyList`. Don't introduce a second, divergent way to accumulate.
- **Plain fluent usability without HKT** — sealed interface + records, fully usable and `switch`-matchable without touching `Kind`.
- **Total accessors only** — `getLeft`/`getRight` return `Maybe`/`Optional`; no throwing getters.
- **Unambiguous, documented `flatMap` contract** — the meaning of `Left` (fatal / short-circuit) vs `Both` (carry warnings, continue) must be crystal clear to users who arrive with `Either` intuitions.
- **Don't over-scope** — `Ior` is for value-plus-warnings; it is **not** a replacement for `Validated` (pure accumulation with no partial value) or `Either` (clean exclusive choice). Position it as complementary in the docs.
- **HKJ idioms** — records, immutability, null-safety (`Unit`/`Maybe`, never `null`); additive and dependency-neutral.

MagnusSmith:
## Summary

Adds **`EitherOrBoth<L, R>`**, an *inclusive*-or (known elsewhere as `Ior` / `These`): `Left` | `Right` | `Both`. It is the principled type for **"success that also carries accumulated, non-fatal warnings"** — config that parses but reports deprecations, an import that yields records *and* a skipped-rows list — which neither `Either` (exclusive) nor `Validated` (no partial value) can express.

Right-biased, with **total** `getLeft()`/`getRight()` accessors that return `Maybe` and never throw. **Purely additive.**

Closes #551.

## What's included

**Core (`hkj-core`)**
- Sealed `EitherOrBoth` + `Left`/`Right`/`Both` records; dual `Kind`/`Kind2` witnesses, `KindHelper`/`ConverterOps`; right-biased `Functor`/`Monad`/`Traverse` + `Bifunctor`.
- **Accumulating `flatMap`**: `Left` short-circuits; `Both` carries its warnings forward, combined via a `Semigroup<L>` (default `NonEmptyList.semigroup()`). The monad's `ap` is monad-consistent (short-circuits) and lawful for any associative semigroup — deliberately **not** `Validated`-style accumulation (that lives on the Path). Exposed as `Instances.eitherOrBoth(Semigroup)`.
- **`EitherOrBothPath`** (effect): implements `Chainable` + `Accumulating` + `Recoverable` with a baked-in `Semigroup`. `via`/`zipWith` short-circuit and accumulate `Both`; `zipWithAccum`/`andAlso` collect every warning; `recover`/`mapError`; `.focus()` bridges; conversions. Factories `Path.right`/`left`/`both` + NEL conveniences `rightNel`/`leftNel`/`bothNel`. Added to the sealed `permits` of the capability interfaces and registered in the `hkj-checker` `PathTypeRegistry`.

**Spring (`hkj-spring`)** — tag-based `{"kind": ...}` Jackson serializer/deserializer (contextual generic resolution); `EitherOrBothPathReturnValueHandler` (`Right` → 200; `Both` → 200 with warnings in an `X-Hkj-Warnings` header, never dropped; `Left` → status via `ErrorStatusCodeStrategy`), behind `hkj.web.either-or-both-path-enabled`.

**Test lib (`hkj-test`)** — `EitherOrBothAssert` (`isLeft`/`isRight`/`isBoth`, `hasLeft`/`hasRight`/`hasBoth`, `*Satisfying`) + `AssertContract` coverage.

**Docs & examples** — new `monads/either_or_both_monad.md` and `effect/path_either_or_both.md` pages (wired into `SUMMARY` + glossary/cheatsheet/supported-types/release-history, with Related-Types cross-links from `Either`/`Validated`/`Writer`); runnable `EitherOrBothExample` and `EitherOrBothPathExample`.

## Testing & quality
- Full `gradle build` green across all modules.
- `hkj-core` JaCoCo coverage remains **100%** on the new package and `EitherOrBothPath`.
- Canonical per-type-class test layout (Laws + contract + type-specific), a jqwik property test reproducing the design POC, and the path three-layer tests.
- A multi-agent expert-panel review returned **APPROVE_WITH_NITS** (0 blockers, 0 majors); all surfaced doc-style nits have been addressed.

## Notes
- Position is complementary, not a replacement: `Either` for clean exclusive choice, `Validated` for pure accumulation with no partial value, `EitherOrBoth` for a value *and* warnings.
- The package segment is `eitherorboth` (matching the `nonemptylist` precedent); `EitherOrBoth`/`EitherOrBothPath` are the public type names.


MagnusSmith:
Thanks for the review. Resolutions below (1 of 3 applied; the other two are declined with rationale).

**1. `EitherOrBothDeserializer` — `asString()` → `asText()` (critical): declined, false positive.**
This module is on **Jackson 3** (`tools.jackson.*`; the resolved dependency is `jackson-databind-3.0.4`), where `JsonNode.asText()` was **renamed to `asString()`**. `asString()` is the correct Jackson 3 API, the project compiles and the full build is green — a genuine compilation error couldn't pass CI. Switching to `asText()` would actually break the build here. No change.

**2. `EitherOrBothPath.toTryPath` — null-check the mapped exception (medium): declined, redundant.**
`Try.failure(Throwable)` already guards its argument via `Validation.coreType().requireError(...)`, which throws a clear `NullPointerException` and never constructs a `Failure(null)`, so there is no NPE-or-invalid-state risk. The sibling handlers `EitherPath.toTryPath` and `ValidationPath.toTryPath` deliberately rely on that single upstream guard and pass the result straight through; adding a second check only here would make `EitherOrBothPath` inconsistent with the family for no behavioural gain. No change.

**3. `EitherOrBothPath.zipWith3Accum` — use `Tuple2::new` (medium): applied.**
Good catch on the inconsistency — `zipWith3` already used the inferred `Tuple2::new`. Switched `zipWith3Accum` to match; it infers fine and the build/tests stay green. Fixed in c496f461f.

PR #585, opened 2026-07-01
feat(pathops): NonEmptyList overloads for the race / first-success combinators

### LINKED ISSUE #579: [FEAT] NonEmptyList overloads for the PathOps race / first-success combinators
## The gap

Five `PathOps` "race / first to succeed" combinators take a `List` that **must be non-empty** and pay
for the missing guarantee with a runtime check:

```java
public static <A> TryPath<A> firstSuccess(List<TryPath<A>> paths) {
  if (paths.isEmpty()) {
    throw new IllegalArgumentException("paths must not be empty");   // the type permitted []
  }
  ...
}
```

The combinators sharing this guard:

- `PathOps.firstSuccess(List<TryPath<A>>)`
- `PathOps.raceVTask(List<VTaskPath<A>>)`
- `PathOps.firstVTaskSuccess(List<VTaskPath<A>>)`
- `PathOps.firstCompletedSuccess(List<CompletableFuturePath<A>>)`
- `PathOps.raceIO(List<IOPath<A>>)`

When the caller already knows there is at least one path (a fixed set of fallbacks, a literal list),
the `List` parameter still admits `[]`, the precondition is a runtime `IllegalArgumentException` rather
than a compile error, and the signature gives no type-level signal that "at least one" is required.
[`NonEmptyList`](https://github.com/higher-kinded-j/higher-kinded-j/issues/549) can encode that
directly.

For the known-non-empty case, the call is total and the guard is gone:

```java
// Today
TryPath<A> winner = PathOps.firstSuccess(paths);   // throws if paths is empty

// Goal (additive overload)
TryPath<A> winner = PathOps.firstSuccess(NonEmptyList.of(primary, secondary, backup));
// a NonEmptyList cannot be empty, so there is no empty case to guard or throw
```

For a list whose size is only known at runtime, the existing `List` overload still applies, and a
caller who wants the empty case as a value rather than a throw uses the `Maybe` bridge:

```java
Maybe<TryPath<A>> winner = NonEmptyList.fromList(dynamicPaths).map(PathOps::firstSuccess);
```

## Proposed change

Add a `NonEmptyList`-accepting overload beside each of the five `List` methods:

```java
public static <A> TryPath<A> firstSuccess(NonEmptyList<TryPath<A>> paths);          // total, no guard
public static <A> VTaskPath<A> raceVTask(NonEmptyList<VTaskPath<A>> paths);
public static <A> VTaskPath<A> firstVTaskSuccess(NonEmptyList<VTaskPath<A>> paths);
public static <A> CompletableFuturePath<A> firstCompletedSuccess(NonEmptyList<CompletableFuturePath<A>> paths);
public static <A> IOPath<A> raceIO(NonEmptyList<IOPath<A>> paths);
```

- **Keep the `List` overloads, undeprecated.** They are the right API for dynamic-size inputs.
- The `NonEmptyList` overload is total: no `isEmpty()` guard, no `IllegalArgumentException`; `head()`
  is safe inside the implementation.
- Purely additive; no existing caller changes.

## Things to explore

- **Overload resolution.** `NonEmptyList` does **not** implement `java.util.List`, so
  `firstSuccess(aNonEmptyList)` and `firstSuccess(aList)` are unambiguous for any non-null argument;
  only a literal `null` would be ambiguous (a non-issue in practice). Confirm with a compile test.
- **Shared implementation.** Either have the `List` overload's post-guard body delegate to the
  `NonEmptyList` overload (`firstSuccess(NonEmptyList.fromList(paths).orElseThrow(...))`-style), or
  factor a private core that both call, so the race logic lives once.
- **Docs.** Point the `List` overload's javadoc at the typed form and the `NonEmptyList.fromList`
  bridge, so callers discover the total option without a deprecation warning.

## Scope

- **In:** `NonEmptyList` overloads for the five `PathOps` race / first-success combinators; `List`
  overloads retained.
- **Out:** `sequence*` / `traverse*` (an empty input is *valid* there and returns an empty result, so
  `List` is correct and `NonEmptyList` would wrongly forbid it); `VStream.chunk` and `SagaBuilder`
  (sibling ticket — they need an API-shape decision, not a parameter overload).

## Dependencies & release

- **Depends on #549 (`NonEmptyList`).**
- **Sibling:** `[FEAT] NonEmptyList for VStream chunks and Saga steps` (the breaking / new-method half).
- **Release: 0.4.x.** Purely additive and non-breaking, so it does not need to wait for the 0.5.0
  breaking-change batch.

MagnusSmith:
Closes #579.

## What

Adds a `NonEmptyList`-accepting overload beside each of the five `PathOps` race / first-success combinators, so the "at least one path" precondition is encoded in the type instead of a runtime `IllegalArgumentException`:

- `PathOps.firstSuccess`
- `PathOps.raceVTask`
- `PathOps.firstVTaskSuccess`
- `PathOps.firstCompletedSuccess`
- `PathOps.raceIO`

```java
// Today — throws if the list is empty
TryPath<A> winner = PathOps.firstSuccess(paths);

// New — total, no guard to trip
TryPath<A> winner = PathOps.firstSuccess(NonEmptyList.of(primary, List.of(secondary, backup)));

// Runtime-sized input: bridge the empty case to a value
Maybe<TryPath<A>> winner = NonEmptyList.fromList(dynamicPaths).map(PathOps::firstSuccess);
```

## Design

- Each `List` overload keeps its `requireNonNull` + `isEmpty` guard and delegates to a new `private …Unchecked` core that holds the original race logic. The `NonEmptyList` overload is **total** (no guard, no throw) and routes the same core via `toJavaList()`, so both forms are behaviourally identical by construction.
- `List` overloads are **retained, undeprecated** — they remain the right API for dynamic-size inputs. Javadoc on each `List` form points at the typed overload; `firstSuccess` documents the `NonEmptyList.fromList` bridge.
- `NonEmptyList` does not implement `java.util.List`, so the overloads are unambiguous for any non-null argument.

## Source-compat note (worth a glance on review)

A second overload makes a **bare `PathOps.x(null)`** argument ambiguous. This broke six existing null-guard tests at compile time; they are cast to `(List<…>) null` so they keep asserting the `List` overload's guard. Any downstream caller passing a bare `null` literal today would hit the same and need the same cast — the only non-additive edge of an otherwise purely-additive change.

## Tests & verification

- 17 new tests (`@Nested "NonEmptyList overloads"`): first-success, all-fail (last-failure/last-exception propagation), single-element fast paths, and per-overload null-guards.
- `gradle :hkj-core:check` green (spotless + full suite); `jacocoTestCoverageVerification` passes (100% gate holds — the private cores stay covered by the existing `List` tests plus the new `NonEmptyList` tests).

## Scope

- **In:** the five combinators above; `List` overloads retained.
- **Out:** `sequence*` / `traverse*` (empty input is valid there), and the `VStream.chunk` / `SagaBuilder` sibling ticket.


MagnusSmith:
Thanks for the suggestion. I looked at it across all five overloads and I'm going to leave them as-is; the same reasoning applies to each, so replying once.

The five overloads deliberately route through a shared private `*Unchecked` core so the `List` and `NonEmptyList` forms are guaranteed to behave identically. That parity is the point of the design, and this change would break it asymmetrically.

Three of the cores (`raceVTaskUnchecked`, `firstCompletedSuccessUnchecked`, `raceIOUnchecked`) already short-circuit a single element to `getFirst()`, so `head()` here would only skip a one-element `toJavaList()` allocation: no behavioural change, but no meaningful saving either on combinators that race virtual threads, futures, or `IO`.

The other two cores do not short-circuit, and there the change is not equivalent to the current behaviour:

- `firstSuccessUnchecked` calls `run().isSuccess()` on the sole element before returning it; `return head()` skips that call, so the `NonEmptyList` overload would diverge from the `List` one.
- `firstVTaskSuccessUnchecked` returns a fresh deferred `VTaskPath` whose failure branch rethrows a checked cause as `new RuntimeException(cause)`, whereas returning the original `head()` surfaces a checked failure as `VTaskExecutionException`. So `firstVTaskSuccess(NonEmptyList.single(pathThatThrowsChecked))` would throw a different exception type.

Since the allocation saved is negligible for these I/O and concurrency combinators, and short-circuiting two of the five is a behavioural change rather than an optimisation, I'd rather keep all five uniform and behaviourally identical to their `List` counterparts.

An independent five-lens review panel (correctness, type safety, immutability, ergonomics, architecture) looked at the same spot and classified the single-element copy as an over-defensive nit, not a behaviour-changing optimisation, so I'm keeping the uniform shared-core delegation.

Issue #582, opened 2026-06-27
[FEAT] Add Edits builder + Update<S> (sparse, accumulating multi-edit over optics)


MagnusSmith:
> **Revised 2026-07-05** — dependency status corrected (#581/#586/#551 have landed since this was filed), the `Update<S>` monoid split out to #591, auto-labelling tracked as #592 (no longer a blocker), the accumulation mechanism corrected, and the API renamed for Java ergonomics (`set`/`modify`/`…IfPresent`/`parseIfPresent` mirroring `FocusPath.set/modify` and `Optional.ifPresent`, rather than the optics-jargon `over` family).

## The gap

HKJ can edit one path at a time (`FocusPath.set/modify`, `Setter.modify`) and update two *coupled* fields atomically (`Lens.paired`). What it has no ergonomic form for is the everyday case: **apply N *independent* edits at different paths in one reusable operation** — and, for the partial-update/REST-`PATCH` shape, **skip the edits whose incoming value is absent**.

Today that means hand-threading: `var s1 = pathA.set(...); var s2 = pathB.modify(..., s1); …`, with a manual `if (v != null)` around each, and no way to (a) reuse the patch across sources, (b) report *all* bad fields at once, or (c) attach a path to each failure.

This is the in-place sibling of a DTO mapper's construction: where a mapper *builds* a domain value from a DTO, a patch *updates* an existing aggregate from a sparse DTO. It rides the accumulation machinery that landed with #581 and is, in fact, the general form of a mapper's `patch`.

## What "good" looks like (user perspective)

A reusable, composable patch — and a sparse, all-errors-at-once validated PATCH:

```java
import static org.higherkindedj.optics.edit.Edit.*;

// Pure: reusable across sources; composes (Update<S> is a Monoid — #591).
Update<Order> normalise = Edits.combine(modify(EMAIL, String::toLowerCase),
                                        modify(SKU,   String::trim));
Order a = normalise.apply(orderA);
Order b = normalise.andThen(applyDiscount).apply(orderB);
```

```java
// Sparse PATCH controller: nullable DTO fields land 1:1; every bad field reported at once, located.
Validated<NonEmptyList<FieldError>, Order> updated =
    Edits.accumulate(
        setIfPresent(ORDER_NUMBER, req.orderNumber()),                     // null -> no-op
        parseIfPresent(EMAIL, req.email(), Email::parse).at("email"),      // String -> Validated, located
[...]

PR #593, opened 2026-07-05
Add Update<S> and Monoids.update(), the function-composition monoid

### LINKED ISSUE #591: [FEAT] Add Update<S> — composable update function with Monoids.update()
## The gap

HKJ has no function-composition monoid: there is no first-class "an update you can name, apply, reuse, and compose" value, and no `Monoid` instance to fold a list of updates through. Every `Monoids` factory today covers **data** (`list()`, `set()`, `string()`, the numeric families, `Optional`) — none covers **behaviour**.

The classical FP name for this is `Endo` (endomorphism). Following the same ergonomic call that renamed `Ior` to `EitherOrBoth` (#551), it is surfaced under a name every Java developer already understands: **`Update<S>`**.

```java
Update<Order> normalise     = order -> ...;   // any S -> S
Update<Order> applyDiscount = order -> ...;

// Named, reusable, composable:
Order patched = normalise.andThen(applyDiscount).apply(order);

// Fold any number of updates through the monoid:
Monoid<Update<Order>> m = Monoids.update();
Update<Order> all = updates.stream().reduce(m.empty(), m::combine);
```

- **`Update<S> extends UnaryOperator<S>`** — drops into `Stream.map`, `Optional.map`, and every existing function-shaped API for free.
- **`Update.identity()`** — the do-nothing update; the monoid identity.
- **`andThen(Update<S>)`** — covariant overload returning `Update<S>`, so composition stays in the type.
- **`Monoids.update()`** — `empty = identity()`, `combine = andThen` (left-to-right application), joining the existing noun factories (`Monoids.list()/set()/string()/…`).

## Design notes

- **Placement:** `hkj-api`, beside `Monoid`/`Monoids`. Zero new dependencies; no `Kind`/HKT surface — `Update` is deliberately a plain `@FunctionalInterface`.
- **Order pinned:** `combine(f, g)` applies `f` first, then `g`. Function composition is non-commutative, so the tests must pin application order **observably** (a non-commutative example), not just verify the monoid laws.
- **Naming:** `Update`, not `Endo` — per the `EitherOrBoth` precedent of preferring plain names over FP jargon. The javadoc mentions the classical name for searchability.

## Scope

- **In:** `Update<S>` (`identity()`, covariant `andThen`), `Monoids.update()`, monoid law tests plus an order-pinning test.
- **Out:** the `Edits` builder that consumes this (#582); any HKT integration.

## Relationship to other issues

- **Unblocks #582** (`Edits` builder / sparse PATCH) — the pure patch is a fold of per-field writes into `Monoids.update()`, and its `setIfPresent(absent)` no-op is exactly the monoid identity.
- Useful independently anywhere a pipeline of transformations is folded: normalisation chains, config layering, event replay.

MagnusSmith:
Closes #591. First domino for the `Edits` multi-edit builder (#582).

## What

- **`Update<S>`** (`hkj-api`, beside `Monoid`/`Monoids`): a named, composable update function. Extends `UnaryOperator<S>` so it drops into any `Function`-shaped API for free; `identity()` is the do-nothing update; `andThen(Update<S>)` is the covariant overload that keeps composition in the type. Javadoc names the classical `Endo` for searchability — the plain name follows the `Ior` → `EitherOrBoth` precedent.
- **`Monoids.update()`**: identity = `Update.identity()`, `combine(f, g)` applies `f` first, then `g` — left-to-right, matching the accumulation order pinned by the assembly builders. Javadoc warns that `Monoid.isEmpty` cannot detect a behaviourally do-nothing update (functions compare by reference).

## Tests

Monoid laws are observed **by application to probe values** (functions have no meaningful `equals`, so the entry can't join the equals-based `MonoidLawsTestFactory` table): identity laws, associativity, plus a non-commutative case pinning the left-to-right order, `combineAll` fold order, the `andThen` null guard, and `UnaryOperator` interop via `Stream.map`.

Deferred deliberately: a probe-based variant of `MonoidLawsTestFactory` — worth building when a second function-valued monoid arrives (likely with #582).


MagnusSmith:
Adopted in 1201940b0 — `andThen` now takes `UnaryOperator<S>`.

Verified by compile probe before adopting: the overload still wins resolution against the inherited generic `Function.andThen` (`UnaryOperator<S>` is more specific), `andThen(null)` remains unambiguous and hits the null guard, lambdas/method references/fluent chains all stay in the `Update` type, and non-endomorphic composition (`a.andThen(String::length)`) correctly still falls through to `Function.andThen`.

One residual documented in the javadoc: an argument statically typed as plain `Function<S, S>` (not `UnaryOperator`) still selects the inherited method — `fn::apply` stays in the type. Added a regression test composing with an unwrapped `UnaryOperator`. `gradle :hkj-api:check :hkj-core:check` green.

Issue #592, opened 2026-07-05
[FEAT] Optic-path labelling — auto-located validation errors from generated optics


MagnusSmith:
## The gap

Since #581 landed, accumulated validation errors can be **located**: `FieldError(List<String> path, String message)` exists, the assembly builders label leaves via `.field("email", value)`, and nesting prepends segments (`FieldError.at`), so paths compose outward-in (`"address.zip"`).

But every label is still **hand-written at the call site**. The information to do better already exists in the codebase, unconnected:

- **The codegen already knows every field's name** — `LensProcessor`/`FocusProcessor` read `component.getSimpleName()` and use it as the generated method's name. The label is in hand; it just isn't emitted as *data* on the generated optic.
- **Location already flows through optic composition** — `IndexedOptic<I, S, A>` threads a "where" into modification and composes it (`iandThen`). But it nests as `Pair<I,J>`, isn't a flat renderable path, and isn't wired to errors.
- **Value-level paths are anonymous** — a `FocusPath<Order, String>` obtained from `OrderFocus.email()` does not know it is `"email"`, so consumers like the `Edits` builder (#582) must label failures manually (`.at("email")`).

## What "good" looks like (user perspective)

Generated optics carry their own segment; composition concatenates; consumers label automatically:

```java
// Generated: OrderFocus.email() knows its own name.
FocusPath<Order, String> email = OrderFocus.email();
// composition builds the full path:
customer.via(address).via(zip)   // locates errors at "customer.address.zip"
```

```java
// #582's validated PATCH — no manual .at(...) needed for generated paths:
Edits.accumulate(
        parseIfPresent(OrderFocus.email(), req.email(), Email::parse))
    .apply(order);
// Invalid(NEL[ "email: not an address" ])
```

- **Auto-located when generated** — record-field segments come from the codegen (the name it already holds); hand-labelling (`.at(...)`, `.field(label, v)`) remains only for hand-written optics and third-party paths.
- **Leaf validators stay plain** — `Email.parse(String)` takes no path parameter; location is attached around it.
- **One rendering** — `FieldError.pathString()` (`"address.zip"`) stays the single wire format (it is what hkj-spring keys 422 bodies by).

## Things to explore

- **Emission.** `@GenerateFocus`/`@GenerateLenses` emit the component name as a segment on the generated paths/optics — the processors already read it; this is surfacing, not new analysis.
[...]

Issue #581, opened 2026-06-27
[FEAT] Add open-arity accumulating assembly (accumulate() / fields() builder) for Validated / ValidationPath / EitherOrBoth


MagnusSmith:
> **Updated 2026-07-02** — design review locked the two open decisions (representation → fields-first staged types; labelling → two entry points with a concrete `FieldError`), renamed `Ior` → `EitherOrBoth` to match the landed #551 naming, and corrected the description of the current `zipWithNAccum` family (it stops at arity 3). Second pass: the stage families are **processor-generated**, not hand-written (see Decision 1) — `For`'s stages already are — and per-record generated assembly is split out as fast-follow #586.

## The gap

HKJ accumulates validation errors today, but assembling a **record from N validated fields** — the everyday case (a request DTO → a domain aggregate; a raw config → a settings object) — has no ergonomic, open-arity form. What exists:

- **Pairwise** — `Validated.ap(Validated<E, Fn>, Semigroup<E>)` threads a `Semigroup<E>` on **every** call; `ValidationPath.zipWithAccum(...)` binds one at Path construction.
- **Fixed low arity** — `Applicative.map2 … map5`, capped at **five** fields and expressed in full `Kind` / HKT ceremony.
- **A family that stops at three** — `Accumulating.zipWithAccum` / `zipWith3Accum` (implemented by `ValidationPath` and `EitherOrBothPath`), each arity a separate `Tuple2`-nesting method. Every field past three invites yet another `zipWithNAccum`; the builder heads that proliferation off before it happens.

Three problems, all felt exactly when you reach for it:

- **It doesn't scale past a handful of fields.** `map5` caps at five; the accumulating zips cap at three.
- **It's ceremony-heavy.** Raw `Validated.ap` passes `Semigroups.list()` by hand, and the docs currently lead with `Instances.validated(Semigroups.list())` + `map3` in full `Kind` dress.
- **Errors can't say *where*.** None of these carries a per-field label, so an accumulated failure is a flat list with no `"address.zip"` path — consumers thread location manually.

This is the **assembly half** of HKJ's headline validation story, and it is the piece that a bidirectional mapper's `parse` and an accumulating partial-update both stand on. It also pairs directly with #549 and #551 (below).

## What "good" looks like (user perspective)

One flat, labelled, all-errors-at-once expression — no `Semigroup` argument, no arity wall, no `Kind`:

```java
// Labelled (fields()): every bad field reported at once, each with its path.
[...]

PR #594, opened 2026-07-05
Add Edits builder: sparse, accumulating multi-edit over optics

### LINKED ISSUE #582: [FEAT] Add Edits builder + Update<S> (sparse, accumulating multi-edit over optics)

MagnusSmith:
Closes #582. Builds on #591 (`Update<S>`/`Monoids.update()`) and #581 (`FieldError`, NEL channel). #592 (auto-labelling) remains an upgrade, not a blocker — this ships with explicit `.at(label)`.

## What

New exported package **`org.higherkindedj.optics.edit`**:

- **`Edit<S>`** (sealed, pure) — factories `set` / `modify` / `setIfPresent` / `modifyIfPresent` over both `FocusPath` and `Setter`. The `…IfPresent` family treats `null` as *absent*: the edit contributes the identity update, so sparse DTO fields land 1:1 with no `if` ceremony (a documented, deliberate exception to the eager null-guard convention — user functions are never invoked with null, and a sparse edit cannot "clear" a field).
- **`FallibleEdit<S>`** (sealed) — `Edit.parseIfPresent(path, raw, parser)` parses the incoming value first; failures are located `FieldError`s, and **`.at("email")`** prepends path segments exactly like `FieldError.at`. Absent raw values skip the parser entirely.
- **`Edits.combine(Edit<S>...)` → `Update<S>`** — folds pure edits via `Monoids.update()`. **Compile-time purity**: a `FallibleEdit` does not fit the signature, so validation failures can never be silently dropped (verified by compile probe in the design POC).
- **`Edits.accumulate(FallibleEdit<S>...)`** — validates every edit independently, reports **all** bad fields at once, in edit order, on the `NonEmptyList` channel. Same applicative semantics as the landed `accumulate()` assembly, but as a **homogeneous fold — no arity ceiling**. Terminal surface: `apply(S)` → `Validated<NEL<FieldError>, S>`, `applyPath(S)` → the `ValidationPath` twin, `toValidated()` → the folded `Update` for reuse across sources.

## Semantics (FINAL §6 / F7)

Two phases: ① each edit's incoming value is validated independently — validation never sees a source, which makes accumulation sound and patches reusable; ② writes run in a single left-to-right fold only if everything validated. Disjoint paths commute (tested); overlapping paths see earlier writes (pinned); coupled fields belong in one atomic `Lens.paired` edit. Deferred per the issue: Affine/Traversal edits, fail-fast `EitherPath` variant, overlap debug check (#592), tolerant `EitherOrBoth` twin.

## Verification

- `gradle :hkj-core:check :hkj-examples:test` — BUILD SUCCESSFUL (full suite, ArchUnit rules, spotless).
- **JaCoCo 100 %** on the new package (instructions/branches/lines/methods/classes all 0 missed).
- Tests pin: fold order, sparse no-ops, parser-not-invoked-on-absent (throwing parser), error accumulation order across mixed pure/fallible edits, multi-error relabelling, `applyPath().run()` round-trip, and every eager null guard.
- `EditsExample` driven end-to-end — sparse PATCH changes only present fields; two bad fields produce one `Invalid` listing both located errors.

## Review gate

Multi-angle review (line-by-line incl. an `ap` error-order audit against `Invalid.ap`, cross-file/arch tracer, cleanup, conventions); three confirmed findings fixed pre-PR: varargs null-element now reports the documented message (was `List.of`'s bare NPE), the ten near-identical FocusPath/Setter factory bodies collapsed into four shared private helpers, and the new book page is wired into the chapter's Previous/Next navigation (both neighbours updated).

## Docs

New book page **Multi-Edit and Sparse Updates** (optics → Integration and Recipes), cross-linked from the `Update` monoid section; release-history entry under 0.4.8-SNAPSHOT; `EditsExample` + EXAMPLES-GUIDE row; `module-info` export.


MagnusSmith:
Two follow-up commits on this PR:

- **ddd62012b** — adoption + doc fix: `IntegrationPatternsExample.validateCustomer` now uses `Edits.accumulate` with located errors (replacing manual `List<String>` collection); **cookbook Recipe 6 rewritten** — its previous `flatMap` chain short-circuited and discarded earlier writes, contradicting its own "accumulate all errors" problem statement; reverse cross-references added from `validated_assembly.md` and `coupled_fields.md`.
- **Tutorial 24: Multi-Edit and Sparse Updates** — the update-side twin of Tutorial 12, five exercises (pure folds, order pinning, sparse patches, validated PATCH), solutions green in the default suite, cross-linked from the book page and Tutorial 12.

PR #595, opened 2026-07-05
Add optic-path labelling: generated paths carry field-name segments

### LINKED ISSUE #592: [FEAT] Optic-path labelling — auto-located validation errors from generated optics

MagnusSmith:
Closes #592 — the last Phase-2 substrate item. Builds on #581 (`FieldError`) and #582 (`Edits`).

## What

- **Paths self-describe**: `FocusPath`/`AffinePath`/`TraversalPath` gain `segments()` (field-name segments, outermost first; empty = unlabelled) and `pathString()` (`"customer.address.zip"` — same rendering as `FieldError.pathString()`). The package-private impl records carry the segments; no public API break, and there is deliberately **no public type named `Path`** (Effect-Path collision).
- **Composition concatenates**: path-with-path `via`/`then` merges both sides' segments (all nine cross-type cases); raw-optic links and widening steps (`each()`/`some()`/`nullable()`/`at(i)`) contribute nothing; `filter` and `traced` preserve.
- **Codegen emits the name it already reads**: `@GenerateFocus` companions now produce `FocusPath.of(Lens.of(...), "email")` — one template change; widened methods label themselves because they compose from the labelled base.
- **`Edits` auto-locates**: `Edit.parseIfPresent(path, raw, parser)` folds the path's segments through `FieldError.at`, so a failure through a generated path reads `"email: not an address"` with **no explicit `.at(...)`**. An explicit `.at(label)` still prepends outward — auto-labels compose under it.

## Verification

- `:hkj-core:check` + `:hkj-processor:test` + `:hkj-examples:test` — BUILD SUCCESSFUL (ArchUnit, spotless, and the examples module regenerating its Focus companions against the new emission).
- **JaCoCo 100 %** on the `focus` and `edit` packages (a first pass caught two uncovered `pathString` defaults; now covered).
- Tests pin the 3×3 composition matrix, unlabelled-link/widening/traced/filter preservation, rendering, null guards, and the three Edits auto-label shapes (leaf, composed two-segment, explicit-`.at`-over-auto).
- Review gate: a deep correctness pass over the nine rewritten `via` defaults (vs the old delegation), record-equality impact, traced behaviour, the `locate` fold order, and the processor format string — zero findings.

## Deferred (per issue scope)

Sealed `PathSegment` model · index/key segments (`items[3]`) · labels on raw `Lens`/`Setter` · Navigator generator · **the `Edits` overlap debug check** (now feasible since paths have identity — natural follow-up) · hkj-spring 422 rendering (separate repo).

### DIFF
--- a/hkj-book/src/optics/multi_edit.md
+++ b/hkj-book/src/optics/multi_edit.md
@@ -82,7 +82,13 @@ Validated<NonEmptyList<FieldError>, Order> updated =
 // Invalid(NEL[ "email: not an address" ]) — or Valid(order') with only the present fields changed
 ```
 
-Errors accumulate in edit order on the `NonEmptyList` channel, exactly like the [accumulating assembly](../monads/validated_assembly.md) — but as a homogeneous fold, so there is **no arity ceiling**. For the railway, `applyPath(order)` is the `ValidationPath` twin of `apply(order)`, and `toValidated()` exposes the folded `Update` itself for reuse.
+Errors accumulate in edit order on the `NonEmptyList` channel, exactly like the [accumulating assembly](../monads/validated_assembly.md) — but as a homogeneous fold, so there is **no arity ceiling**.
+
+~~~admonish tip title="Generated paths label themselves"
+A path from a `@GenerateFocus` companion carries its record-component name as a **segment** (`OrderFocus.email()` → `"email"`); composing paths concatenates segments (`customer.via(address).via(zip)` → `"customer.address.zip"`, surfaced by `segments()`/`pathString()`), and `parseIfPresent` locates failures with them **automatically** — no `.at(...)` needed for generated paths. An explicit `.at(label)` still prepends outward, for hand-written optics or extra context.
+~~~
+
+For the railway, `applyPath(order)` is the `ValidationPath` twin of `apply(order)`, and `toValidated()` exposes the folded `Update` itself for reuse.
 
 ## Semantics: validate everything, then write once
 
--- a/hkj-book/src/release-history.md
+++ b/hkj-book/src/release-history.md
@@ -14,6 +14,10 @@ This page documents the evolution of Higher-Kinded-J from its initial release th
 
 ### 0.4.8-SNAPSHOT (Latest)
 
+**Optic-path labelling: generated paths know their own names**
+
+`@GenerateFocus` companions now emit the record-component name as a path segment (`FocusPath.of(lens, "email")`), and every path type surfaces `segments()` and `pathString()` (`"customer.address.zip"`). Composing paths concatenates segments; raw-optic and widening links (`each()`/`some()`/`nullable()`) contribute nothing. `Edit.parseIfPresent` locates parse failures with the path's own segments automatically, so sparse-PATCH errors arrive located with no `.at(...)` ceremony — an explicit `.at(label)` still prepends outward. Purely additive ([#592](https://github.com/higher-kinded-j/higher-kinded-j/issues/592)). See [Multi-Edit and Sparse Updates](optics/multi_edit.md).
+
 **`Edits`: sparse, accumulating multi-edit over optics**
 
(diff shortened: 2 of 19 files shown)


MagnusSmith:
Good catch — fixed in 8fa7451d6. The tip block's insertion had split the surrounding paragraph, leaving the "For the railway…" sentence on the closing-fence line; it's now a separate paragraph after the properly closed block. Verified the rest of the page's fences are paired.

Issue #597, opened 2026-07-05
[FEAT] Add ValidatedPrism — a Prism whose match accumulates reasons (parse-don't-validate boundaries)


MagnusSmith:
## The gap

HKJ's `Prism<S, A>` models "`S` *might* be an `A`" with `getOptional(S) : Optional<A>` (match) and `build(A) : S` (construct). The match answers **yes/no** — it cannot say **why not**, and it cannot report **more than one** problem.

That is exactly the wrong shape for a **validated boundary** — the "parse, don't validate" pattern where an unvalidated wire/DTO value is turned into an always-valid domain value:

- The forward direction (**parse**, wire → domain) is **fallible and wants reasons**: `"not an email"`, and ideally *all* the reasons at once, each located.
- The backward direction (**build**, domain → wire) is **total**: a valid domain value always renders.

A `Prism` collapses the fallible direction to `Optional` — it throws the reasons away. So today, parsing-with-reasons through the optic lattice is impossible: you either compose a `Prism` and lose the *why*, or you drop out of optics entirely into hand-written `Validated` code and lose composition with `Lens`/`Iso`/`Traversal` and the Effect-Path railway.

The missing piece is a **`Prism` whose match is `Validated` instead of `Optional`** — a *smart-constructor optic*: **`ValidatedPrism`**. `build` stays total; `parse` returns `Validated<NonEmptyList<FieldError>, A>` and **accumulates**. It is the optic that the assembly builder (#581), `NonEmptyList` (#549), and optic-path labelling (#592) were built to feed, and it is the lawful, composable home for a bidirectional mapper's `parse`.

## What "good" looks like (user perspective)

A leaf smart-constructor as an optic — fallible parse, total build — that composes like any other optic and lands on the railway:

```java
// build is total; parse fails with accumulated, located reasons.
ValidatedPrism<String, EmailAddress> email = ValidatedPrism.of(
    wire   -> EmailAddress.parse(wire),   // String -> Validated<NEL<FieldError>, EmailAddress>
    addr   -> addr.value());              // EmailAddress -> String   (total)

Validated<NonEmptyList<FieldError>, EmailAddress> parsed = email.parse("  NOPE ");
String rendered = email.build(addr);      // always succeeds
```

```java
// Composes with the lattice; sibling fields ACCUMULATE (via the assembly builder),
// nested fields SHORT-CIRCUIT (you can't parse the inner if the outer failed).
[...]

PR #598, opened 2026-07-05
Publish optic-law harness in hkj-test

### LINKED ISSUE #596: [FEAT] Publish an optic-law harness in hkj-test (law-test your own Lens/Prism/Iso/Affine/Traversal)
## The gap

HKJ tells users to write lawful optics — and gives them no way to check. The `Lens`/`Prism`/`Affine` law factories are **private JUnit classes in `hkj-core/src/test`**: unreachable from user code, unusable by an annotation processor, and invisible to `hkj-examples`. There is **no `IsoLaws` at all**, and nothing for `Traversal`. Anyone hand-writing a `Lens.of(...)` for a third-party type, an `OpticsSpec`, or a `Codec` is on their own for the very properties the documentation says define these optics.

The gap also blocks the roadmap: the `Refraction` optic must ship with **both** round-trip laws tested, and `@GenerateMapping`'s "law-checked" guarantee needs a published harness a processor can point generated tests at. Neither can be built against private test classes.

One dependency (`hkj-test`, already the home of `assertThatValidated` and friends), one import, laws as runnable tests:

```java
import static org.higherkindedj.optics.laws.OpticLaws.lensLaws;

@TestFactory
Stream<DynamicNode> nameLensIsLawful() {
    return lensLaws(UserLenses.name(), new User("Ada", 36), "Grace", "Alan");
}
// -> "get-set: setting what you got changes nothing"
// -> "set-get: you get back what you set"
// -> "set-set: the second set wins"
```

```java
// Or inline, inside an ordinary test:
OpticLaws.checkPrismLaws(Shapes.circle(), new Circle(1.0), new Square(2.0));
OpticLaws.checkIsoLaws(celsiusToFahrenheit(), 21.5, 70.7);
```

- **Every optic family covered** — `Iso` (round-trip both ways), `Lens` (get-set / set-get / set-set), `Prism` (build-match / match-build), `Affine` (the conditional variants), `Traversal` (identity + composition).
- **Failures name the law and the counterexample** — `"set-get law violated: set(Grace, User[Ada]) then get returned Ada"`, not a bare `expected true`.
- **Double-leveraged** — the same harness validates HKJ's own optics (the private hkj-core factories become thin delegates) and every downstream user's hand-written ones.
- **No HKT tax** — plain static methods over `Lens`/`Prism`/`Iso`/`Affine`/`Traversal`; `Kind` never surfaces.

## Things to explore

- **Entry point.** A single `OpticLaws` facade vs per-family classes (`LensLaws`, `PrismLaws`, …). Facade reads better in user code; per-family keeps javadoc focused. Possibly both (facade delegating).
- **Sample discipline.** Each law states exactly which samples it needs (a source plus two distinct focus values for set-set; a matching and a non-matching source for prisms). Guard rails: reject equal samples that would vacuously pass.
- **Equality by probe.** Laws compare via `equals` — right for records. Document the assumption; consider an optional `BiPredicate` equality overload for types without meaningful `equals` (the lesson from `Update`'s monoid tests).
- **Traversal laws.** Identity via `Id` applicative, composition/fusion where practical without dragging in `Kind` at the user surface.
- **`Refraction` readiness.** The harness should be extensible enough that the Refraction issue adds `parse`/`build` round-trip laws as one more family, not a new mechanism.
- **jqwik integration (later).** `Arbitrary`-driven law properties with shrinking, bounded by the repo's `tries = 50` convention — an additive follow-up, not MVP.
- **Migration.** Re-point hkj-core's private `LensLawsTestFactory`/`PrismLawsTestFactory`/`AffineLawsTestFactory` at the published harness so there is one source of law truth (and delete the duplicates).

## Ergonomic constraints

- **No HKT tax** — static methods over plain optics; JUnit-native (`DynamicNode`), no bespoke runner.
- **Deterministic and fast** — explicit samples, no randomness in the MVP; safe inside the existing test-speed budget.
- **Actionable failures** — law name + counterexample in every message.
- **Backward compatible** — purely additive to `hkj-test`; hkj-core's internal factories keep working until migrated.

- **No hard dependencies** — buildable today.
- **Underpins the `Refraction` optic** (companion draft) — its two round-trip laws (F10) should be written against this harness from day one.
- **Prerequisite for `@GenerateMapping`'s "law-checked" guarantee** — generated per-mapping law tests need a published target.
- **Continues the hkj-test double-leverage pattern** — like `assertThatX` and `NonEmptyListAssert` (#549), internal test needs become user-facing value.

## Scope

- **In:** law factories for `Iso`/`Lens`/`Prism`/`Affine`/`Traversal` in `hkj-test` (`DynamicNode` streams + inline `check…` forms); counterexample-bearing failure messages; migration of hkj-core's private optic-law factories onto the harness; book page (testing chapter) + example.
- **Out (defer):** jqwik/`Arbitrary` integration; `Refraction` laws (that issue adds its family); type-class law publication (`Monoid`/`Functor`/… — separate decision); `Getter`/`Fold`/`Setter` "laws" (too weak to be useful as laws); generated per-mapping law tests (the mapper feature).

MagnusSmith:
Closes #596 — first Phase-3 item; the published target for #597's `ValidatedPrism` laws and `@GenerateMapping`'s law-checked guarantee.

## What

New exported package **`org.higherkindedj.optics.laws`** in hkj-test: `IsoLaws`, `LensLaws`, `PrismLaws`, `AffineLaws`, `TraversalLaws`. Per-law `assert…` methods plus one `assertXxxLaws` aggregator per family; failures name the violated law with the offending values (`"Lens set-get: get(set(Grace, …)) == the value set; got Ada"`); guard rails reject vacuous fixtures. Affine includes the conditional laws (get-set/set-get/set-set on present targets, absence stability); Traversal covers identity + fusion.

## Surface adaptation vs the issue sketch

The issue sketched `DynamicNode` factories. As built, the harness follows the **existing `hkt.laws` convention** discovered in hkj-test (seven flat law classes, AssertJ-only, driven by `@ParameterizedTest`/property fixtures from the call site): hkj-test main has no JUnit dependency, and adding one for `DynamicNode` would churn the module graph for a wrapper the established style does without. A `DynamicNode` layer can be added later if demanded.

## Deferred

Migration of hkj-core's private `LensLawsTestFactory`/`PrismLawsTestFactory`/`AffineLawsTestFactory` onto this harness (follow-up — internal-only consolidation, ~1,200 lines of test restructuring best reviewed separately); jqwik arbitraries; `ValidatedPrism` laws (#597 adds its family).

## Verification

`:hkj-test:test` green — 6 tests covering every family's pass path, four unlawful-optic failure paths asserting law-named messages (set-ignoring lens, normalising iso, normalising prism, absent-target-writing affine), and both guard rails. Book: "Optic Laws" section on the Testing With hkj-test page; release-history entry under 0.4.8-SNAPSHOT.

### DIFF
--- a/hkj-book/src/release-history.md
+++ b/hkj-book/src/release-history.md
@@ -14,6 +14,10 @@ This page documents the evolution of Higher-Kinded-J from its initial release th
 
 ### 0.4.8-SNAPSHOT (Latest)
 
+**Published optic-law harness: law-test your own optics**
+
+`hkj-test` gains `org.higherkindedj.optics.laws` — `IsoLaws`, `LensLaws`, `PrismLaws`, `AffineLaws`, `TraversalLaws` — flat `assert…` helpers in the established `hkt.laws` style, so every user can verify the defining properties of hand-written optics (`Lens.of(...)`, spec interfaces, codecs). Failures name the violated law with the offending values; guard rails reject vacuous fixtures. Also the published target for the upcoming `ValidatedPrism` laws and `@GenerateMapping`'s law-checked guarantee. Purely additive ([#596](https://github.com/higher-kinded-j/higher-kinded-j/issues/596)). See [Testing With hkj-test](tooling/test_assertions.md).
+
 **Optic-path labelling: generated paths know their own names**
 
 `@GenerateFocus` companions now emit the record-component name as a path segment (`FocusPath.of(lens, "email")`), and every path type surfaces `segments()` and `pathString()` (`"customer.address.zip"`). Composing paths concatenates segments; raw-optic and widening links (`each()`/`some()`/`nullable()`) contribute nothing. `Edit.parseIfPresent` locates parse failures with the path's own segments automatically, so sparse-PATCH errors arrive located with no `.at(...)` ceremony — an explicit `.at(label)` still prepends outward. Purely additive ([#592](https://github.com/higher-kinded-j/higher-kinded-j/issues/592)). See [Multi-Edit and Sparse Updates](optics/multi_edit.md).
--- a/hkj-book/src/tooling/test_assertions.md
+++ b/hkj-book/src/tooling/test_assertions.md
@@ -225,9 +225,32 @@ Coverage is enforced at 100% line and 100% instruction on the `hkj-test` bundle.
 
 ---
 
+## Optic Laws
+
+`org.higherkindedj.optics.laws` publishes law-verification helpers for every optic family — the properties that *define* a lawful `Iso`, `Lens`, `Prism`, `Affine`, or `Traversal` — in the same flat `assert…` style as the `hkt.laws` type-class helpers:
+
+``` java
+import org.higherkindedj.optics.laws.LensLaws;
+import org.higherkindedj.optics.laws.PrismLaws;
+
+@Test
+void nameLensIsLawful() {
+    LensLaws.assertLensLaws(UserLenses.name(), new User("Ada", 36), "Grace", "Alan");
+    // get-set, set-get (both values) and set-set, with counterexample-bearing messages
+}
+
+@Test
+void statusPrismIsLawful() {
+    PrismLaws.assertPrismLaws(ShapePrisms.circle(), new Circle(1.0), new Square(2.0));
+}
+```
+
+Each family also exposes the individual laws (`assertGetSet`, `assertBuildMatch`, `assertSetNoOpWhenAbsent`, …) for targeted checks, and failures name the violated law with the offending values — `"Lens set-get: get(set(Grace, …)) == the value set; got Ada"`. Guard rails reject vacuous fixtures (equal set-set values, non-matching prism sources). Drive broader coverage with `@ParameterizedTest` or property fixtures at the call site.
+
 ~~~admonish tip title="See Also"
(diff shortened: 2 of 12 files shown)


MagnusSmith:
Second commit 7a2abd86b: the deferred migration is now done on this branch — the core law assertions in hkj-core's three private factories (Lens get-put/put-get/put-put, Prism review/non-match, Affine get-set/set-set/getOptional-set) delegate to org.higherkindedj.optics.laws, keeping their DynamicTest table-driven coverage. Full :hkj-core:check green.

MagnusSmith:
Confirmed and fixed — the aggregator's match-build check was indeed tautological (build-match guarantees getOptional(build(a)) == Some(a), so testing match-build on build(a) proves nothing). assertPrismLaws now takes a real matching source with a presence guard, deriving the focus value from it; the test passes "42" as the source. The individual assertMatchBuild was already non-vacuous (the zero-padded-prism failure test exercises it independently).

MagnusSmith:
All four applied in the latest commit — set-set presence guard, distinct-from-current-focus guards on the lens and affine aggregators, and the distinct-sources guard on the prism aggregator. hkj-test suite and the migrated hkj-core factories re-run green (the factory fixtures all satisfy the tightened guards).

Issue #600, opened 2026-07-05
[FEAT] Add @GenerateMapping — compile-time bidirectional record↔DTO mapper (lawful, effect-aware)


MagnusSmith:
## The gap

Entity↔DTO conversion is the most common boilerplate in mainstream Java, and HKJ has no first-class answer. `Iso` is a hand-built primitive and `@GenerateIsos` only *wraps* a user-supplied function pair — neither *derives* a mapping between two record shapes by matching fields, recursing into nested records, and lifting through `List`/`Optional`/`Map`. MapStruct owns this space but is one-directional-per-interface and not composable with optics/effects; telescope is bidirectional but reflective-by-default and not lawful (it can fabricate a total inverse for a lossy mapping).

HKJ can do better than both, because **a mapping already lives in the optic lattice**: derive the *weakest lawful optic the relationship admits* — `Iso` (lossless), `Lens` (lossy projection; `patch` falls out), or a `ValidatedPrism`-backed `parse` (validated boundary) — at **compile time**, **reflection-free**, **law-checked** (against the published harness, #596), and **effect-aware** (a fallible `parse` accumulates every field error, located).

The **classify** step genuinely reuses existing `hkj-processor` machinery (`TypeKindAnalyser`, `ContainerType`, the `ImportOpticsProcessor` pattern). Everything else is **net-new processor work, not "glue"**: the cross-record **emit** body (component-wise construction, container *recursion*, sealed→prism dispatch), the **`parse` assembly** (the `accumulate()` builder over per-field `ValidatedPrism`s), **exhaustiveness** enforcement (`Messager.ERROR` over a record-component set-difference), and **recursion/termination** for cyclic record graphs. The critical path is the processor + diagnostics — where annotation-processor projects routinely overrun.

## What "good" looks like (user perspective)

```java
// Lossless: empty body — same-named/typed fields auto-match, nested records recurse, containers lift.
@GenerateMapping
public interface UserMapping extends MappingSpec<User, UserDto> {}
// generates:  UserDto build(User);   User parse(UserDto);   Iso<User, UserDto> asIso();
//             + a round-trip law test on the published harness
```

```java
// Renames + a validated leaf. Renames are @MapField methods; a fallible leaf is a typed
// default method returning the boundary optic - no class-literal annotations, fully
// refactor-safe, and the leaf's type infects parse.
@GenerateMapping
public interface UserMapping extends MappingSpec<User, UserDto> {
[...]

PR #603, opened 2026-07-05
Add what/why/fix processor diagnostics and adopt in Focus/Lens/ImportOptics

### LINKED ISSUE #601: [FEAT] Processor diagnostics — what/why/fix messages across the optics annotation processors
## The gap
HKJ’s annotation processors report via terse one-liners — e.g. `FocusProcessor.java:147` *“The @GenerateFocus annotation can only be applied to records.”* — with no *why* and no *fix*. Telescope’s diagnostics are exemplary: they name the offending element, list what was found, and prescribe the exact remedy (including JPMS `opens`). The new mapper processors will emit many new diagnostics (exhaustiveness failure, same-name/different-type with no codec, non-invertible-declared-`Iso`, missing builder); these should meet a *what / why / fix* bar from day one — far cheaper than retrofitting.

- A small diagnostics helper (`Diagnostics.error(element, what, why, fix)`) shared across processors via `ProcessorUtils`.
- Bake into the **new** processors first; then audit `FocusProcessor`/`LensProcessor`/`ImportOpticsProcessor`.
- Coordinate with `hkj-checker` (exhaustiveness/raw-Kind rules) so build-time messages are consistent.

## Scope
**In:** a shared diagnostics format + adoption in the new processors + an audit pass on the existing ones. **Out:** a full message-catalogue/i18n.

---

MagnusSmith:
Closes #601 (A4) — the diagnostics standard, landed **before** any mapper processor exists so #600 meets the bar from its first line.

## What

- **`Diagnostics`** (`org.higherkindedj.optics.processing.util`): `error`/`warning`/`format` taking `(annotation, what, why, fix)` — the exact shape #601 specified: name the offending element, state what was found or needed, prescribe the remedy as an imperative sentence.
- **Audit pass on the three named processors**: `@GenerateFocus`/`@GenerateLenses` non-record errors upgrade from `"The @GenerateFocus annotation can only be applied to records."` to naming the element and its actual kind, the why (derivation from record components), and both remedies (move to a record, or `@ImportOptics` + `OpticsSpec` for types you cannot change). Filer `IOException`s gain context and a remedy. `@ImportOptics`' four sites reshaped into the shared format — their test-asserted substrings (`"has mutable fields"`, `"not a record, sealed interface, …"`) deliberately preserved.
- Remaining processors keep their current messages and migrate as touched, per the issue's scope.

## Verification

Full `:hkj-processor:check` green (all compile-testing suites, the coverage gate, arch rules). `DiagnosticsTest` covers `format`, both severities, and the eager guards using `reflect.Proxy` recorders — no new test dependencies. One compile-testing expectation tied to the old wording updated. Release-history entry with a worked example message.

## Note

`hkj-checker` message-consistency coordination is deliberately light here (its rules already follow an id-prefixed style); revisit when the mapper's exhaustiveness diagnostics land.

### DIFF
--- a/hkj-book/src/release-history.md
+++ b/hkj-book/src/release-history.md
@@ -14,6 +14,10 @@ This page documents the evolution of Higher-Kinded-J from its initial release th
 
 ### 0.4.8-SNAPSHOT (Latest)
 
+**Processor diagnostics: what / why / fix**
+
+Annotation-processor errors now follow a shared three-part format (`Diagnostics` in `hkj-processor`): **what** is wrong (naming the offending element and its kind), **why** (what the processor found or needs), and the exact **fix** — e.g. `@GenerateFocus: can only be applied to records, but 'Config' is a class. The processor derives FocusPath methods from record components. Move the annotation to a record, or use @ImportOptics with an OpticsSpec interface for types you cannot change.` Adopted across the `@GenerateFocus`, `@GenerateLenses`, and `@ImportOptics` error paths; the standard is in place for every future processor ([#601](https://github.com/higher-kinded-j/higher-kinded-j/issues/601)).
+
 **Codegen on-ramp: `-parameters` wired by both build plugins**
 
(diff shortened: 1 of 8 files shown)


MagnusSmith:
Applied — format now trims each part and appends a full stop unless the sentence already ends in ., ? or ! (empty parts pass through). Tests cover every branch; full :hkj-processor:check green.

PR #604, opened 2026-07-05
Wire -parameters automatically in the Gradle and Maven plugins

### LINKED ISSUE #602: [FEAT] Frictionless codegen setup — auto-wire the optics processor via the Gradle/Maven plugins + a sub-minute quickstart
## The gap
HKJ’s optics/Focus DSL is codegen-only — without the annotation processor wired, there is no `FocusPath`/`MappingSpec` to call. The mapper (A1) is the marquee entry point that draws new users; high-friction setup loses them at the first step (§ reflection/codegen axis — HKJ keeps codegen-first *by design*, so the answer is a frictionless on-ramp, not a reflective fallback).

## What “good” looks like
- `hkj-gradle-plugin` / `hkj-maven-plugin` (which already exist) **auto-wire** the annotation processor when applied — one line, no manual `annotationProcessor`/`-processor` configuration.
- A copy-pasteable quickstart that goes from zero to a working `FocusPath`/`MappingSpec` in under a minute.

- Plugin applies the processor dependency + `-parameters` (needed for some bean strategies) automatically.
- Quickstart in the book’s “where to start”; verify on a fresh project.
- Optional: a reflective `Focus.of(Class)` fallback **only** if onboarding data shows the build step is a real drop-off (keeps codegen the default).

## Scope
**In:** plugin auto-wiring + quickstart + a fresh-project smoke test. **Out:** a reflective core (explicitly not the default — see Part D of the telescope analysis).

MagnusSmith:
Closes #602 (A5) — run in parallel with #603.

## Honest scoping

Surveying the plugins showed most of A5 already shipped since the issue was drafted: the **Gradle plugin** registers `hkj-processor-plugins` on every source set's annotationProcessor configuration, the **Maven lifecycle participant** patches `annotationProcessorPaths`/`compilerArgs` at plugin level, execution level, and test overrides, and the book's top-level **Quickstart already leads with the one-line plugin setup**. This PR closes the residual gaps the issue names explicitly.

## What

- **Gradle**: `-parameters` added to every `JavaCompile` task (unconditional, idempotent) — parameter names in class files are read by copy strategies and the upcoming `@GenerateMapping` work (#600).
- **Maven**: `-parameters` added alongside `-Xplugin:HKJChecker` in every `compilerArgs` location the participant already patches, so it inherits the same plugin-level/execution-level/test-override coverage.
- **Docs**: quickstart and gradle-plugin pages now state the complete configured list; release-history entry.

## Verification

Both plugins' full `check` tasks green. New Gradle test asserts the flag on all compile tasks; Maven's exact-list `compilerArgs` assertions extended across the plugin-level, execution-override, and test-override cases.

### DIFF
--- a/hkj-book/src/quickstart.md
+++ b/hkj-book/src/quickstart.md
@@ -27,7 +27,7 @@ plugins {
 }
 ```
 
-This single line configures dependencies, preview features, annotation processors, and compile-time Path type checking automatically. See the [Gradle Plugin](tooling/gradle_plugin.md) documentation for the full DSL reference.
+This single line configures dependencies, preview features, annotation processors, `-parameters` (parameter names for copy strategies and the upcoming mapper), and compile-time Path type checking automatically. See the [Gradle Plugin](tooling/gradle_plugin.md) documentation for the full DSL reference.
 
 For **SNAPSHOT** versions of the plugin, add the Sonatype snapshots repository to your `settings.gradle.kts`:
 
--- a/hkj-book/src/release-history.md
+++ b/hkj-book/src/release-history.md
@@ -14,6 +14,10 @@ This page documents the evolution of Higher-Kinded-J from its initial release th
 
 ### 0.4.8-SNAPSHOT (Latest)
 
+**Codegen on-ramp: `-parameters` wired by both build plugins**
+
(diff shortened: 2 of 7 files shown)


MagnusSmith:
Confirmed and fixed — -parameters had ridden the checker-gated path, so disabling pathTypeMismatch dropped it (inconsistent with Gradle's unconditional flag). The compilerArgs patching now runs unconditionally with only -Xplugin:HKJChecker gated; the checks-disabled test asserts the new contract: -parameters present in both compilerArgs and testCompilerArgs, checker absent. Both plugins' check tasks green.

Issue #588, opened 2026-07-03
[CHORE] Repo-wide processor hygiene: latestSupported() source version and Filer originating elements in all 14 processors


MagnusSmith:
## Context

Gemini Code Assist flagged two annotation-processor hygiene issues on PR #587, fixed there for `AccumulatorProcessor`/`AccumulatorStepGenerator` (commit `aa62d8f06`). **This issue applies the same two fixes repo-wide: every annotation processor and every Filer/JavaPoet write site.** A survey confirms the patterns live only in `hkj-processor` (neither `hkj-processor-plugins` nor any other module declares supported versions or touches the `Filer` directly), so the module sweep below is the complete repo-wide fix. `ForComprehensionProcessor` is the first sibling, not the boundary.

## The two patterns

**1. Hardcoded `@SupportedSourceVersion(SourceVersion.RELEASE_25)`.**
Overriding `getSupportedSourceVersion()` to return `SourceVersion.latestSupported()` avoids compiler warnings when the processor runs on newer JDKs (26+). Affects **14 processors**:

- `ForComprehensionProcessor` (the headline; see PR #587 discussion)
- `LensProcessor`, `PrismProcessor`, `IsoProcessor`, `TraversalProcessor`, `GetterProcessor`, `SetterProcessor`, `FoldProcessor`
- `effect/PathSourceProcessor`, `effect/EffectAlgebraProcessor`, and the remaining `@SupportedSourceVersion` carriers (`grep -rl "@SupportedSourceVersion" hkj-processor/src/main/java` lists all 14)

**2. No originating element passed to the `Filer`.**
Without it, Gradle's incremental annotation processing cannot track which input produced each generated file, risking stale outputs or unnecessary full recompilations.

- String-template generators calling `createSourceFile(name)` with no element: `TupleGenerator`, `ForStepGenerator`, `ForPathStepGenerator`, `CoupledLensGenerator` (the trigger `package-info` element is available in `ForComprehensionProcessor.process` and just needs threading through, exactly as `AccumulatorStepGenerator.generate` now does).
- JavaPoet-based processors: there are **zero** uses of `TypeSpec.Builder.addOriginatingElement` in the module, so the Lens/Prism/Focus/... processors have the same gap via the JavaPoet route. Add `.addOriginatingElement(annotatedElement)` where the `TypeSpec` is built.

## Also in scope (same sibling-parity family)

[...]

PR #605, opened 2026-07-05
Add @GenerateMapping Step-0 slice: the de-risking cut of the record mapper

MagnusSmith:
The Step-0 slice from #600 (not `Closes` — the issue stays open for the full mapper; this is the go/no-go measurement).

## What

- **hkj-annotations**: `MappingSpec<A, B>` (beside `OpticsSpec`), `@GenerateMapping`, and `@MapField` (defined as API surface; Step 0 rejects it with a pointer to the full mapper).
- **hkj-processor `MappingProcessor`**: for `interface UserMapping extends MappingSpec<User, UserDto>`, generates `UserMappingImpl` (+ `INSTANCE`) with:
  - `UserDto build(User)` — total, component-wise, leaves via `email().build(...)`;
  - `Validated<NonEmptyList<FieldError>, User> parse(UserDto)` — `Validated.fields().field("name", …).field("email", email().parse(…))….apply(User::new)` — **located, accumulating errors free from the #581 assembly builder**, leaf boundaries as typed `ValidatedPrism` default methods per the ratified design.
- **Diagnostics from day one**: every error path uses #601's what/why/fix helper — including the exhaustiveness message that names the exact `default ValidatedPrism<…> email()` method to add (the very message #601 used as its exemplar).

## The risk measurement (what Step 0 exists for)

**Verdict so far: the processor cost is modest.** The slice — classifier, emit, exhaustiveness, diagnostics, tests — landed in one sitting with the compile-testing suite passing first run. The substrate did its job: `parse` is literally a `fields()` chain, leaves are literally `ValidatedPrism`s, and no new mechanism was invented. The known-expensive parts of the FULL mapper (container recursion, sealed dispatch, cyclic-graph termination, `Iso`/`Lens` tier selection) remain unmeasured — but the foundation walk suggests they are incremental, not architectural.

## Verification

`:hkj-processor:check` green (clean build). Compile-testing suite: the generated surface (ctor order, fields chain, labels, `apply(User::new)`) plus four diagnostics (non-interface, non-record arguments, `@MapField` rejection, missing-leaf exhaustiveness with the exact fix). Release-history entry.

## Not in this slice

Runtime round-trip example in hkj-examples and the hkj-spring 422 leg (separate repo), plus everything the issue defers to the full mapper.


MagnusSmith:
Slice 2 landed on this PR: @MapField renames (with declaration-only stubs in the Impl — the new test caught that abstract spec methods broke generated compilation), truthful asIso() on lossless mappings only, and GenerateMappingExample — the first RUNTIME execution of generated code, which immediately caught a nested-spec naming flaw (fixed via the OuterInnerAssembly convention). 7 processor tests + example verified end-to-end; :hkj-processor:check and :hkj-examples:test green. Two real defects found by exactly the verification steps the staging plan prescribed — the staged approach is earning its keep.

MagnusSmith:
**Slice 3 landed: container lifting (List + Optional; Map deferred)**

- `ValidatedPrism.parseAll(List<S>)` / `buildAll(List<A>)` added to hkj-core — element-wise accumulating parse that preserves order and collects **every** bad element, plus the total reverse direction. Fully covered by `ValidatedPrismTest`.
- `MappingProcessor` now classifies each correspondence as `IDENTITY | LEAF | LIST | OPTIONAL`:
  - `List<Wire> ↔ List<Domain>` lifts through the element leaf via `parseAll`/`buildAll`.
  - `Optional<Wire> ↔ Optional<Domain>` lifts inline: absent stays `Valid(Optional.empty())`, present parses through the leaf.
  - `Map` value lifting is **deferred** with a what/why/fix diagnostic (arrives with the full mapper).
- Container-lifted specs are correctly excluded from the lossless tier (no `asIso()`).
- Runtime proof in `GenerateMappingExample`: `TeamMapping` builds/round-trips a `Team(String, List<EmailAddress>, Optional<EmailAddress>)`, and a DTO with a bad member **and** a bad lead reports both failures at once: `Invalid(NonEmptyList[members: not an email address, lead: not an email address])`.

Verification: `:hkj-processor:check`, `:hkj-core:check` (100% coverage gate), `:hkj-examples:build` all green; example executed against the generated code.

Next: **Slice 4 — the CHECKPOINT** (nested-record recursion + cyclic-reference termination). Kill-switch criteria from the go/no-go apply: if slice 4's cost exceeds the sum of slices 1–3 or forces diagnostics/readability compromises, we stop and ship the flat mapper.

MagnusSmith:
**Slice 4 landed — and this was the CHECKPOINT. Verdict: PASSED, kill-switch NOT triggered.**

### What landed
- Every generated Impl now exposes `asValidatedPrism()`: the impl's `parse`/`build` pair is *exactly* a `ValidatedPrism<Wire, Domain>`, so a whole mapping plugs in wherever a leaf does.
- Nesting = **delegation, not inlining**: the processor keeps a per-round registry of valid specs; a component pair mapped by a sibling spec resolves to `SiblingImpl.INSTANCE.asValidatedPrism()` (an explicit leaf method still wins). Because generation never recurses, **cyclic termination holds by construction** — a self-recursive `Tree ↔ TreeDto` spec delegates to itself, compiles, and round-trips a depth-3 tree at runtime.
- `Correspondence` now carries a prism accessor *expression*, so nesting composes with slice-3 container lifting for free: `List<Customer> ↔ List<CustomerDto>` needs zero new emit logic.
- Emergent win: nested failures locate as **dotted paths** — `Invalid(NonEmptyList[customer.email: not an email address])` — from `Validated.fields()` label composition alone.
- Diagnostics: ambiguity (two specs for one pair) is rejected with a what/why/fix naming the disambiguating leaf method; the no-usable-source fix now also offers declaring a spec for the pair.

### Checkpoint assessment (against the go/no-go criteria)
| Criterion | Threshold | Actual |
|---|---|---|
| Cost vs slices 1–3 | must not exceed their sum | ~⅓ of it: one processor patch (+138 net lines, mostly the registry + resolver), **zero** new hkj-core runtime, 4 tests, single pass, no defects found in verification |
| Diagnostics quality | no compromises | improved: dotted-path location + a new ambiguity diagnostic |
| Generated-code readability | no compromises | unchanged shape; nested calls read as `CustomerMappingImpl.INSTANCE.asValidatedPrism().parse(wire.customer())` |

The feared complexity (recursive generation with cycle detection) never materialised because delegation makes it structurally impossible to need. **GO for Slice 5** (sealed dispatch + Iso/Lens tier selection).

Verification: `MappingProcessorTest` 13/13; `:hkj-processor:check` + `:hkj-examples:build` green; example executed — nested build/parse, located nested failure, and self-recursive tree round trip all verified against the generated code.

MagnusSmith:
**Slice 5 landed: sealed dispatch + the Lens emission tier. The staged Step-0 scope (slices 1–5) is now complete.**

### Sealed dispatch
- `MappingSpec<SealedDomain, SealedWire>` now generates `build`/`parse` as **exhaustive switches over the permitted subtype pairs**, each arm delegating to that pair's spec impl (resolved through the same round registry as nesting):
  ```java
  return switch (domain) {
    case Card v -> CardMappingImpl.INSTANCE.build(v);
    case Bank v -> BankMappingImpl.INSTANCE.build(v);
  };
  ```
- Exhaustiveness both ways: every domain subtype needs exactly one spec targeting a wire subtype, and every wire subtype must be produced by exactly one domain subtype — each violation gets its own what/why/fix.
- Sealed impls expose `asValidatedPrism()`, so a sealed mapping **nests inside records and lifts through containers** like any other spec — `Wallet(String, Payment)` ↔ `WalletDto(String, PaymentDto)` works with an empty spec.

### Lens tier (lossy projection)
- A wire record with **fewer** components maps as a projection: total `build` + a lawful `asLens()` whose `set` writes the projected components back and keeps the rest of the domain. **No `parse` is emitted** — dropped components can't be reconstructed (truthful types) — and projection specs are excluded from the nesting registry.
- Projection fields must identity-match (a leaf would make the write-back fallible); type changes and wider wires get dedicated diagnostics.
- Runtime proof: `Employee(name, department, age=36)` projects to a badge DTO; `asLens().set(badge with department=Platform)` returns the Employee with the new department **and age still 36**.

### Verification
`MappingProcessorTest` 20/20 (sealed dispatch, sealed-in-record nesting, missing-subtype-spec, uncovered-wire-subtype, projection happy path incl. rename, type-change, wider-wire); `:hkj-processor:check` + `:hkj-examples:build` green; the example executed all seven sections against generated code.

### Step-0 wrap-up
Emission tiers now read: **lossless → `asIso()`; lossy projection → `asLens()`; anything fallible → accumulating located `parse`** — the weakest-lawful-optic principle from the issue, delivered. Deferred to the full mapper per the issue: Map value lifting, Getter/Fold for derived wire fields, generated law tests on the #596 harness, and the hkj-spring `parse`→422 leg (separate repo).

MagnusSmith:
**Pre-PR expert-panel review complete — 5 lenses (correctness, type safety, immutability, ergonomics, HKJ architecture), 28 candidates, deduped and adversarially verified. 14 confirmed findings fixed in 6450d921a; 4 deferred with rationale; 1 refuted.**

### Fixed — correctness (each reproduced against the built processor before fixing)
| Finding | Before | After |
|---|---|---|
| Rename collides with a same-named component | uncaught `NoSuchElementException` crashes javac | what/why/fix naming both domain components |
| Two renames claim one wire component | second silently wins / crash | what/why/fix |
| Projection shares one domain source across two wire components | generated `asLens()` silently violated the get–set law | what/why/fix explaining the unlawful write-back |
| >12 components | non-compiling generated code (`Validated.fields()` ceiling) | diagnostic suggesting nested records (which nest via their own specs) |
| Generic specs/records | generated code referencing undeclared `T` | diagnostic |
| Abstract non-`@MapField` methods / parameterised `@MapField` | non-compiling Impl | diagnostics |

### Fixed — diagnostics, conventions, runtime, docs
- **Near-miss leaf**: a same-named default method with wrong `ValidatedPrism` arguments is now *named* in the error, with the wire-first argument order spelled out (the natural mistake, since `MappingSpec` is domain-first).
- **Projection specs** now register as non-parse-capable, so failed nesting/sealed dispatch names the spec and says why it cannot join instead of claiming no spec exists.
- **`FilerException`** gets a dedicated diagnostic for Impl-name collisions (`Outer.Inner` vs `OuterInner`) with a rename fix; **stale fix text** ('wait for the full mapper's renames') corrected.
- **Conventions**: generated Impls now carry `addOriginatingElement(spec)` (repo-wide Filer hygiene, #588/#589); `MappingProcessor` added to `ProcessorSourceVersionTest`; the three untested sealed/extends diagnostic branches now have tests.
- **`ValidatedPrism.parseAll`** rewritten from O(n²) copy-per-element to a single O(n) pass; errors asserted in list order; null elements in `parseAll`/`buildAll` fail fast naming the offending index.
- **Docs**: `@MapField`/`@GenerateMapping`/`MappingSpec` Javadoc rewritten to match what actually shipped (it still claimed renames 'arrive with the full mapper', and `MappingSpec` documented leaf naming by the *wire* component when the processor matches the *domain* component); Impl naming convention (nested name joining + `INSTANCE`) documented on the annotation; release-history entry updated.

### Deferred (documented, deliberate)
- Registry admits specs before classification succeeds → a failed nested spec cascades 'cannot find symbol' *after* the genuine diagnostic. Proper fix is a two-phase classify-then-emit pipeline — full mapper.
- JSpecify `@Nullable` erasure in identity matching; null-tolerant handling of null containers/leaf-returns inside records — nullness contracts are full-mapper scope; the wire argument itself is guarded.

### Refuted
- 'asIso()/asLens() lambdas lack null guards' — hand-written `Iso`/`Lens` have no null guards either; the generated optics match the existing optics convention.

Verification: `MappingProcessorTest` 34/34 (14 new), `ProcessorSourceVersionTest` green, `:hkj-processor:check` + `:hkj-core:check` (100% coverage gate) + `:hkj-annotations:check` + `:hkj-examples:build` all BUILD SUCCESSFUL.

MagnusSmith:
**Gemini review applied in 2a1c5e8a9** — two of the three findings were live; the third was already fixed.

1. **Identity match skipping a custom leaf (high) — adopted, with a scope decision.** `classify()` now checks for an explicit leaf method *before* the same-type identity match, so `ValidatedPrism<String, String>` can validate/normalise a component the types alone would copy verbatim. A same-type leaf makes the mapping fallible, so `asIso()` correctly disappears (truthful types). Deliberately narrower than the suggested reorder: only the **explicit leaf method** wins over identity — implicit *nested-spec* resolution does not, otherwise a stray `(X, X)` spec in the round would silently hijack every identity component of type X. Explicit-beats-implicit also means a whole-component leaf now wins over implicit container lifting. New test: `sameTypedLeafWinsOverIdentity`.

2. **`findLeaf` matching parameterised default methods (high) — adopted.** Only zero-parameter default methods match now; previously the generated no-arg call site failed to compile. The near-miss hint names the problem ("…exists but returns 'X' and declares parameters — a leaf must be a zero-parameter default method…"). New test: `parameterisedDefaultMethodIsNotALeaf`.

3. **Redundant `.build().toBuilder().build()` (medium) — already fixed.** The expert-panel hardening commit (6450d921a) refactored file writing into `writeFile()` with a single `build()`, so this no longer exists on the branch.

Javadoc updated (processor + `@GenerateMapping`) to document the leaf-over-identity override. Verification: `MappingProcessorTest` 36/36, `:hkj-processor:check` + `:hkj-examples:build` green.

MagnusSmith:
**Gemini round-3 applied in 425ff639d** — all three findings adopted.

1. **`parseAll` indexed iteration (medium)** — switched to an enhanced for with a counter, so a non-random-access `List` (e.g. `LinkedList`) stays O(n) instead of degrading to O(n²) through `get(i)`. The indexed null-element messages (`sources[1] must not be null`) are unchanged.
2. **`buildAll` indexed iteration (medium)** — same fix.
3. **`@MapField` on a non-abstract method (medium)** — confirmed live: a `default` method annotated `@MapField` was silently collected as a rename while remaining callable code, and slipped past the sealed-spec check (which only examined abstract methods). `validateSpecMethods` now rejects `@MapField` on any method with a body with a what/why/fix ('Remove the body, or remove the @MapField annotation'). New test: `mapFieldOnDefaultMethodRejected`.

Verification: `MappingProcessorTest` 37/37, `ValidatedPrismTest` green, `:hkj-processor:check` + `:hkj-core:check` (100% coverage gate) + `:hkj-examples:build` all BUILD SUCCESSFUL.

MagnusSmith:
**Gemini round-4 applied in e397d3467** — the finding is correct; the fix takes the Step-0 route.

The bug class was real: `getEnclosedElements()` misses inherited members, so a spec extending a base interface with an abstract `@MapField` method generated an Impl that failed to compile with a raw javac error (no stub), and an inherited leaf method was silently ignored (classification fell through to a misleading 'no matching leaf method was found').

Rather than switch to `getAllMembers()` — which needs generic-supertype substitution (`Types.asMemberOf`), filtering of the implicit `Object` members it returns for interfaces, and decisions about inherited-leaf precedence — the slice now **rejects spec inheritance with a what/why/fix**:

> `'InheritingMapping' extends interfaces besides MappingSpec.` Renames and leaves declared on other supertypes are invisible to this slice's classification, so the generated Impl could silently miss inherited members. Declare every rename and leaf directly on the spec; spec inheritance arrives with the full mapper.

This matches the PR's standing discipline (generics, >12 components, Map lifting): unsupported shapes are diagnosed, never silently miscompiled. Full inheritance support is noted for the full mapper alongside the other deferred items.

Documented on `@GenerateMapping` and in the book's limits section; covered by `specInheritanceRejected` — which the new 100% `MappingProcessor` coverage pin verifies automatically.

Verification: `MappingProcessorTest` 60/60, `:hkj-processor:check` (incl. the 100% class rule) + `:hkj-examples:build` green, `mdbook build` clean.

PR #608, opened 2026-07-06
Processor coverage hardening: 99% line / 97% branch module-wide, gate raised

MagnusSmith:
Applies the MappingProcessor coverage methodology (PR #605) across the whole of hkj-processor: measure per-branch, test what is reachable, delete what is provably dead, accept (and document) only the genuinely unreachable.

## Result
- **hkj-processor: ~99.7% line / ~98.3% branch** (was ~97.7%/92.4%), gate raised **95/87 → 99/97** with a comment naming the verified-unreachable residual; MappingProcessor stays pinned at 100/100.
- 77 missed lines / 137 missed branches across 17 classes → **13 lines / 32 branches across 8 classes**, all documented as unreachable (compiler-synthesised exhaustive-switch defaults, @Target-unproducible arms, fixed descriptor-table combos, guards dominated by earlier validation).

## How
- **Reuse fix doubling as coverage fix:** `ProcessorUtils.capitalise` (locale-neutral, unit-tested) replaces eight drifted private copies — also a determinism fix (the Lens/Getter/Setter copies used default-locale `toUpperCase()`, which generates different identifiers on e.g. tr_TR JVMs).
- **Proxy-driven unit tests** for paths javac cannot produce: MirroredTypeException guards (PathSource/ComposeEffects), ForComprehension's non-package/duplicate/unreadable guards, and writeFile-style filer fallbacks.
- **Analysis-verified dead-guard deletions** (~a dozen), each with the dominating invariant traced; enum switch statements converted to exhaustive expressions where JaCoCo filters the implicit default.
- **Targeted fixtures** for every reachable diagnostic branch across the external/import cluster, Focus/Navigator, and the effect processors, including additive test-scope SPI TraversableGenerators (marker-typed, inert for real types and golden files).

### DIFF
--- a/hkj-processor/build.gradle.kts
+++ b/hkj-processor/build.gradle.kts
@@ -76,12 +76,12 @@ tasks.jacocoTestCoverageVerification {
             limit {
                 counter = "LINE"
                 value = "COVEREDRATIO"
-                minimum = "0.95".toBigDecimal()
+                minimum = "0.99".toBigDecimal()
             }
             limit {
                 counter = "BRANCH"
                 value = "COVEREDRATIO"
-                minimum = "0.87".toBigDecimal()
+                minimum = "0.97".toBigDecimal()
             }
         }
         // MappingProcessor is held at 100% (see PR #605).
--- a/hkj-processor/build.gradle.kts
+++ b/hkj-processor/build.gradle.kts
@@ -72,6 +72,9 @@ tasks.jacocoTestCoverageVerification {
     )
 
     violationRules {
+        // 99/97 (not 100): the residual is verified-unreachable - compiler-synthesised
+        // exhaustive-switch defaults, @Target-unproducible arms, and descriptor-table
+        // combos that cannot occur (see the coverage-hardening PR for the per-class list).
         rule {
             limit {
                 counter = "LINE"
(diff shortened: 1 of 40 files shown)


MagnusSmith:
The three author's-call minors from the panel are applied in the latest commit: FQN→import tidy across the new test additions; the two traversal skip-tests now assert the skipped traversal is genuinely absent (shared `assertGeneratedCodeDoesNotContain` added to `GeneratorTestHelper`) plus `@DisplayName`s, and the class-list-absent test asserts zero generated sources; and every package-private-for-tests member now carries a marker comment so hygiene sweeps don't re-privatise them. `:hkj-processor:check` green with the 99/97 gate.

MagnusSmith:
**Gemini's FoldProcessor finding — declined, with the scenario pinned as a passing test.**

The claimed `ClassCastException` cannot occur: `getElementType` is called only when `isIterableType(component.asType())` returned true on the same component (FoldProcessor lines 124–125), and `isIterableType`'s **first statement** is `if (!(type instanceof DeclaredType)) return false;` — a `TypeVariable` bound to `Iterable<String>` fails that gate and the component is treated as a plain fold target. The cast site carries a comment naming exactly this invariant, and the expert-panel review verified it independently before the guard was collapsed.

Rather than restore a guard whose else-branch is unreachable (which the 99/97 coverage gate would then count against us), the suggested failure scenario is now **encoded as a fixture**: `@GenerateFolds record GenericHolder<T extends Iterable<String>>(T items)` compiles successfully and generates `GenericHolderFolds` — `typeVariableComponentBoundToIterableIsNotTreatedAsIterable` in `FoldProcessorIntegrationTest`. If anyone ever weakens `isIterableType`'s gate, that test (not a user's build) catches it.

`:hkj-processor:check` green with the gate.

Issue #609, opened 2026-07-07
[FEAT] TimeSource — effectful time (java.time.Clock lifted into IO/VTask) for deterministic tests


MagnusSmith:
> **Naming note:** `TimeSource` and the factory names below should be confirmed for best ergonomic fit during implementation. The one fixed decision is what it is **not** called: a public HKJ type named `Clock` would collide with `java.time.Clock` in exactly the files that use both.

## The gap

HKJ has no injectable, effectful time source. `Instant.now()` is called directly in error factories and domain factories (`InventoryReservation.expiresAt`, `BackOrder.estimatedAvailable`, …), so any test asserting on a timestamp — or on time-dependent behaviour generally — is non-deterministic. The order example had to introduce a `setReservationHold(Duration)` test seam *specifically* to make expiry testable without sleeping; every HKJ user hits the same wall the first time they assert on a timestamp.

## Design: lift `java.time.Clock`, don't reinvent it

`java.time.Clock` already *is* the JDK's injectable-time abstraction, with `fixed`/`offset`/`tick` instances built in. What HKJ needs is the **effectful surface** — time as a lazy, composable effect rather than an eager side effect:

```java
public final class TimeSource {
    public static TimeSource system() { ... }                 // wraps Clock.systemUTC()
    public static TimeSource of(java.time.Clock clock) { ... } // lift ANY JDK clock, incl. Clock.fixed(...)
    public static TimeSource fixed(Instant instant) { ... }    // test convenience

    public IO<Instant> now();          // lazy - referentially transparent until run
    public VTask<Instant> nowAsync();  // the VTask twin
}
```

Deterministic tests become one-liners instead of seams:

```java
TimeSource time = TimeSource.fixed(Instant.parse("2026-07-07T00:00:00Z"));

IO<Reservation> reserve(Order order) {
    return time.now().map(t -> new Reservation(order.id(), order.items(), t.plus(hold)));
}
```

## Benefit

- **Library-wide:** any time-dependent HKJ code becomes deterministically testable; the JDK's own `Clock.fixed`/`Clock.offset` instances work unchanged through `TimeSource.of`.
- **Immediate consumers:** the planned typed-error-envelope work (deterministic error timestamps), a planned event-time windowing issue for `VStream` (the same testable-time need — ship one time source, not two), and resilience testing around [#607](https://github.com/higher-kinded-j/higher-kinded-j/issues/607)'s retry backoff.
[...]

PR #615, opened 2026-07-07
TimeSource: java.time.Clock lifted into IO/VTask (#609)

MagnusSmith:
Implements [#609](https://github.com/higher-kinded-j/higher-kinded-j/issues/609).

## What
- **`org.higherkindedj.hkt.time.TimeSource`** (new exported, @NullMarked package): a record wrapping `java.time.Clock` with `system()`/`of(Clock)`/`fixed(Instant)` factories and lazy `now() : IO<Instant>` / `nowAsync() : VTask<Instant>` reads. Deliberately **not named `Clock`** — it must never clash with `java.time.Clock` in the files that use both. Any JDK clock (`fixed`/`offset`/`tick`) lifts through `of()`; the Javadoc states the canonical construction path and when to read `clock().instant()` directly (synchronous edges) vs compose `now()` (effectful pipelines).
- **hkj-test ships `SteppableClock`** (`startingAt`/`advance`/`set`, volatile current instant) — the test convenience the issue anticipated, at 100% own-bundle coverage.
- **Adoption — the issue's evidence case:** `InMemoryInventoryService` takes an injected `TimeSource` (`system()` by default); the reservation hold becomes a constant and the `setReservationHold` test seam is **deleted**. The expiry test now injects a `SteppableClock`, advances it past the hold, and watches the next reservation reclaim the expired one — no seams, no negative-duration hacks, no sleeping.

## Deliberate scope slice
This converts the issue's named evidence site; ~20 further `Instant.now()` call sites across the order/market examples (`BackOrder.estimatedAvailable`, workflow timestamps, `AnomalyDetector`, …) migrate incrementally now the pattern exists — tracked on #609 rather than bloating this PR.

### DIFF
--- a/hkj-book/src/monads/io_monad.md
+++ b/hkj-book/src/monads/io_monad.md
@@ -222,6 +222,22 @@ IO_OP.unsafeRunSync(finalProgram);
 
 ---
 
+## Reading the Clock
+
+The most common side effect after console/file I/O is reading the time — and `Instant.now()` scattered through your code makes every timestamp untestable. `TimeSource` lifts `java.time.Clock` into `IO` so time is a lazy, composable effect:
+
+``` java
+import org.higherkindedj.hkt.time.TimeSource;
+
+TimeSource time = TimeSource.system();               // or TimeSource.fixed(instant) in tests
+
+IO<Reservation> reserve(Order order) {
+  return time.now().map(t -> new Reservation(order.id(), order.items(), t.plus(hold)));
+}
+```
+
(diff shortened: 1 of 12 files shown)


MagnusSmith:
**Gemini review applied** — both findings adopted in the latest commit.

1. **`advance()` atomicity**: the instant now lives in an `AtomicReference` with `updateAndGet`, so concurrent stepping cannot lose updates. Adopted because the class Javadoc explicitly advertises cross-thread use (stepping from a test thread while virtual threads read), so the non-atomic read-modify-write was a real hazard for a published fixture, not a theoretical one.
2. **`withZone` contract**: now returns a new `SteppableClock` with the requested zone **sharing the same timeline** (the `AtomicReference` is shared), per the `java.time.Clock` contract — strictly better than the documented deviation, and it makes the fixture usable for zone-sensitive code under test. Covered by a shared-timeline test (step the original, observe through the zoned view) plus a null guard.

`:hkj-test:check` (100% own-bundle gate) and `:hkj-examples:build` green.

PR #637, opened 2026-07-16
docs: Fix and align ASCII diagrams across the functional book pages

MagnusSmith:
## Summary

Fixes several ASCII-diagram rendering errors in the `functional/` book section, all reported as visibly misaligned on the published site. After the first two, I swept the rest of the book for the same class of bug and fixed every affected page.

### 1. Monad Hierarchy diagram — box width consistency
Two pages share this diagram, and on each a box border didn't match its content width, so a box overhung / shifted off the central axis. Every box is now a uniform **11 columns**, centred on the same vertical axis (connectors aligned at the `┼` branch point).

- **`functional_api.md`** — the `Applicative` box was 10 wide but "Applicative" is 11 chars, so its right `│` overhung. (Also widened `Functor` to match.)
- **`ch_intro.md`** — same diagram; here the `Functor` box was the 10-wide one (Applicative was already 11), so Functor's left edge was offset by one.

**After (both pages)**

```
                   ┌───────────┐
                   │  Functor  │  map
                   └─────┬─────┘
                         │
                   ┌─────┴─────┐
                   │Applicative│  of, ap
                   └─────┬─────┘
```

### 2. `for_optics.md` — "Choosing Between traverseOver and modifyThrough" box
The title row was **70 columns** while the box border is **69**, so the title's right `|` overhung the right edge. Recentred the title to the box's interior width; all 18 diagram lines are now a uniform 69 columns.

### 3. `semigroup_and_monoid.md` — mangled identity law
The two identity-law equations were concatenated onto one line with a broken backtick run (`` `a``empty()` ``). Split into a clean two-item list:

- `a.combine(empty())` must be equal to `a`
- `empty().combine(a)` must be equal to `a`

## Not changed
`monads/vstream_hkt.md` has a similar type-class-tower diagram but already uses uniform 15-wide boxes — verified aligned, left as-is.

## Verification
`mdbook build` succeeds. Connector columns and box widths verified programmatically: the hierarchy diagrams share one central axis, and the `for_optics` box lines are all equal width.


MagnusSmith:
Consolidated into https://github.com/higher-kinded-j/higher-kinded-j/pull/639 together with #638, so all the doc fixes live on one branch (more small fixes to follow).

PR #657, opened 2026-07-25
@GenerateMapping: reject spec methods that collide with generated Impl members (#654)

### LINKED ISSUE #654: [FEAT] @GenerateMapping: reject spec methods that collide with generated Impl members
> **Design note:** the core decision is the rejection rule — **override-equivalent signatures only** (recommended: overloads with different signatures are harmless and legal) versus name-level reservation (simpler message, but bans harmless helpers). Names and diagnostic wording should be confirmed for best ergonomic fit during implementation.

## The gap

A mapping spec is an interface the generated Impl implements, and the generator adds members (`build`, `parse`, `asIso`, `asLens`, `asValidatedPrism`, `updateFrom`, and now `patch`, #625) without checking whether the spec already declares a method with an override-equivalent signature. Two failure modes, both reproduced during the #625 review:

- a spec `default` with the same signature but a **different return type** makes the generated file fail javac ("cannot implement patch(...): incompatible return type") — a raw compiler error inside generated code, with no what/why/fix diagnostic pointing at the spec;
- a spec `default` with the **exact generated signature** compiles, and the generated method silently overrides it — the user's hand-written logic is ignored on `Impl.INSTANCE` with no warning at all.

The silent-override mode is the dangerous one: it inverts the user's mental model ("my default method customises the mapping") without any signal.

```java
@GenerateMapping
interface AccountPatchMapping extends MappingSpec<Account, AccountPatch> {
  default Validated<NonEmptyList<FieldError>, Account> patch(Account a, AccountPatch w) { ... }
}
// error: [@GenerateMapping] 'patch(Account, AccountPatch)' collides with a member the generated
// AccountPatchMappingImpl emits for this spec shape (leaf-carrying projection).
// The Impl's generated member would silently override the default.
// Rename the helper, or delete it and rely on the generated patch.
```

A what/why/fix diagnostic at the spec, before any generated code exists to fail.

- **Rejection rule.** Override-equivalent signatures only (recommended), so `parse(String raw)` helpers and differently-parameterised overloads stay legal; name-level reservation would ban harmless code.
- **Tier-aware member sets.** The reserved set depends on the spec shape: full mappings reserve `build`/`parse`/`asValidatedPrism` (+ `asIso` when lossless); identity projections `build`/`asLens`; leaf-carrying projections `build`/`patch`; `UpdateSpec`s `updateFrom`. The check must run after tier classification, or conservatively reserve the union.
- **Leaf/derived interplay.** Leaf methods (`ValidatedPrism`-returning, zero-parameter) and derived fields (`Getter`-returning) are matched by name today and are never override-equivalent to generated members; the new check must not disturb them, nor the `@MapField` rename stubs (`addRenameStubs`).
- **Where it hooks.** `validateSpecMethods` already walks the spec's methods; the sweep extends it rather than adding a pass.
- **Sibling processors.** `@GenerateMerge` Impls implement the spec's own `assemble` signature (no additional members), and `@GenerateAssembly`/`@GenerateErrorEnvelope` emit companions, not implementations — so the sweep is `MappingProcessor`-scoped unless review finds otherwise.

## Dependencies / scope

- **Depends on** the shipped tier classification (#600/#621/#625/#645); no new machinery.
- **In:** an override-equivalence check across the generated member set, tier-aware, with what/why/fix diagnostics and compile tests for both failure modes per tier.
- **Out:** renaming any generated member; reserving names on companion-style generators.

MagnusSmith:
Closes #654.

## What

`@GenerateMapping` specs are interfaces the generated Impl implements, and until now the generator added members without checking whether the spec already declared an override-equivalent method. A spec `default` with the exact generated signature was silently overridden — the user's "customisation" never ran on `Impl.INSTANCE` — and one with a different return type failed javac *inside the generated file*, with no diagnostic pointing at the spec.

Each `write*Impl` now sweeps the spec's methods against exactly the member set that tier emits, before anything is written, and rejects override-equivalent ones (JLS 8.4.2: name + erased parameter types) with a what/why/fix diagnostic at the offending method:

```
error: [@GenerateMapping] 'patch(Account, AccountPatchDto)' collides with the 'patch' member the
generated AccountPatchMappingImpl emits for this tier (a leaf-carrying projection). The generated
Impl declares an override-equivalent 'patch', so this method is either silently overridden (its
logic never runs on INSTANCE) or fails the generated file's compile with a raw javac error.
Rename the method, or remove it and rely on the generated 'patch'; to customise how a component
maps, declare a ValidatedPrism leaf default named after it.
```

## Design

**Per-writer sweep, not a `validateSpecMethods` extension.** The issue suggested extending `validateSpecMethods`, but that runs before tier classification and cannot know the exact member set (`asIso` losslessness is only computed inside `writeImpl`); the conservative union would falsely ban legal helpers like a `patch(Domain, Wire)` default on a full mapping. Instead each writer passes its own set (recorded on the issue as review finding 3):

| Tier | Reserved members |
|---|---|
| Full mapping | `build(Domain)`, `parse(Wire)`, `asValidatedPrism()`, + `asIso()` when lossless |
| Lossy projection | `build(Domain)`, `asLens()` |
| Leaf-carrying projection | `build(Domain)`, `patch(Domain, Wire)` |
| Sealed dispatch | `build(Domain)`, `parse(Wire)`, `asValidatedPrism()` |
| Sparse update | `updateFrom(Wire)` |

Static and private spec methods are never inherited by the Impl, so they pass; overloads with a different erased signature stay legal; unresolved (error) parameter types are never a collision, so the real cannot-find-symbol diagnostic is not shadowed by a spurious one.

**The `ifPresent` guard moves to the `$` namespace (`hkj$ifPresent`) instead of being reserved.** The review comment proposed reserving the private static null-guard, but implementation showed both available enforcement shapes are wrong: an erasure-based reservation over-rejects (javac only clashes a class static with a signature-*identical* inherited default — same-erasure variants coexist legally), and qualifying the internal calls does not help either (javac resolves `Impl.ifPresent(...)` against inherited instance members too, and errors when a user's more-specific default wins — while *unqualified* calls would be silently captured by such a default). Renaming the private helper into the `$` namespace, which JLS 3.8 reserves for mechanically generated code, kills every mode at once: no ordinary spec method can collide with the declaration or capture the call sites, and the sweep needs no reservation for it. Booby-trapped runtime tests pin capture-immunity on both guarded tiers.

**Diagnostics match the book.** Tier names use the emission-tier vocabulary from `record_mapping.md` ("a lossy projection", "a leaf-carrying projection", …), and the fix sentence routes the customisation-seeking user to the supported mechanism (a `ValidatedPrism` leaf default).

## Tests

Both #654 failure modes are pinned (exact-signature silent override; different-return with `hadErrorCount(1)` proving the raw javac error is replaced, `.inFile` anchoring the diagnostic at the spec). One collision per remaining tier (patch/lens/sealed/update — the sweep never consults return types, so the second mode is exercised once rather than per tier). Negative space: erased-overload and arity overloads, static/private skips, tier-crossing `patch`-on-full-mapping, `asIso` legal when fallible, bounded-generic collision (`<T extends User> build(T)` — erasure match), a `@MapField` rename named after a zero-param member, `build`/`parse` helpers legal on a sparse update, an unresolved-type non-collision, and user `ifPresent` helpers of every shape compiling beside the guard with the guard still running.

`MappingProcessor` remains at 100% instruction and branch coverage; the full multi-module build passes (no existing spec in the repo trips the sweep).

A five-lens panel review (correctness, type safety, ergonomics, architecture, test quality) ran on the initial implementation; every surviving finding is folded in, including the two the panels proved empirically — the `ErrorType` false positive and the guard capture/clash semantics above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compile-time detection of naming/signature collisions between generated mapping helpers and user-declared methods, preventing silent overrides and build failures.
  * Enhanced diagnostics to point to the correct generated mapping context across tiers (including sealed dispatch, projections, and sparse update generation).
  * Namespaced the generated null-guard helper so user-defined `ifPresent`-named methods can’t interfere.

* **Tests**
  * Expanded regression coverage for collision sweep scenarios (including `parse`, `build`, `asValidatedPrism`, `patch`, `asLens`, and sparse `updateFrom`) and for null-guard helper behavior.

### DIFF
--- a/hkj-processor/src/main/java/org/higherkindedj/optics/processing/MappingProcessor.java
+++ b/hkj-processor/src/main/java/org/higherkindedj/optics/processing/MappingProcessor.java
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.StringJoiner;
 import javax.annotation.processing.AbstractProcessor;
 import javax.annotation.processing.FilerException;
 import javax.annotation.processing.ProcessingEnvironment;
@@ -34,11 +35,13 @@ import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.Modifier;
 import javax.lang.model.element.RecordComponentElement;
 import javax.lang.model.element.TypeElement;
+import javax.lang.model.element.VariableElement;
 import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.PrimitiveType;
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
 import javax.lang.model.util.ElementFilter;
+import javax.lang.model.util.Types;
 import org.higherkindedj.optics.annotations.ArityCeilings;
 import org.higherkindedj.optics.annotations.GenerateMapping;
 import org.higherkindedj.optics.annotations.MapField;
@@ -86,6 +89,12 @@ import org.higherkindedj.optics.processing.util.Diagnostics;
  * into an {@code Update} via {@code Edits.accumulate} — no {@code build}, {@code parse}, or {@code
  * as*} tier. A primitive wire property (which can never be absent) is rejected with a diagnostic,
  * and an {@code UpdateSpec} never registers for nesting (it has no {@code parse}).
+ *
+ * <p>A spec method that is override-equivalent to a member the Impl emits for the classified shape
+ * is rejected with a diagnostic at the spec (issue #654): a colliding {@code default} would
+ * otherwise be silently overridden by the generated method, or — with a different return type —
+ * fail javac inside the generated file. Overloads with a different erased signature, and static or
+ * private spec methods (never inherited by the Impl), stay legal.
  */
 @AutoService(Processor.class)
 @SupportedAnnotationTypes("org.higherkindedj.optics.annotations.GenerateMapping")
@@ -866,6 +875,14 @@ public class MappingProcessor extends AbstractProcessor {
     ClassName implName = implClassName(spec);
     ClassName domainClass = ClassName.get(domain);
     TypeName wireName = TypeName.get(wire.element().asType());
+
+    if (!checkNoEmittedCollisions(
+        spec,
+        "a sparse update",
+        List.of(EmittedMember.of("updateFrom", wire.element().asType())))) {
+      return;
+    }
+
     TypeName accumulatedReturn =
         ParameterizedTypeName.get(ACCUMULATED, TypeName.get(domain.asType()));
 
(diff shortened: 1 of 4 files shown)


MagnusSmith:
@coderabbitai help

MagnusSmith:
@coderabbitai full review

MagnusSmith:
@coderabbitai generate sequence diagram

Issue #653, opened 2026-07-23
[FEAT] @GenerateMapping: located nulls for record-wire parse — retire the "a record can never read null" premise


MagnusSmith:
> **Design note:** the core decision to confirm during implementation is the **lossless tier**. `asIso()`'s reverse direction does not go through `parse`, but guarding reference legs makes `parse` partial where it previously threw, so the parse-iso coherence law (`parse(w) == Valid(reverseGet(w))`) no longer holds for null-carrying wires. Recommended: scope coherence to non-null wires and keep `asIso`; the alternative (bean-rule parity, `asIso` only for all-primitive records) breaks every lossless record mapping with a reference component and needs its own migration story.

## The gap

Bean-wire `parse` null-guards every reference read into a located `FieldError` (`must not be null`, #628), because an unset bean property is `null`. Record-wire `parse` legs are built on the premise that "a record can never read null" — identity legs emit bare `Validated.validNel(wire.id())` and leaf legs call `prism.parse(wire.email())` directly. The premise is false at every JSON boundary: Jackson binds a missing property on a record component to `null`, so the most plausible bad input NPEs instead of accumulating:

- `parse(new UserDto(null, ...))` throws (`Validated` rejects null) rather than returning `Invalid([id: must not be null])` — hit empirically in the #627 review, where the Spring showcase returned 500 instead of the documented 422 and had to switch its wire to a bean;
- the #625 patch tier guards its own projected reads, but a nested wire value delegates to the nested spec's full parse, so `patch(domain, CustomerPatchDto(AddressDto(zip=null)))` NPEs out of the user's prism lambda — the located-null promise holds only at depth 1, and the 422 leg turns into a 500 for `{"address":{"zip":null}}`.

## What "good" looks like (user perspective)

```java
UserMappingImpl.INSTANCE.parse(new UserDto(null, "not-an-email", 36));
// Invalid(NonEmptyList[id: must not be null, email: not an email address])
// — total and accumulating over untrusted input, exactly like a bean wire

customerPatch.patch(customer, new CustomerPatchDto(new AddressDto(null)));
// Invalid(NonEmptyList[address.zip: must not be null]) — located through the nesting
```

One null doctrine across both wire shapes: null is a located, accumulated *invalid*, never an exception; absence-as-a-contract remains exclusively `UpdateSpec`'s (bean wires only — a record still cannot express absence, it can only be wrong).

## Things to explore / constraints

[...]

Issue #625, opened 2026-07-10
[FEAT] @GenerateMapping: leaf-carrying projections — a lossy projection that also normalises a field


MagnusSmith:
> **Naming note:** the write-back contract for a fallible projection — a fallible setter returning `Validated`, versus keeping the lens total and routing the validated write through `Edits`/`Update` — is the core design decision to confirm during implementation.

## The gap

A lossy **projection** (wire record with fewer components than the domain) emits an `asLens()` whose `set` writes the projected components back — but only when every projected component is **identity**. A projection that also *normalises* a field through a leaf (trim/lowercase an email, parse a code) is rejected today, because a leaf makes the write-back fallible and `Lens.set` is total. So you must choose: project, or validate — not both. See the "Diagnostics and limits" note on [Record Mapping](https://higher-kinded-j.github.io/optics/record_mapping.html).

## What "good" looks like (user perspective)

```java
public record Account(String id, String email, Instant createdAt) {}
public record AccountPatch(String email) {}     // projection: only email is writable

@GenerateMapping
public interface AccountPatchMapping extends MappingSpec<Account, AccountPatch> {
  default ValidatedPrism<String, EmailAddress> email() { return EmailCodecs.EMAIL; }
}
// desired: a projecting mapping whose write-back validates email, surfacing a located
// FieldError instead of silently writing an invalid value or refusing to generate.
```

- **A truthful surface** for "project + validate": either a fallible write returning `Validated<NonEmptyList<FieldError>, Account>`, or the existing `Edits.accumulate` railway (a `patch` that accumulates), chosen by the naming note.
- **The lawful-optic guarantee holds** — whatever tier ships is law-checked against the published harness (#596; `MappingLaws` in #621).

## Things to explore / constraints

- **Which optic.** A plain `Lens` cannot be fallible; options are an affine/`ValidatedPrism`-flavoured write-back, or routing projection-writes through `Edits.accumulate` ([#582](https://github.com/higher-kinded-j/higher-kinded-j/issues/582)) so the fallible field lands on the validation railway.
- **Composition with derived fields.** #621 rejects projection + derived field; this issue must say whether project + leaf is allowed while project + derived stays rejected, and why.
- **Law obligations.** Define and check the laws for the chosen fallible-write tier.
- **Diagnostics** for the still-unsupported combinations, what/why/fix.

## Dependencies / scope

[...]

Issue #553, opened 2026-06-04
[FEAT] Align TryMonad.recoverWith with Either/Maybe: fail-fast on a null fallback


MagnusSmith:
## Summary

  `TryMonad` does not override `recoverWith`, so it inherits the `MonadError` default, which never null-checks its `fallback` argument. `EitherMonad`
  and `MaybeMonad` both override `recoverWith` to reject a null `fallback` eagerly. As a result, `Try.recoverWith(ma, null)` has surprising,
  **state-dependent** behaviour that diverges from its siblings and from the rest of `Try`'s own null-argument handling.

  ## Current behaviour

  The inherited `MonadError` default:

  ```java
  default <A> Kind<F, A> recoverWith(Kind<F, A> ma, Kind<F, A> fallback) {
      return handleErrorWith(ma, error -> fallback);   // fallback only read if ma is a Failure
  }
  ```

  Because `Try` inherits this, a **null `fallback`** behaves differently depending on `ma`'s runtime state:

  | Call | Result today |
  |------|--------------|
  | `recoverWith(success, nonNull)` | returns the Success ?  |
  | `recoverWith(failure, nonNull)` | returns the fallback ?  |
  | `recoverWith(success, null)` | **returns the Success** ? the null is silently ignored |
  | `recoverWith(failure, null)` | **returns `Failure(NullPointerException)`** whose message is `"Function handler in handleErrorWith returned null,
  which is not allowed"` |
  | `recoverWith(null, ?)` | throws NPE (via `handleErrorWith`'s validation) |

  The Javadoc already documents `fallback` as *"Assumed non-null."*, so passing null violates the declared contract.

  ## Why this is a problem

  - **State-dependent outcome for the same illegal input.** The identical bug (null `fallback`) is silently swallowed when `ma` is a Success but
  surfaces as a `Failure` when `ma` is a Failure ? i.e. only on the error path, which is the least-tested path. Classic heisenbug.
  - **Leaky, misleading error.** The captured message mentions `handleErrorWith` / `handler`, neither of which the caller used ? it points debugging at
  the wrong method. This is an accident of the default's `recoverWith = handleErrorWith(ma, _ -> fallback)` delegation, not designed behaviour.
  - **Inconsistent with the rest of `Try`.** `Try` already fails fast on every other null *API argument* ? `Try.failure(null)`, `map(null)`,
[...]

PR #662, opened 2026-07-25
ValidatedPrism: located element nulls and index-located container failures (#660)

### LINKED ISSUE #660: [FEAT] ValidatedPrism: located element nulls and index-located failures in parseAll/parseValues
> **Design note:** the core decision is the **index-segment convention**. `parseValues` already key-locates via `FieldError.at(key.toString())`, rendering `tagged.k: ...`; the list analogue needs a segment form for `emails[1]` — either a bracketed segment (`emails[1]`, reads naturally in 422 payloads) or a bare positional segment (`emails.1`, consistent with the key grammar). The choice flows through to `hkj-spring`'s `FieldErrorItem(path, segments, message)`, so confirm rendering there before locking it.

## The gap

Two related holes in `ValidatedPrism`'s container parsing, both left explicitly on the table by #653:

- **A null element throws out of an otherwise fully-guarded parse.** #653's doctrine — a component-level null is a located, accumulated invalid, never an exception — stops at the container reference: `parseAll` and `parseValues` `requireNonNull` each element/value, so `{"emails": ["a@b", null]}` NPEs (`sources[1] must not be null`) where every sibling failure accumulates. The boundary is documented and test-pinned in #653, with the message naming the index but not the field.
- **List element parse failures are not index-located at all.** `parseAll` combines raw element errors, so two bad emails render as two indistinguishable `emails: not an email address` entries — the caller cannot tell which element failed. `parseValues` key-locates; the list side lacks the analogue, and the patch-tier test named "List elements locate by index" currently overstates (its assertion pins only the `emails` prefix).

These are one change: locating a null element requires exactly the index machinery that element parse failures are missing.

```java
rosterMapping.parse(new RosterDto("7", List.of("a@b", "nope")));
// Invalid(NonEmptyList[emails[1]: not an email address]) — the failing element is identified

rosterMapping.parse(new RosterDto("7", Arrays.asList("a@b", null)));
// Invalid(NonEmptyList[emails[1]: must not be null]) — located and accumulated, never thrown
```

- **hkj-core public-API behaviour change.** `parseAll`/`parseValues` document `@throws NullPointerException` for elements today; converting the element case to a located `Invalid` is the same friendly-direction change #653 made for legs, but on a public core method — the release note carries it. The *container* null itself stays the caller's contract at the direct-call surface (generated legs already guard it since #653).
- **Index location changes existing rendered paths.** `emails: not an email address` becomes `emails[1]: ...` — observable in 422 payloads and asserted in tests; the patch-tier "locate by index" test finally asserts what it claims.
- **Null doctrine boundaries that do NOT move:** `Validated.of`/`validNel` rejecting null is deliberate (GH-553) and untouched; a null Map *key* is structurally broken and stays `requireNonNull`; the build direction (`buildAll`/`buildValues`) remains total-and-throwing — only the parse direction is a validation surface.
- **Retirements.** The #653 boundary pins (the `sources[1] must not be null` NPE test, the book's "a null element follows the container's own contract" sentences in `record_mapping.md` and the generated-class javadoc, and the skill's boundary line) all get updated to the new located behaviour.

## Dependencies / scope

- **Depends on** #653 (doctrine, boundary pins to retire) and the `hkj-spring` path rendering (`FieldErrorItem` segments).
- **In:** index-located element parse failures and located element nulls in `parseAll`; located null values in `parseValues`; the segment-convention decision; retiring the #653 boundary pins; book, skill and 0.4.9 release-note updates.
- **Out:** the build direction; `Validated`'s own null doctrine; Map null keys; any leg-emission change in the processors (the generated calls are already correct — the change is inside the two hkj-core methods).

MagnusSmith:
Closes #660.

## What

The null doctrine (#653/#659) now reaches *inside* containers, and list failures finally say **which** element failed:

```java
rosterMapping.parse(new RosterDto("7", List.of("a@b", "nope")));
// Invalid(NonEmptyList[emails.1: not an email address])   — the failing element is identified

rosterMapping.parse(new RosterDto("7", Arrays.asList("a@b", null)));
// Invalid(NonEmptyList[emails.1: must not be null])       — located and accumulated, never thrown
```

`ValidatedPrism.parseAll` locates every failing element by its index and treats a null element as a located `must not be null`; `parseValues` treats a null value the same under its key. Accumulation continues past a null (pinned: `[valid, null, invalid]` reports both failures in order), and the composed location through a nested spec comes out label-index-inner (`customers.1.email`, pinned end to end at the mapping seam).

## The index-segment decision

The issue left the convention open (`emails[1]` vs `emails.1`). This lands **`emails.1`** — a plain positional segment:

- it matches `parseValues`' existing key grammar (`attributes.en`), so one rendering rule covers both container kinds;
- it needs zero changes to `FieldError`, the dotted renderer, or `hkj-spring`'s `FieldErrorItem` (segments stay opaque strings);
- bracketed rendering is exactly what the deferred sealed `PathSegment` model provides properly (`Index` as a typed case) — doing brackets now via string conventions would pre-empt it. The deferral is recorded in the release note, the book's new bulk-forms section, and as a warning in the PathSegment draft (switching live 422 paths to brackets later is a second observable behaviour change).

## Boundaries that do not move

A null *list*, *map*, or map *key* stays the caller's `requireNonNull` (a null key is a structurally broken map, not a wrong value), and the build direction (`buildAll`/`buildValues`) remains total-and-throwing. `Validated`'s own null rejection (GH-553) is untouched, and no processor emission changes — the generated container legs were already correct.

## Behaviour changes (0.4.9 release note)

Both observable, both the friendly direction: list-element failure paths gain the index segment (`emails:` → `emails.1:` in 422 payloads — consumers keying on exact paths should note the added segment), and element/value nulls stop throwing `NullPointerException` out of `parseAll`/`parseValues`.

## Docs and tests

The #653 boundary pins are retired at every site (prism tests, the mapping container tests, the book doctrine paragraph, the skill, the class javadoc), and the patch tier's "List elements locate by index" test now asserts exactly what its name has claimed since #625. `validated_prism.md` gains the bulk-forms section the doctrine paragraph links to; the nesting page's container bullets gain the list-index analogue their `Map` sibling always had; the `hkj-bridge`/`hkj-optics` skill tables note the index location; and the "only a null wire" phrasing at three sites now names the null-map-key boundary.

`ValidatedPrism` holds 100% instruction and branch coverage; the full multi-module build passes (hkj-spring included — `FieldErrorItem` treats segments as opaque, verified by the panel).

A three-lens panel review ran: correctness came back clean (index bookkeeping, composition order, semigroup ordering, and an exhaustive caller sweep all verified); the tests and docs findings above are folded in.

## Summary by CodeRabbit

* **Bug Fixes**
  * Null elements in `parseAll` (lists) and null values in `parseValues` (maps) are now returned as located validation errors (`must not be null`) instead of throwing exceptions.
  * Failures in bulk parsing are now precisely located by element index or map key and accumulate across the entire input.
  * Nested container/list/map errors now report more accurate dotted paths, including container fields and indexes.
* **Documentation**
  * Updated guidance on “null doctrine,” bulk parsing/build behavior, and indexed/key-based error locations.
* **Tests**
  * Adjusted assertions to expect index/key-located errors and continued accumulation past nulls.


MagnusSmith:
Both CodeRabbit comments addressed in 435e642bb.

**buildAll row (accepted):** both bulk-form rows in the hkj-optics skill table now attribute failure location to the parse form and totality to the build form — the `parseValues`/`buildValues` row had the identical ambiguity, so it is swept in the same pass.

**Null-container boundary (rebutted, then disambiguated):** adding null lists/maps to the doctrine paragraphs' caller-contract list would have documented a false boundary — inside a *generated* `parse`, a null container **component** is guarded like any reference read and reports `emails: must not be null` (pinned by `containerLegAndElementNullsAreLocated`); only calling `parseAll`/`parseValues` **directly** with a null list/map is the caller's `requireNonNull`, which the bulk-forms section in `validated_prism.md` documents. The ambiguity was real though, so both flagged sites now spell out the two surfaces explicitly.

Issue #624, opened 2026-07-10
[FEAT] @GenerateMapping: generic record types — map parameterised records with threaded type parameters


MagnusSmith:
## The gap

`@GenerateMapping` ([#600](https://github.com/higher-kinded-j/higher-kinded-j/issues/600), Step-0 [#605](https://github.com/higher-kinded-j/higher-kinded-j/pull/605), remainder [#621](https://github.com/higher-kinded-j/higher-kinded-j/pull/621)) resolves **concrete** record types only. A generic domain/wire pair — `Page<T>` ↔ `PageDto<T>`, `Result<E, A>` ↔ `ResultDto<E, A>` — cannot be mapped: the classifier matches components by concrete name + type and has no way to thread a type parameter (or a per-element leaf keyed on it) through `build`/`parse`.

## What "good" looks like (user perspective)

A parameterised spec whose element mapping is supplied (identity, a leaf, or a sibling spec), lifted through the container shape:

```java
public record Page<T>(List<T> items, int total) {}
public record PageDto<T>(List<T> items, int total) {}

@GenerateMapping
public interface PageMapping<T> extends MappingSpec<Page<T>, PageDto<T>> {}
// build/parse are generic in T; identity element T lifts through the List component,
// or an in-scope MappingSpec<T, TDto> maps Page<T> -> PageDto<TDto>.
```

- **Type parameters thread through** `build`/`parse` and the emitted optic (`asIso`/`asLens`/`asValidatedPrism`) as real generics, not erased casts.
- **Bounded and multi-parameter** records are in scope; wildcards stay a diagnosed error (consistent with `Map` in #621).
- **Emission tiers are unchanged** — a generic lossless mapping still offers `asIso()`.

## Things to explore / constraints

- **Classifier + emit under type variables.** Match components whose types differ only by a shared type variable; generate generic methods; avoid unchecked casts (the type-safety initiative bar, #560–#565).
- **Element-mapping source.** Same-round spec registry keyed by the *element* type, or an explicit leaf — define how a `Page<User>` finds `UserMapping`.
- **Interaction with containers.** `List`/`Optional`/`Map` lifting (#621) must compose with a type variable as the element/value type.
- **Diagnostics**: raw generic use, unresolved element mapping, wildcard arguments — each what/why/fix.

## Dependencies / scope

- **Depends on** #605 and container lifting (#621).
- **In:** generic record ↔ DTO mappings with threaded type parameters and truthful generic optics.
[...]

PR #663, opened 2026-07-25
@GenerateMapping: concrete instantiations of generic records (#624, first slice)

MagnusSmith:
Part of #624 (first slice — the issue stays open for threaded type parameters).

## What

`@GenerateMapping` now maps generic records as **concrete instantiations**:

```java
public record Page<T>(List<T> items, int total) {}
public record PageDto<T>(List<T> items, int total) {}

@GenerateMapping
public interface UserPageMapping extends MappingSpec<Page<User>, PageDto<UserDto>> {}
// items : List<User> <-> List<UserDto>, lifted through the sibling UserMapping;
// a bad element parses to items.1.email: not an email address, a null one to items.1: must not be null
```

Every component classifies under `Types.asMemberOf` substitution, so the existing machinery — leaves, renames, sibling nesting, `List`/`Optional`/`Map` lifting, all emission tiers including the lossy projection and the validated `patch`, the null doctrine and index location, the #654 collision sweep — applies unchanged. An instantiated mapping registers like any other, so `Report(Page<User> results)` nests it automatically, and `@GenerateMerge` fills resolve through it (panel-verified).

## Design

- **POC-first, user-confirmed direction.** A spike validated the hypothesis before committing: concrete instantiation *is* element-mapped generics resolved statically per type — the registry supplies the element mapping at compile time, at the cost of a two-line spec per instantiation, with the `INSTANCE` convention untouched. **Threaded generic specs** (`PageMapping<T>`) remain the issue's open remainder: they force the singleton-vs-factory decision (a generic Impl cannot carry a typed static `INSTANCE`), and deferring them creates no lock-in — this slice emits ordinary non-generic Impls that threading would later add to, not reshape.
- **Substitution only where it means something.** `componentType` substitutes only for generic owners, keeping every pre-#624 spec on the exact prior path (byte-identical emission, panel-verified) and clear of any ECJ `asMemberOf`-on-record-component question.
- **The gate**: generic *specs* are diagnosed with a pointer to concrete instantiation; raw uses — including raw *nested* arguments, the slice's one panel-caught defect (the recursion was vacuously true for them while the safer wildcard was rejected) — and wildcard or type-variable arguments get what/why/fix diagnostics naming the full offending type; array arguments (`Page<String[]>`, `Page<int[]>`) are concrete and map fine. **Record-record pairs only** for now: a bean-shaped wire and a sparse `UpdateSpec` keep their pairs concrete (pinned), and the surviving `checkNotGeneric` fix-line on those paths now states that constraint instead of denying the record-path feature.

## Tests, laws, docs

Emission, runtime (full-stack `items.1.email` and null-element legs), free sibling nesting, projection and patch tiers under substitution, `Map`/`Optional` lifting, the #654 sweep on an instantiated spec, array/deep/bounded accepts, and every diagnostic shape are pinned; the two old generic-rejection pins flip to positive pins; `GeneratedMappingLawsTest` law-checks the instantiated pair through the fallible-tier overload. The book gains a compiled section (`RecordMappingBook.java` anchors, inside the snippet-verification gate), and the annotation javadoc, 0.4.9 release note and `hkj-mapping` skill all state the scoped story. Strictly additive — every newly-accepted shape was previously a compile error, so no behaviour-change flag.

`MappingProcessor` holds 100% instruction and branch coverage; the full multi-module build passes.

A three-lens panel review ran on the initial commit; all surviving findings (the nested-raw defect, array acceptance and diagnostics, the misleading surviving fix-line, doc scoping, the unverified book snippet, and the tier/container/sweep test gaps) are folded into the second commit.

## Summary by CodeRabbit

* **New Features**
  * Added support for mapping **concretely instantiated** generic records (including nested records, container lifting, arrays, and bounded generics), with correct propagation of validation and projection behavior.
  * Added generic-record examples demonstrating indexed validation error locations.
* **Documentation**
  * Clarified which generic forms are accepted vs rejected (raw uses, wildcards, and threaded type parameters).
  * Documented that `@GenerateAssembly` does **not** support generic records.
* **Bug Fixes**
  * Improved diagnostics and ensured generic-record mappings are rejected for unsupported bean-wire scenarios.
* **Tests**
  * Added coverage validating mapping laws and acceptance/rejection behavior for generic instantiations.


MagnusSmith:
Both nitpicks addressed in the latest commit: the `int[]` and `List<String>` instantiations now pin their generated `build` signatures (the suggestion matched our own panel's single-carrier flag), and the 'keeps its pair concrete' phrasing at all four sites is replaced with the explicit boundary — concrete generic instantiations are record-to-record only; bean-shaped wires and `UpdateSpec` mappings do not support them yet.

PR #664, opened 2026-07-25
@GenerateMapping: threaded generic specs (#624 slice 2)

MagnusSmith:
Part of #624 (slice 2 of the locked full-mapper sequence — the element-mapped slice 3 follows).

## What

Generic specs now **thread their type parameters**: one mapping serves every instantiation.

```java
@GenerateMapping
public interface PageMapping<T> extends MappingSpec<Page<T>, PageDto<T>> {}

PageDto<String>  tags   = PageMappingImpl.<String>instance().build(stringPage);
Validated<NonEmptyList<FieldError>, Page<Integer>> counts =
    PageMappingImpl.<Integer>instance().parse(intPageDto);
```

The generated Impl is generic — the spec's variables and bounds on the class, the parameterised spec implemented, the whole surface threaded (`build`/`parse`/`asValidatedPrism`, and `asIso` on a lossless pair) — and is reached through the **locked decision (a)**: a cached stateless singleton behind a `@SuppressWarnings("unchecked") static instance()` factory, which is hkj-core's own generic-singleton convention (`EitherMonad.instance()`, `ReaderMonad.instance()`). Concrete Impls keep `INSTANCE` untouched — byte-identical output, verified against JavaPoet's category-ordered emission.

## Design

- **One gate, two composable forms.** `supportedArgument` accepts, per type argument and recursively, either a concrete type (slice 1) or one of the spec's own variables — so `Page<List<T>>` threads and `Result<E, A>`/`Ranked<T extends Number>` work. Raw uses (including nested), wildcards, and (defensively) foreign variables are diagnosed under one message grammar, with the fix-hint naming the failing side.
- **Everything already built composes**: same-variable elements copy by identity (pinned, including the deliberate null-element pass-through on identity legs), same-typed `default` leaves route elements, `@MapField` renames, derived `Getter`s, and the lossy-projection and `patch` tiers all thread; the #653 guards and #654 sweep run unchanged — and the sweep now also **reserves `instance()`** on generic Impls, the panel's one production catch (a spec's `default String instance()` would otherwise have broken the generated file with a raw javac error).
- **Deliberately deferred to slice 3, both pinned**: threaded specs are excluded from the nesting registry (use sites need unification), and an *abstract* leaf (`ValidatedPrism<TDto, T>` — the locked (A) element-mapped shape) stays diagnosed.
- **Cast soundness probe-verified**: no tier ever puts `T`-typed state on the Impl — leaves and derived getters live on the spec and are invoked per call — so one shared instance across instantiations is observable only as reference identity, which is pinned.

## Panel review

Three lenses ran on the implementation commit. Correctness came back clean with javac probes (bounded-diamond emission, `Page<T>::new` scope, the sweep converting the JLS 8.4.8.3 erasure clash into a located diagnostic, registry-exclusion completeness — a generic Impl's `INSTANCE` being private means no leak path could compile). The docs and tests findings are folded into the second commit: the `instance()` reservation, the fabricated-name fix-hint, the self-contradicting release-note paragraph, the book's stale limits bullet and dead anchor, honest labelling of the wildcard-in-holder test, and carriers for every threaded tier, renames, derived fields, nested-variable shapes, and the mixed-argument component diagnostic.

`MappingProcessor` holds 100% instruction and branch coverage; the full multi-module build passes, including the compiled `threaded_spec`/`threaded_usage` book anchors.

## Summary by CodeRabbit

* **New Features**
  * Added threaded generic record mappings that reuse a single generated mapper across type instantiations via a shared singleton accessor.
  * Generated mappers for generic specs now support a typed `instance()` access pattern for reuse.
* **Documentation**
  * Expanded “Generic records” guidance, including singleton behavior and supported/unsupported mapping shapes and limits.
* **Bug Fixes**
  * Improved diagnostics for unsupported raw/wildcard/invalid generic shapes and clarified “not supported instantiation” cases.
  * Resolved collisions related to the shared singleton accessor.
* **Tests**
  * Added/updated law, runtime, and processor tests covering threaded generics, signatures, tier interactions, and diagnostics.


MagnusSmith:
All six nitpicks addressed in the latest commit. One note on the ERROR-kind suggestion: the premise needed a correction — `ErrorType` *extends* `DeclaredType`, so an unresolved argument would have hit the `DeclaredType` pattern arm (with unpredictable results), not the default arm. The step-aside check therefore runs first, explicitly, with a comment explaining the ordering and a test pinning that javac's `cannot find symbol` is the only diagnostic. The rest landed as suggested: `reserveInstanceIfGeneric` dedupes all three tier lists, `CompiledResult.genericInstance()` consolidates the reflection, the fixture is renamed to `wildcardInHolderSpec`, and both weak pins now assert the emitted expression and the actual classifier diagnostic.

PR #665, opened 2026-07-26
@GenerateMapping: spec inheritance via plain mix-in interfaces (#623)

### LINKED ISSUE #623: [FEAT] @GenerateMapping: spec inheritance — inherited @MapField renames and leaf methods across a spec hierarchy
> **Naming note:** whether inheritance is expressed by a spec `extends` another `@GenerateMapping` spec, by shared non-`MappingSpec` interface mix-ins carrying `@MapField`/leaf methods, or both, is a design point to confirm during implementation.

## The gap

`@GenerateMapping` ([#600](https://github.com/higher-kinded-j/higher-kinded-j/issues/600), Step-0 in [#605](https://github.com/higher-kinded-j/higher-kinded-j/pull/605); remainder in [#621](https://github.com/higher-kinded-j/higher-kinded-j/pull/621)) resolves a spec that `extends MappingSpec<D, W>` **and nothing else**: it cannot inherit `@MapField` renames or `default ValidatedPrism`/`Getter` leaf methods from a shared parent interface. Families of DTOs that share a rename/leaf vocabulary (every wire type renames `id` → `ref`, validates `email` the same way) must re-declare that vocabulary on every spec.

A shared mix-in of correspondences, inherited by each concrete spec:

```java
public interface WireConventions {                      // shared vocabulary, not a MappingSpec
  @MapField(to = "ref") String id();                    // every wire renames id -> ref
  default ValidatedPrism<String, EmailAddress> email() { return EmailCodecs.EMAIL; }
}

@GenerateMapping
public interface UserMapping  extends MappingSpec<User,  UserDto>,  WireConventions {}  // inherits both

@GenerateMapping
public interface AdminMapping extends MappingSpec<Admin, AdminDto>, WireConventions {}  // same, no repeat
```

- **Inherited renames and leaves resolve** exactly as if declared directly, with the same what/why/fix diagnostics.
- **A spec may override** an inherited correspondence locally; local wins, and a redundant identical override is optionally flagged.
- **Exhaustiveness is unchanged** — every target component is still accounted for, counting inherited correspondences.

- **Resolution order.** Walk the spec's interface hierarchy for `@MapField` and leaf `default` methods; define precedence (local over inherited, and between multiple parents) with a diagnostic on a genuine conflict.
- **Interaction with the claiming rules.** Lifting "extends `MappingSpec` and nothing else" must not let an unrelated interface's methods be mis-claimed as leaves/derived fields — the derived-field claiming rule from #621 (a zero-arg `Getter`-returning `default` is *always* a derived field) has to stay predictable across a hierarchy.
- **Diagnostics** follow the what/why/fix standard, naming the declaring interface of an inherited correspondence.
- **Processor conventions**: `latestSupported()`, originating elements (now spanning several source files), 100% coverage per the raised hkj-processor gate.

## Dependencies / scope

- **Depends on** `@GenerateMapping` Step-0 (#605, merged) and the remainder (#621).
- **In:** inheriting `@MapField` renames and leaf/derived `default` methods from parent interfaces; precedence + conflict diagnostics; unchanged exhaustiveness.
- **Out:** generic type parameters on specs (separate issue); bean-shaped sources (separate issue).

MagnusSmith:
Closes #623. Stacked on #664 (threaded generics) — will retarget `main` once that merges.

## What

A `@GenerateMapping` spec may now extend plain **mix-in interfaces** alongside `MappingSpec`/`UpdateSpec`, sharing renames, leaves and derived fields across specs:

```java
interface ContactVocabulary {
  @MapField(to = "fullName")
  String name();

  default ValidatedPrism<String, EmailAddress> email() { return EmailCodecs.EMAIL; }
}

@GenerateMapping
interface ClientMapping extends ContactVocabulary, MappingSpec<Client, ClientDto> {}

@GenerateMapping
interface SupplierMapping extends ContactVocabulary, MappingSpec<Supplier, SupplierDto> {}
```

## Design

- **Java's own precedence, not an invented order.** Every member scan now walks the hierarchy through `Elements.getAllMembers(spec)` (filtered to interface-declared methods), so an inherited member counts exactly as if declared on the spec, a local override hides the mix-in's member, and genuinely conflicting parents are already a javac error before the processor runs. Interface statics are not inherited (JLS 8.4.8) and stay inert.
- **The #654 collision sweep is the load-bearing walk.** Its completeness proof used to rest on the single-interface rule; a mix-in's `default build(User)` would have bypassed it. The sweep now sees inherited members and its diagnostic names the declaring interface — `'build(Account)' (inherited from 'BuildVocabulary') collides with the 'build' member…` — as do the validate-stage diagnostics.
- **Mix-ins only, two gates.** A mix-in that is itself a mapping spec (directly or transitively extends `MappingSpec`/`UpdateSpec`) is rejected — vocabulary sharing and Impl-generating specs stay distinct concepts. A generic mix-in is rejected for now (inherited member types are read as declared; substitution under an instantiation would need `asMemberOf` threading — a later slice if wanted).
- **Everything composes**: threaded generic specs (#664) extend non-generic mix-ins, `UpdateSpec` mappings inherit the same way (leaves and renames both pinned), a sealed pair rejects an inherited `@MapField` exactly like a local one, and the #653/#660 null doctrine applies to inherited legs unchanged. `@GenerateMerge` deliberately keeps its single-interface gate.

## Panel review

Three lenses ran on the implementation commit; the correctness lens probed the semantics against javac itself. The probe's catch: `Elements.getAllMembers` returns **both** override-equivalent abstract renames from unrelated mix-ins (JLS 9.4.1 lets abstracts coexist where defaults may not), so two mix-ins *agreeing* on a rename were falsely rejected as a claim conflict — and would have emitted duplicate stubs past that guard — while *conflicting* targets resolved by order-dependent last-wins. The second commit folds agreement into one rename/one stub, gives conflicts a dedicated diagnostic naming both declaring interfaces, and adds the ERROR-parent step-aside. The probes confirmed the good news too: a true diamond is one symbol, and a spec-local re-declaration hides the mix-in's member — both now pinned as suite tests alongside the docs adjustments.

`MappingProcessor` and `MergeProcessor` hold 100% instruction and branch coverage; the full multi-module build passes, including the compiled `mixin_spec`/`mixin_usage` book anchors shared by two specs.

## Summary by CodeRabbit

* **New Features**
  * Mapping specifications can reuse shared vocabulary mix-in interfaces for renames, leaves, and derived fields (including inherited and diamond-shaped reuse).
  * Update mappings now honor mix-in-provided renames/leaves, including sparse PATCH write-back behavior.
* **Bug Fixes**
  * Improved validation and diagnostics for conflicting inherited renames and invalid mix-ins (including generic mix-ins and mix-ins that are themselves mapping/update specs).
  * Added/clarified sealed-mapping diagnostic behavior when `@MapField` originates from inherited vocabulary.
* **Documentation**
  * Updated “Shared vocabulary: mix-in interfaces” guidance, Javadoc, examples, and release-history for spec mix-in support.


MagnusSmith:
Both applied in f7d9a909e. `extendsMappingFamily` now has the same `TypeKind.ERROR` step-aside as `checkMixins` and `supportedArgument` — the symmetry argument holds (an error type can neither be nor extend the mapping family), and it is pinned by a new carrier test where a resolvable mix-in's own superinterface is unresolved: javac's cannot-find-symbol is the only error, no mix-in diagnostic, no processor crash. The stranded `validateSpecMethods` javadoc is back on its method. `MappingProcessor`/`MergeProcessor` hold 100% line and branch coverage.

PR #666, opened 2026-07-26
@GenerateMapping: element-mapped generic specs + nesting for every generic form (#624 slice 3)

### LINKED ISSUE #624: [FEAT] @GenerateMapping: generic record types — map parameterised records with threaded type parameters

MagnusSmith:
Closes #624 (the final slice: element-mapped generic specs, and nesting for every generic form).

## What

`@GenerateMapping` now maps generic records **three** ways, and all three nest.

The new **element-mapped** form threads the two sides under *different* variables and declares the element mapping as an abstract leaf:

```java
@GenerateMapping
interface CodecPageMapping<T, TDto> extends MappingSpec<Page<T>, PageDto<TDto>> {
  ValidatedPrism<TDto, T> items();          // no body: supplied at of() time
}

Validated<NonEmptyList<FieldError>, Page<EmailAddress>> mail =
    CodecPageMappingImpl.of(EmailCodecs.EMAIL).parse(pageDto);   // one prism per leaf
```

And the two slice-2 deferrals are retired — nothing in the mapper is "not yet nestable" any more.

## Design

- **`of()` emission (3a).** An abstract zero-parameter `ValidatedPrism` method on a generic record spec is a leaf. The generated Impl stores each as a constructor-supplied `private final` field behind a public `of(...)` factory (one prism per leaf, declaration order, agreeing mix-in leaves folded), implements the abstract method from the field, and has **no singleton in either spelling** — it carries leaf-typed state, so every `of(...)` is a fresh immutable instance. The #654 sweep reserves `of(prisms)` in place of `instance()`. An abstract leaf on a concrete or sealed spec is diagnosed (nothing defers its parser).
- **Threaded nesting by unification (3b).** Threaded specs register with their declared mirrors (`Page<T>`, `PageDto<T>`); use sites resolve them by structural first-order unification — every spec variable bound consistently across both sides to an argument the *using* spec supports, ERROR kinds stepping aside — and emit `PageMappingImpl.<String>instance().asValidatedPrism()`. A generic outer spec may thread its own variable straight through.
- **Element-mapped nesting by composition (3c).** Element-mapped specs register too; a use site whose pair unifies against one composes the factory in place — `CodecPageMappingImpl.of(entries()).asValidatedPrism()` — resolving each leaf's element pair (its `ValidatedPrism` type read under the use-site bindings via `asMemberOf`) through a component-named leaf on the using spec (single-leaf specs) or recursively through the registry. An unresolvable element pair is a what/why/fix diagnostic naming the pair and both levers. Composition **recurses and terminates** structurally: `Page<Page<String>>` yields `of(of(nested()))`.

## Tests & laws

`MappingLaws` covers the `of()` tier. The suite pins each form's compile shape and runtime behaviour (round-trip plus located failures — `entries.items.1: not an email address` through a composed path), the full unification refusal matrix (inconsistent bindings, unbound/wildcard/raw/unresolved use sites, array structural unification), multi-leaf composition, ambiguity propagation, and the two non-obvious compositions the inline review surfaced (recursion, generic-outer element-mapped threading).

## Panel review

Ran **inline** (subagents were unavailable). javac probes cleared the highest-risk paths: leaves named `of`/`domain`/`build`/`wire` all compile (accessor and tier method coexist as overloads); the generic-outer element-mapped case composes cleanly with type-variable bindings; recursion terminates. No defects found; the two most interesting probes are now pins.

`MappingProcessor` and `MergeProcessor` hold 100% line and branch coverage; the full multi-module build passes, including the compiled `element_spec`/`element_usage` book anchors. Docs updated across the book, `GenerateMapping` javadoc, release note and mapping skill (STYLE-GUIDE-compliant, timeless wording).

## Summary by CodeRabbit

- **New Features**
  - Added support for *element-mapped* generic record mappings, with externally supplied per-element prisms via a generated `of(...)` factory.
  - Improved generic nested mapping composition for both threaded and element-mapped forms, including more precise element-level error location.

- **Documentation**
  - Updated `@GenerateMapping`, record-mapping docs, and release notes to describe three supported generic mapping forms and clarify constraints/diagnostics (record-to-record, raw/wildcard handling, and array unification).

- **Tests**
  - Added law and compilation coverage for element-mapped runtime behavior and enhanced diagnostic expectations.


MagnusSmith:
Both actionable findings applied in f7c5e2c4f, plus three of the nitpicks.

**Self-covering recursion (Major) — real, fixed.** Confirmed with a javac probe: `NodeMapping<T,U> extends MappingSpec<Node<T>, NodeDto<U>>` with `ValidatedPrism<NodeDto<U>, Node<T>> value()` overflowed the stack exactly as described (the leaf's element pair *is* the spec's own declared pair). `elementMappedComposition` now threads the instantiated spec types on the recursion stack and emits a `which maps itself … never terminates` what/why/fix diagnostic when one repeats. I keyed the check on the fully **instantiated** spec type (one `isSameType`) rather than the raw (domain, wire) pair — it captures both sides at once, so a legitimately shrinking recursion like `Page<Page<String>>` (distinct instantiations at each depth) is untouched, and there's no asymmetric partial-match branch. Pinned by a new `selfCoveringElementMappedSpecDiagnosed` test.

**Ignored unify results (nitpick) — left as-is, with reason.** The two `unify` calls in the single-match path are guaranteed total because `covers()` just filtered the candidate with the same unification plus the `bindings.size() == typeParameters.size()` check. Adding a defensive short-circuit would be an unreachable branch, which the module's 100% line-and-branch JaCoCo gate rejects; the invariant is now noted in `covers()`.

**Doc heading (Minor) — done.** Renamed to *Generic records: concrete, threaded, and element-mapped* and updated every anchor consumer (the two in-page links and the release-note link); no stale `#generic-records-concrete-instantiations-and-threaded-specs` remains.

**Nitpicks:** reworded the nested-null test comment (the null is on the outer `id`, not the nested wire); narrowed `nearMissUseSitesRefuseToUnify` to `wireSideContainerMismatchRefusesToUnify` since it pins one case (the other refusals have their own tests). I left the `of()` argument order on `getAllMembers` encounter order: `abstractLeaves` deduplicates by name and the sweep reserves the exact signature, but more to the point the whole feature already relies on that iteration order for renames, derived fields and the #654 collision sweep, so pinning `of()` alone to source order would be inconsistent — worth a separate sweep if the order guarantee is ever a concern. I also skipped the fixture hoist: the three `Report` shapes aren't all identical (the near-miss and unbound tests use different ones), and the fixtures read fine locally.

`MappingProcessor`/`MergeProcessor` hold 100% line and branch; full multi-module build passes.

MagnusSmith:
Tightened all five refusal-path assertions and squashed the branch to a single commit.

Each refusal now asserts the **field-qualified** what-line (`target field 'ReportDto.results' has no usable source`) rather than the bare substring, so the test pins the exact component that refused:
- **Wire-side mismatch** additionally asserts the diagnostic names the wrong container (`com.example.OtherDto`), so it pins the wire-side specifically.
- **Inconsistent binding** gains a positive companion: the same `MirrorMapping<T>` *does* resolve a homogeneous `Mirror<String, String>` (emitting `MirrorMappingImpl.<String>instance()`), which is what proves the refusal is the inconsistent `T` binding and not an unmappable shape.
- **Unbound-variable** and **wildcard** are now two separate tests (`unboundSpecVariableRefusesToUnify`, `wildcardArgumentRefusesToUnify`), each field-qualified, over shared `REPORT`/`REPORT_DTO`/`REPORT_MAPPING` fixtures.
- **Raw-container and raw-argument** paths each assert the field-qualified line; the unresolved-argument leg keeps its `Missing` check (javac's own error).
- **Array** failing leg is now as specific as the passing leg.

`MappingProcessor`/`MergeProcessor` hold 100% line and branch; full build passes.

PR #668, opened 2026-07-26
@GenerateMapping: golden Impl files per emission tier (#655)

### LINKED ISSUE #655: [TEST] @GenerateMapping: golden Impl files per emission tier
> **Scope note:** purely internal test infrastructure — no API change. The open question is which tiers get a golden first (all six below, or the three highest-churn: full fallible, patch, sparse update).

## The gap

`@GenerateMapping`'s emitted Impls are pinned only by `contains()` fragments in `MappingProcessorTest`. The Accumulator, Assembly and lens/prism/traversal generators all have golden-file tests (`AccumulatorGoldenFileTest`, `AssemblyGoldenFileTest`, `GoldenFileTest`, resources under `hkj-processor/src/test/resources/golden/`), regenerable with `-DupdateGolden=true`; the mapping tiers have none. Claims like "identity projections stay byte-identical" (#625) currently rest on emission code being textually untouched, verified by hand during review rather than by the build.

## What "good" looks like (developer perspective)

One golden Impl per mapping emission tier, following the Assembly precedent (normalised comparison, `-DupdateGolden=true` regeneration):

- lossless full mapping (`asIso` present)
- fallible full mapping (leaf + nested spec + `List`/`Optional`/`Map` lifting + derived field, `ifPresent`-guarded bean variant included)
- identity projection (`asLens`)
- leaf-carrying projection (validated `patch`, #625)
- sparse update (`updateFrom`, #645)
- sealed dispatch

A refactor that drifts any tier's output fails the build with a readable diff instead of shipping silently.

- **Fixture reuse.** The golden sources should mirror the existing `MappingProcessorTest` fixtures rather than inventing new shapes, so a fragment assertion and its golden agree about what "the" canonical spec looks like.
- **Normalisation.** Match `GoldenFileTest`'s existing normalisation (line endings, trailing whitespace) — nothing cleverer; byte-level fidelity is the point.
- **Coverage interaction.** Golden tests execute the same emission paths the JaCoCo-pinned suite already covers; no rule changes expected.
- **Not a replacement.** Fragment assertions stay — they document intent per feature; goldens catch everything else.

## Dependencies / scope

- **Depends on** nothing new; the golden harness and regen flag exist.
- **In:** golden Impl files + a `MappingGoldenFileTest` for the tiers above.
- **Out:** goldens for diagnostics text (pinned by `hadErrorContaining`), MergeProcessor/AssemblyProcessor (Assembly already has goldens; Merge could follow the same pattern later).

MagnusSmith:
Closes #655.

## What

`@GenerateMapping`'s emitted Impls were pinned only by `contains()` fragments in `MappingProcessorTest`. Claims like "identity projections stay byte-identical" rested on the emission code being textually untouched, verified by hand at review. This adds one **golden Impl per emission tier**, following the Assembly precedent (normalised comparison, regenerable with `./gradlew :hkj-processor:updateGoldenFiles`), so a refactor that drifts any tier's output fails the build with a readable diff instead of shipping silently.

Held for last on purpose: every source of emission churn — the `hkj$ifPresent` null guard, guarded merge legs, index-located container failures, and the generic `instance()`/`of()` factories — is now landed, so the captured text is stable.

## Tiers covered

All six, plus the bean variant, in `MappingGoldenFileTest` (one package per tier, mirroring the canonical `MappingProcessorTest` fixtures):

| Golden | Tier |
|---|---|
| `PersonMappingImpl` | lossless full — `asIso` present |
| `AccountMappingImpl` | fallible full, record wire — leaf + nested spec + `List`/`Optional`/`Map` lifting + derived field |
| `BeanUserMappingImpl` | fallible full, bean wire — `ifPresent`-guarded getters, setter build |
| `EmployeeCardMappingImpl` | identity projection — `asLens` |
| `SubscriberPatchMappingImpl` | leaf-carrying projection — validated `patch` |
| `UserPatchMappingImpl` | sparse update — `updateFrom` |
| `PaymentMappingImpl` | sealed dispatch |

## Notes

- **Fragment assertions stay** — they document intent per feature; the goldens catch everything else.
- Normalisation matches the existing `GoldenFileTest` (line endings, trailing whitespace); byte-level fidelity is the point.
- Golden tests execute the same emission paths the JaCoCo-pinned suite already covers, so `MappingProcessor`/`MergeProcessor` hold 100% line and branch; full multi-module build passes.
- Out of scope (per the issue): diagnostics text (pinned by `hadErrorContaining`) and a Merge golden (could follow the same pattern later).

## Summary by CodeRabbit

* **Tests**
  * Expanded golden-file coverage for generated mapping implementations, including bidirectional conversion with validation, patching, projection mappings (lenses), sealed-type dispatch, and singleton adapter support (validated prisms / isomorphisms).
  * Improved test reliability by fail-fast asserting compilation succeeds before attempting to locate and compare generated sources.
  * Enhanced golden comparisons with line-ending normalization and trailing-whitespace tolerance, while preserving the ability to refresh golden outputs when enabled.


MagnusSmith:
Valid finding, applied in 41a6ed386 — and swept across the whole golden harness, not just this file.

**The mechanism holds:** `compilation.generatedSourceFile(...)` returns the emitted source regardless of whether the overall javac run succeeded (an Impl can be emitted and *then* fail to compile). So without a success check, a regression that makes a fixture or the emitted code uncompilable would slip through — and `updateGolden` would write the broken output back into resources as the new reference. That's exactly the class of drift these goldens exist to catch, so the harness must not be blind to it.

**Consistency sweep:** the original `GoldenFileTest` already guards on `Compilation.Status.SUCCESS`; the three later golden tests (`AssemblyGoldenFileTest`, `AccumulatorGoldenFileTest`, `ForComprehensionGoldenFileTest`) had dropped it, and this new mapping test mirrored them. I restored the same guard in all four so the latent gap is closed everywhere, not just where it was flagged:

```java
assertThat(compilation.status())
    .as("Compilation should succeed for %s", testCase.generatedClassName())
    .isEqualTo(Compilation.Status.SUCCESS);
```

All 54 golden tests across the five classes pass; full multi-module build green.

MagnusSmith:
This one is a **false positive**, driven by a stale comment I've now corrected. The compilation does *not* fail.

**Evidence:** `ForComprehensionGoldenFileTest` runs **26 tests, 0 failures, 0 errors** with the `Compilation.Status.SUCCESS` assertion in place. If the fixture failed to compile, that assertion would throw and every case would fail — they all pass, which proves the status is `SUCCESS`.

**Why:** the class comment claimed "the overall compilation will not succeed because generated code references types from `hkj-core` which is not on the test classpath." That's stale — `hkj-core` is a test dependency of `hkj-processor`, so it's on the classpath compile-testing uses, and the generated code compiles cleanly. The comment (and a matching inline one at the `generatedSourceFile` read) predates that and was never updated, so the fixture wasn't actually an expected-failure setup; the guard simply confirms what already happens.

I've corrected both comments to reflect reality (the golden pins the generated source text; the fixture compiles, so the status is asserted before comparison). Keeping the guard is the right call — it's consistent with the original `GoldenFileTest` and the three siblings, and it's exactly what catches a genuinely-broken emission regression.

All 54 golden tests across the five classes pass; full build green. Squashed to a single commit.

PR #669, opened 2026-07-27
Ergonomics and coherence hardening for the mapper

MagnusSmith:
## Review changes for users

**1. A null inside a copied list or map is now caught and reported.**
Before: if a JSON body contained `{"tags": ["a", null]}` and the `tags` list was copied as-is, the null slipped into your domain object and caused a crash later, far from where it entered. Now: parsing reports it straight away as a normal validation error, with the exact position (`tags.1: must not be null`), alongside every other problem in the payload. Lists and maps are still copied by reference, and nothing else about the mapping changes.

**2. A misspelt validator method is now a compile error instead of doing nothing.**
Before: if you wrote a validator for a field but misspelt its name (`emial()` instead of `email()`), the mapper silently ignored it, and that field was silently no longer validated. Now: the build fails with a helpful message, including a suggestion ("Did you mean 'email()'?"). Shared validator interfaces are not affected; helper methods just need to be `private` or `static`.

**3. Clearer names and simpler declarations.**
- `UpdateSpec` (the REST PATCH feature) is no longer a subtype of `MappingSpec`. They looked related but promise different things, which was misleading.
- The type parameters on both interfaces are now called `<D, W>` (Domain, Wire) instead of `<A, B>`, so your IDE tells you which is which.
- The word "Patch" now only appears in the REST PATCH feature. The other write-back method (`patch`) keeps its name, but examples and documentation no longer use "Patch" in class names for it, so the two are harder to mix up.
- If you write a method like `UserDto toDto(User user)` on a spec (a common habit from other mapping tools), the error now tells you plainly: delete it, the generated class already has `build` and `parse`.

**4. Documentation is more accurate and easier to follow.**
The release notes for 0.4.9 are rewritten as one story instead of nine overlapping entries. The front page and README now describe everything the mapper can do. Several wrong or confusing statements were fixed, including a page that promised the wrong HTTP status code, and guidance was added on how error paths are named when a field is renamed.

## Summary by CodeRabbit

* **New Features**
  * Record-to-wire mappings now support clearer bidirectional workflows, including validated dense patches and sparse updates.
  * Null values, elements, and map entries are reported as located field errors instead of causing exceptions.
  * Validation errors identify field paths and accumulate into a single response.
  * Spring integration documents consolidated HTTP 422 responses for mapping validation failures, with configurable status handling.

* **Documentation**
  * Expanded mapping, generic record, sealed hierarchy, PATCH, error-path, and Spring integration guidance.
  * Added clarification of mapping laws, supported wire shapes, and migration considerations.


MagnusSmith:
All five review comments addressed in a292b7d27 (branch also rebased onto main):

**Major — controller docs vs the actual `updateFrom` contract: fixed.** Correct catch: `updateFrom` returns `Edits.Accumulated`, which the return-value handler does not render. Both the README sentence and the home.md bullet now say a `parse` or `patch` result renders directly while an `updateFrom` update is applied first via `.applyPath(current)`, and both describe the status as `hkj.web.validation-field-error-status` with 422 as the default rather than a hard-coded 422.

**Minor — "law-checked at compile time": fixed** in both files to "law-checked by the build's test suite"; the laws are runtime assertions, the wording overclaimed.

**Nitpicks, all applied:** the two remaining "Applies the patch" javadocs on `Edits.Accumulated#apply`/`applyPath` now say "Applies the update", completing the terminology migration; the duplicated List/Map classification is extracted into a shared `MappingProcessor.isExactly(type, fqcn)` helper using the file's `getQualifiedName().contentEquals` convention, with `MergeProcessor.containerKind` delegating to it (same sharing pattern as the emitted guard helpers); and `hkj$allPresent` now iterates with a counter instead of `values.get(i)`, so validation stays linear on non-random-access lists — generator changed, golden regenerated.

Full build green; `MappingProcessor`/`MergeProcessor` hold 100% line and branch coverage. The PR description is also rewritten in plain English per maintainer request.

MagnusSmith:
All five addressed in the amended commit:

**Minor — total-parse law fixture vs identity containers: fixed, and the underlying doc claim corrected.** The catch was right, and it exposed a subtly false sentence: the javadoc claimed `build` "only ever produces such wires", but `build` copies identity containers verbatim, so a domain sample carrying a null element produces exactly the excluded wire. A generic `<D, W>` helper cannot introspect the sample, so the fix is the documented fixture contract, stated in both the class-list bullet and the method javadoc: `domainSample` must not carry a null reference component or null identity-container element, or the round trip lands on the excluded wire and the law (correctly) fails.

**Nitpicks, all applied:** the stranded `hkj$ifPresent` javadoc moved down to `ifPresentHelper()` (same stacking mistake as an earlier round, same fix); `isGetterShaped` deleted in favour of the pre-existing `isDerivedCandidate` (identical body — the derived-field shape now has a single definition); the `OBJECTS` constant reused in `valuesPresentHelper`; and `MergeProcessor.containerKind` is now `static` with the imported `TypeMirror`.

Full clean build passes; `MappingProcessor`/`MergeProcessor` hold 100% line and branch.

Issue #678, opened 2026-07-28
[DOCS/FEAT] @GenerateMapping: the injection and test-double story


MagnusSmith:
_Suggested target: docs half 0.4.9; evaluation half 0.4.10_

## The gap

A generated mapping is reached through statics (`INSTANCE`, `instance()`, `of(...)`), and the spec interface deliberately declares no `build`/`parse` — it is a declaration, not a service. So `@Autowired UserMapping` injects nothing useful, and there is no obvious seam for a test double. For teams whose house style is constructor injection and slice tests, that reads as a missing story.

The answer largely exists already: `asValidatedPrism()` *is* the interface form of a parse-capable mapping (`ValidatedPrism<Wire, Domain>` injects, composes and fakes trivially), and a stateless total function rarely needs mocking in the first place. But none of it is written anywhere, so an evaluator concludes the capability is absent rather than undocumented.

## What "good" looks like (user perspective)

A short book section (and an hkj-spring note) answering the three questions directly:

- **Injecting:** register the surface you consume as a bean — `@Bean ValidatedPrism<UserDto, User> userCodec() { return UserMappingImpl.INSTANCE.asValidatedPrism(); }` — and depend on `ValidatedPrism<UserDto, User>`; for build-only use, `Function<User, UserDto>` via `UserMappingImpl.INSTANCE::build`.
- **Faking:** `ValidatedPrism.of(...)` builds a stub codec in two lines; no mocking framework needed.
- **When not to bother:** a generated mapping is a pure static function; injecting it buys substitution, not lifecycle — teams comfortable calling `INSTANCE` directly lose nothing.

Plus an evaluation (not necessarily an implementation) of whether generated Spring wiring is worth having at all, given the mapping-registry proposal (unposted internal draft) would provide it wholesale.

## Things to explore / constraints

- **Doc-first.** The documentation half needs no code and could ride the current release.
- **Optional emission?** A component-model-style knob (`@GenerateMapping(spring = true)` emitting an `@Configuration` contributing the prism bean) is deliberately NOT proposed here without the registry decision; two wiring mechanisms would be worse than none.
- **`of(...)` Impls.** Element-mapped mappings are values, not singletons; the doc should show constructing them once in a `@Bean` method as the natural pattern.

## Dependencies / scope

[...]

PR #683, opened 2026-07-30
Document the injection and test-double story for generated mappings (#678)

MagnusSmith:
Closes the docs half of #678 (Target 0.4.9), with the evaluation half's recommendation posted on the issue.

## What

Documents the injection and test-double story for generated mappings, which existed but was written nowhere, so a Spring-idiom evaluator concluded the capability was absent.

- **New book section, "Injecting and testing generated mappings"**: register the surface you consume, charted per tier — `ValidatedPrism<Wire, Domain>` for parse-capable mappings, `Function`/`BiFunction` method references for build-only, `patch` and `updateFrom` (the review on the issue found the prism-only story covered half the example app's real usage). Includes the Spring generics-resolution note, the same-pair `@Qualifier` caveat, and the element-mapped `of(...)` construct-once pattern.
- **Fakes are values, not mocks**: `ValidatedPrism` is sealed, so it cannot be hand-implemented and Mockito cannot mock it. The docs lead with the two-line `ValidatedPrism.of(...)` fake, turning a likely first-contact failure into the recommended pattern.
- **End-to-end demonstration in the example app**: `MappingConfiguration` registers the codec bean; `UserController`'s parse endpoint now injects it, while the PATCH endpoint deliberately keeps the direct `INSTANCE` call so both idioms sit side by side; a new `UserParseFakeCodecSliceTest` substitutes a rejecting fake in a `@WebMvcTest` slice and asserts the located 422 it produces.
- The Spring integration page cross-links the section; the mapping skill carries the summary.

All 100 example-module tests pass (three existing slice contexts gained the config import); full multi-module build green.

## Summary by CodeRabbit

- **Documentation**
  - Added guidance for injecting and testing generated record mappings with Spring.
  - Documented supported mapping surfaces, fake codec creation, and generic type resolution.
  - Added examples for integrating mappings with parse and PATCH endpoints.

- **New Features**
  - Registered the user mapping as an injectable Spring codec.
  - Updated the user parsing endpoint to use the injected codec.

- **Tests**
  - Added coverage for substituting a fake codec and returning validation errors.
  - Updated MVC slice tests to include the mapping configuration.

### DIFF
--- a/.claude/skills/hkj-mapping/SKILL.md
+++ b/.claude/skills/hkj-mapping/SKILL.md
@@ -182,12 +182,15 @@ directly. Diagnostics about inherited members name the declaring interface.
 ### Injecting and faking (Spring)
 
 Register the surface you consume as a bean, not the spec or Impl: `ValidatedPrism<Wire, Domain>`
-via `Impl.INSTANCE.asValidatedPrism()` for parse-capable mappings; `Function`/`BiFunction` method
-references for build-only, `patch` and `updateFrom`. `ValidatedPrism` is SEALED: fakes are built
-as values with `ValidatedPrism.of(...)`, never mocked (Mockito rejects sealed types). Spring
-resolves the generic type, so per-pair codecs coexist; same-pair duplicates need `@Qualifier`.
-Injection is optional: an Impl is a stateless pure function, and calling `INSTANCE` directly
-loses nothing but the substitution seam.
+via `Impl.INSTANCE.asValidatedPrism()` (concrete; threaded specs use `instance()`, element-mapped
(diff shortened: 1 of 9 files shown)


MagnusSmith:
Applied. The blanket claim was wrong for the element-mapped tier exactly as stated: both the book section's opening and the skill paragraph now distinguish concrete/threaded Impls (stateless, reached through `INSTANCE`/`instance()`) from element-mapped ones (immutable values built by `of(...)`, carrying their element prisms, constructed once in the `@Bean` method), and the closing "loses nothing" sentence names all three access shapes rather than `INSTANCE` alone.

PR #684, opened 2026-07-31
Register incremental annotation processing; pin Lombok interop (#677)

### LINKED ISSUE #677: [BUILD] hkj-processor: Gradle incremental annotation processing registration + Lombok interop coverage
## The gap

Two consumer-build concerns, both cheap relative to their adoption impact:

1. **Incremental annotation processing.** Gradle disables incremental compilation for any source set whose processors are unregistered. hkj-processor ships no `META-INF/gradle/incremental.annotation.processors`, so every consuming module pays full recompiles. MapStruct has registered (isolating) since 1.4 and advertises it. The mapping registry is genuinely aggregating (same-round cross-spec resolution), which is registrable as `AGGREGATING`; per-processor classification needs an audit (the optics generators may be `ISOLATING`).
2. **Lombok.** Bean wires read getters/setters, which Lombok generates in the same javac run; processor ordering is exactly where MapStruct needed a dedicated `lombok-mapstruct-binding`. Whether `@Data`/`@Getter` wire classes work with `@GenerateMapping` today is unknown: no test, no documentation either way.

- A consuming Gradle module keeps incremental compilation with hkj-processor on the processor path, and a note in the getting-started docs says so.
- A Lombok `@Data` class used as a bean wire either works (with a passing test and a documented note) or is diagnosed/documented as unsupported with the reason.

- **Classification audit.** One line per processor in the registration file; MappingProcessor/MergeProcessor/AssemblyProcessor are aggregating (registries), the per-type optics generators are likely isolating. Wrong classifications cause silent stale-output bugs, so each needs a stated justification and, ideally, an incremental-build test.
- **Lombok test setup.** Adding Lombok to the processor-path of a test fixture compilation (compile-testing supports multiple processors) or a small integration module; decide whether the repo wants the dependency at test scope.
- **Documentation.** A short "build setup" subsection: Gradle/Maven processor-path snippets, incremental note, Lombok note.

## Dependencies / scope

- Depends on: nothing.
- In: registration file with audited classifications, an incremental-compilation smoke test if practical, Lombok bean-wire fixture test plus documentation of the outcome, build-setup docs.
- Out: IDE plugin work, Maven-specific tooling beyond the doc snippet.

MagnusSmith:
Closes #677.

## What

Two consumer-build concerns, both verified before implementing:

**Incremental annotation processing registered.** Gradle disables incremental compilation for any source set whose processors are unregistered, and none of the twenty HKJ processors were (verified: no `META-INF/gradle/incremental.annotation.processors` anywhere; 19 processors in `hkj-processor`, 1 in `hkj-spring:client-processor`). Both modules now ship the registration file, packaged into the jar.

Every processor is registered as **AGGREGATING**, with the justification stated in the file itself: the mapping family genuinely aggregates (same-round registries drive nested and sealed resolution, so outputs may depend on any annotated element), and for the per-type optics generators a wrong ISOLATING claim produces *silently stale outputs* while AGGREGATING is behaviour-preserving and already restores incrementality for consuming source sets. Refining individual processors to ISOLATING needs a per-processor one-origin-per-file audit first — deliberately left as follow-on. The audit also confirmed no disqualifiers: no processor does resource I/O, and the generators register originating elements.

**Lombok interop pinned, not assumed.** `LombokInteropTest` compiles a `@Data` class as a bean-shaped `@GenerateMapping` wire with both processors in one javac run (Lombok 1.18.42, JDK 25) and asserts the mapping generates against the Lombok-materialised accessors — the exact pairing where MapStruct needed a dedicated binding artefact. **It passes as-is**: no binding, no ordering configuration. Lombok is a test-scope dependency only.

**Docs**: the manual-setup page gains an "Incremental compilation" note (including the caveat that an unregistered third-party processor on the same path disables incrementality for the whole source set — the first thing to audit if compile times regress) and a "Lombok" section with the processor-path snippet.

Out of scope, per the issue: IDE plugin work; a Gradle TestKit incremental-build smoke test (heavy machinery for what the registration file plus Gradle's own behaviour guarantees; can follow if wanted).

Full multi-module build passes.

## Summary by CodeRabbit

- **Documentation**
  - Added guidance on incremental annotation processing and Lombok interoperability.
  - Documented required Gradle processor and compile-only dependencies.

- **Performance**
  - Enabled Gradle incremental processing for optics and HTTP client annotation processors.

- **Compatibility**
  - Improved interoperability with Lombok-generated data classes.
  - Added validation for processor ordering and clear diagnostics when mappings are generated before Lombok accessors.

### DIFF
--- a/hkj-book/src/tooling/manual_setup.md
+++ b/hkj-book/src/tooling/manual_setup.md
@@ -138,5 +138,23 @@ The [HKJ build plugin](gradle_plugin.md) handles all of the above (dependencies,
 
 ---
 
+## Incremental compilation
+
+The HKJ annotation processors register with Gradle's incremental annotation processing (as aggregating processors: cross-spec features such as nested mapping resolution may read any annotated element in the compilation). A source set using them keeps incremental compilation; no configuration is needed. Unregistered third-party processors on the same processor path disable incrementality for the whole source set, so if compile times regress, audit the other entries on the path first.
+
+## Lombok
+
+Lombok and the HKJ processors run in the same javac invocation, and the pairing is covered by a test in the HKJ build: a `@Data` class works as a bean-shaped `@GenerateMapping` wire, with the generated getters and setters visible to the bean analyser. **Order matters: list Lombok before `hkj-processor`** (within a javac round, processors run in listed order, and the bean analyser needs the accessors already materialised; the reverse order fails with a clear diagnostic). No binding artefact is needed beyond that:
+
+```gradle
+dependencies {
+    annotationProcessor("org.projectlombok:lombok:LOMBOK_VERSION")
+    annotationProcessor("io.github.higher-kinded-j:hkj-processor:LATEST_VERSION")
+    compileOnly("org.projectlombok:lombok:LOMBOK_VERSION")
+}
+```
+
+---
+
 **Previous:** [Build Plugins](gradle_plugin.md)
 **Next:** [Compile-Time Checks](compile_checks.md)
--- a/hkj-processor/build.gradle.kts
+++ b/hkj-processor/build.gradle.kts
@@ -29,6 +29,9 @@ dependencies {
     testImplementation(platform(libs.junit.bom))
     testImplementation(libs.junit.jupiter)
     testRuntimeOnly(libs.junit.platform.launcher)
+    // Lombok interop coverage: bean-shaped wires read Lombok-generated accessors in the same
+    // javac run, so the pairing is pinned by LombokInteropTest rather than assumed.
+    testImplementation("org.projectlombok:lombok:1.18.42")
     testImplementation(libs.assertj.core)
     testImplementation(libs.archunit.junit5)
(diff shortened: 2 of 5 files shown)


MagnusSmith:
Applied, and the strengthened test earned its keep: it exposed a real constraint the filename check masked. **Order matters.** With Lombok listed first, everything works — now asserted properly (the generated Impl reads `wire.getName()`/writes `wire.setName(...)` through the Lombok-materialised accessors, plus a full runtime round trip through the Lombok-compiled bean). With `MappingProcessor` first, the bean analyser runs before Lombok has materialised the accessors within the round, and the wire fails with the honest `is not a usable bean-shaped wire` diagnostic — so rather than claiming order independence, the test pins both directions (the working order end to end, the broken order by its diagnostic), and the manual-setup page now states the Lombok-first requirement explicitly instead of the previous "no ordering configuration required" claim, which this test disproved.

PR #688, opened 2026-08-07
@GenerateMapping: lift the 16-component parse ceiling (#670)

### LINKED ISSUE #670: [FEAT] @GenerateMapping: lift the 16-component parse ceiling
## The gap

`parse` and `patch` assemble through `Validated.fields()`, which is capped at 16 components (`ArityCeilings.ASSEMBLY`). The diagnostic's fix is "group related components into nested records", which only works when the wire nests too: an externally fixed flat 20-30 field DTO (the single most common MapStruct workload) cannot be grouped on the domain side, because flat wire fields cannot fill a nested domain component. Such wires are unmappable today — and the refusal is total: the diagnostic kills the whole
mapping, so even the ceiling-free `build` direction is lost (comparator probe, verbatim:
"'Account' has 20 components; the accumulating parse supports at most 16 ... Group related
components into nested records ... or map the record by hand"). MapStruct has no ceiling. Meanwhile `@GenerateAssembly` in this same repo already implements uncapped accumulating assembly via the curried-ap merge, so the machinery exists next door.

Nothing: a 30-component record-record spec simply compiles, `parse` reports every bad field located as before, and the ceiling paragraph disappears from the book's limits section. No API change of any kind.

- **Mechanism.** Either emit chunked `fields()` ladders combined applicatively (16 at a time, results merged into the canonical constructor call), or reuse the `@GenerateAssembly` curried-ap emission shape directly inside the mapping processor. The generated code stays readable either way; goldens pin it.
- **Error semantics unchanged.** Accumulation order stays declaration order; located paths unchanged; the null doctrine unchanged. Laws must show a >16 mapping behaves identically to a <16 one.
- **All parse-capable tiers.** Full record and bean wires, the validated `patch` (wire arity), and `@GenerateMerge`'s fallible target (also capped at 16 today) should lift together; `updateFrom` is already uncapped (varargs fold).
- **Ceiling remnants.** If any genuine limit remains (e.g. constructor parameter count is JVM-bounded at 255), document the real number instead.

## Dependencies / scope

- Depends on: nothing; internal codegen change.
- In: emission change across the parse-capable tiers plus fallible merge, a wide (e.g. 20-component) golden per affected tier, law coverage for a wide mapping, book/limits and diagnostic updates.
- Out: any public API change.

MagnusSmith:
Closes #670.

## What

`@GenerateMapping`'s `parse` and validated `patch`, and `@GenerateMerge`'s fallible merge, now assemble records wider than 16 components. An externally fixed flat 20-30 field DTO (the single most common MapStruct workload) maps without grouping components into nested records, and the old total refusal (the ceiling diagnostic killed the whole mapping, including the ceiling-free `build`) is gone. No public API change; the hand-written `Validated.fields()` ladder keeps its 16-field arity.

## Mechanism

A new package-private `ChunkedAssembly` emitter splits the legs into `Validated.fields()` ladders of at most 16, applies each to a `TupleN` (a singleton trailing ladder applies the identity), and combines the chunks with `ap` + `NonEmptyList.semigroup()` in ladder order. The ladder's own `and()` is internally that same merge, so labels, located paths and declaration-order accumulation are exactly those of a single ladder — by construction, and law-checked at width.

The mechanism was settled by a measured POC ([results on the issue](https://github.com/higher-kinded-j/higher-kinded-j/issues/670#issuecomment-5222043976)): the inline exact-arity curried-ap alternative (the `@GenerateAssembly` shape) is superquadratic in javac inference time and overflows the compiler's default stack near 200 components, while chunked ladders stay linear to the JVM's constructor bound. That rationale is now recorded in the `ChunkedAssembly`/`AssemblyProcessor` javadoc.

The ≤16 emission paths are byte-identical to before: regenerating every golden changed nothing except adding the new wide ones.

The only remaining width bound is the JVM's constructor parameter-slot limit on the record itself (254 components in practice, fewer with `long`/`double`), which javac enforces at the record declaration with a single clean error (verified with a 300-component probe — no diagnostic cascade).

## Review pass

The pre-PR five-lens panel (correctness, type safety, immutability/null doctrine, ergonomics, HKJ architecture) confirmed the mechanics and found one real defect, fixed in the second commit: the chunked merge emitted fixed identifiers (`c1..`, `t1..`, the singleton `v`) into a method whose parameter names the spec author chooses, so a merge parameter named `c1`/`t1`/`v` failed to compile (JLS 6.4). `ChunkedAssembly` now underscore-suffixes its identifiers until free of the enclosing method's parameter names, mirroring the patch tier's existing rename dance, with a hostile-name regression test.

## Coverage

- Wide goldens for the record-wire (20 components, 16+4), bean-wire (17, 16+singleton) and patch (17-of-18) tiers; the merge tier has no golden infrastructure, so its chunked shape is pinned by fragment assertions and covered transitively through the shared emitter (noted in the test).
- The three former ceiling-diagnostic tests inverted into chunked-success tests, plus a pin that a 16-component record still parses in one un-chunked ladder.
- `WideMappingLawsTest` (hkj-examples): a 20-component mapping obeys `MappingLaws`, reports located errors in declaration order across the chunk boundary, and round-trips; the declaration-order test is included on the book's Record Mapping page.
- Book: limits section restructured (no-ceiling lead with the wide-laws snippet, genuine limits below), release-history entry under 0.4.10-SNAPSHOT, `hkj-mapping` skill aligned.

## Summary by CodeRabbit

- **New Features**
  - Generated mappings now support records with more than 16 components through chunked validation assembly.
  - Validated patching and fallible merges support wider records while preserving field-error locations and declaration order.
  - Added a wide-record mapping example covering 20-component records and email validation.
- **Documentation**
  - Clarified supported mapping widths, practical JVM constructor limits, and remaining limits for manually written mapping ladders.
- **Tests**
  - Added coverage for wide mappings, patching, merging, validation errors, and round-trip behavior.


MagnusSmith:
Addressed the review nitpicks:

- **ChunkedAssembly chunk loop**: already restructured to the proposed `subList` slice in 0717d2ee, which raced the review — the inner-loop mutation of `leg` is gone.
- **Duplicated Wide fixture**: hoisted into a shared `WIDE` field on `Diagnostics` with a javadoc noting both chunked-merge tests must cross the 17-component ladder boundary together (a203e0fb).
- **Golden suite compilation**: `MappingGoldenFileTest` now compiles the fixture set once in `@BeforeAll`; all ten parameterised cases read from the cached `Compilation` (a203e0fb). The suite drops from ten javac runs to one (~1.3 s total locally).

PR #689, opened 2026-08-08
@GenerateMapping: a stock codec vocabulary of standard ValidatedPrism leaves (#671)

### LINKED ISSUE #671: [FEAT] @GenerateMapping: a stock codec vocabulary of standard ValidatedPrism leaves
## The gap

MapStruct ships roughly twenty built-in conversion families (String↔numbers with `numberFormat`, String↔enum, all `java.time` types with `dateFormat`, `BigDecimal`, `UUID`, `URI`, `Currency`, `Locale`, ...) at zero user cost. HKJ ships **none**: every type-differing pair needs a hand-written `ValidatedPrism` leaf, and no stock codec utility exists in hkj-core (`ValidatedPrism` offers only the `of`/`fromIso`/`fromPrism` factories; the book's `EmailCodecs` is example code). A typical DTO boundary (dates, enums, money, IDs) therefore starts with authoring a project codec vocabulary before the first spec compiles.

The flip side is the sell: HKJ's fallible conversions return located, accumulated `FieldError`s where MapStruct's throw `DateTimeParseException` mid-mapping, first failure only. A stock vocabulary makes that semantics near-free instead of hand-rolled.

```java
import static org.higherkindedj.optics.validated.StandardCodecs.*;

@GenerateMapping
public interface OrderMapping extends MappingSpec<Order, OrderDto> {
  default ValidatedPrism<String, LocalDate> placedOn() { return localDate(ISO_DATE); }
  default ValidatedPrism<String, OrderStatus> status() { return enumByName(OrderStatus.class); }
  default ValidatedPrism<String, BigDecimal> total()   { return bigDecimal(); }
  default ValidatedPrism<String, UUID> id()            { return uuid(); }
}
```

- One factory per family: `uuid()`, `uri()`, `localDate(fmt)`, `instant()`, `offsetDateTime(fmt)`, `enumByName(Class)` (error message lists the valid constants), `bigDecimal()`, `intFromString()`, `longFromString()`, `doubleFromString()`, `booleanStrict()`, `currency()`, `locale()`.
- Every parse failure is a located `FieldError` with a copy-worthy message (`not an ISO-8601 date (expected e.g. 2026-07-28)`), so the codecs feed the 422 leg unchanged.
- Outbound (`build`) directions are total formatting, matching the leaf contract.
- Shareable across specs by ordinary means (a project mix-in interface delegating to the stock factories).

- **Where it lives.** `org.higherkindedj.optics.validated.StandardCodecs` in hkj-core beside `ValidatedPrism`, or a `codecs` subpackage; static factories returning cached instances where the codec is parameterless.
- **Message consistency.** Reuse the doctrine grammar (`must not be null` precedent): terse, located, actionable; enum errors should name the permitted constants.
- **Format parameterisation.** `localDate(DateTimeFormatter)` plus a no-arg ISO default; decide whether pattern-string overloads (`localDate("dd/MM/yyyy")`) are worth the parse-at-startup cost.
- **Explicitly out: defaults.** A `withDefault(x)` combinator cannot work as a leaf (the null guard short-circuits before the prism sees anything) and "absent means default" contradicts the null doctrine; if defaulting is ever wanted it is a separate per-component policy discussion, not a codec.
- **Explicitly out: implicit application.** Codecs stay explicit leaves; the processor never silently picks a conversion (that implicitness is MapStruct's footgun, not a feature to copy).

## Dependencies / scope

- Depends on nothing new; purely additive hkj-core API plus book section ("Standard codecs") and examples.
- In: the factory class, located-error messages, law tests per codec (round-trip on valid samples via `ValidatedPrism` laws), book/docs.
- Out: automatic/implicit conversion selection; default-value semantics; user-extensible codec registries.

MagnusSmith:
Closes #671.

## What

`StandardCodecs` in `org.higherkindedj.optics.validated`, beside `ValidatedPrism`: one stock factory per standard conversion family, so a typical DTO boundary (identifiers, dates, enums, money) maps without hand-written leaves — `uuid`, `uri`, `localDate()` + `localDate(DateTimeFormatter)`, `instant`, `offsetDateTime()` + formatter overload, `enumByName` (the error names the permitted constants), `bigDecimal`, `intFromString`/`longFromString`/`doubleFromString`, `booleanStrict`, `currency`, `locale`. Parameterless factories return cached instances; formatter overloads render a sample into the error message eagerly. Every failure is one terse, located, copy-worthy `FieldError` in the doctrine grammar, so the codecs feed the 422 leg unchanged. Per the issue's outs: no defaults, no implicit application, no registry.

## Design centre: canonical forms only

The `ValidatedPrism` build-parse section law (`parse(s) == Valid(a)` ⇒ `build(a) == s`, which "forbids a lossy parse-normalise") settles the design: every codec accepts exactly the form it renders, enforced per value by a shared guard (parse via the throwing JDK parser, then require the rendering to equal the source). Case-folded UUIDs, leading zeros, scientific notation and lowercase language tags are located rejections, never silent normalisations — where MapStruct's conversions throw mid-mapping, first failure only, these accumulate. Where a wire's canonical form differs (uppercase-UUID producers, JavaScript's fixed-millis timestamps, `+00:00` UTC spellings), the formatter overloads and hand-written leaves make *that* form the canon lawfully; the book section shows each escape hatch.

## Review pass

The pre-PR five-lens panel confirmed the vocabulary's spine and caught one serious defect, fixed in the second commit: `bigDecimal().parse("1E+2147483647")` — a 14-character wire string — threw `OutOfMemoryError` from inside the guard (`toPlainString` materialises the unbounded plain form). Parse now rejects any value whose scale falls outside plain notation's `0..length` band before rendering, provably without excluding any canonical input. The same commit moves the canonical check inside the exception guard (a lenient caller formatter can no longer leak an exception on wire input), makes empty enums fail fast, corrects wire-visible messages while still SNAPSHOT-revisable (`unknown OrderStatus …`, `32-bit`/`64-bit` integer, `canonical form`/`plain notation`), documents the boxed-component restriction and the star-import self-recursion trap, steers the mapper's primitive-component diagnostic to a wrapper type instead of the uncompilable `ValidatedPrism<String, int>`, and rewrites the book's `validated_prism.md` advice that previously taught a normalising parse — the exact section-law violation the codecs forbid.

## Coverage and docs

- 30 hkj-core tests: `assertValidatedPrismLaws` per codec plus message, canonicality, caching, fail-fast and regression coverage (huge exponents, lenient formatters, empty enums, `Z`-vs-`+00:00`, fraction canonicality); `StandardCodecs` at 100% instructions and branches.
- Book: a "Standard codecs" section on the Record Mapping page with a compiled stock-codec `OrderMapping` example, its located-errors snippet, and the producer escape hatches; `MappingLaws` coverage in hkj-examples; release-history entry; cross-references from the `ValidatedPrism` page and the hkj-optics/hkj-bridge/hkj-mapping skills (the mapping skill snippet is compile-gated).

Panel follow-ups deliberately excluded as beyond #671's locked scope, to be filed separately: a public `ValidatedPrism.canonical(...)` smart constructor, and an `hasFieldErrors(...)` assertion for hkj-test.

## Summary by CodeRabbit

- **New Features**
  - Added standard validated codecs for UUIDs, URIs, dates, date-times, numbers, booleans, currencies, locales, and enums.
  - Added canonical-format validation with clear, field-level errors for invalid values.
  - Added support for custom date and date-time formatting.
  - Added a mapping example demonstrating string-to-domain conversions.

- **Bug Fixes**
  - Improved mapping diagnostics for incompatible primitive field types.

- **Documentation**
  - Expanded guidance on codec usage, canonical representations, validation behavior, and mapping round trips.
  - Added release notes for the new codec functionality.


MagnusSmith:
The two panel follow-ups named at the end of the description are now filed: #690 (public `ValidatedPrism.canonical(...)` smart constructor) and #691 (hkj-test `hasFieldErrors(...)` assertion).

MagnusSmith:
Good catch — "no zero-padding, no reformatting" was a leftover of the pre-#671 sentence and misread as a constraint on `build` alone, which `localDate()`'s zero-padded rendering lawfully contradicts. Reworded to state the law over the pair: `build(parse(s).get()) == s` for every parsing `s`; `build` defines the canonical spelling, `parse` accepts exactly it.

PR #692, opened 2026-08-09
ValidatedPrism.canonical: the section law by construction for user codecs (#690)

### LINKED ISSUE #690: [FEAT] ValidatedPrism: a public canonical smart constructor, so user codecs get the section law by construction
## The gap

Every lawful codec must honour the build-parse section law (`parse(s) == Valid(a)` implies `build(a) == s`), and the natural way to write one — wrap a throwing JDK parser, render on the way out — silently violates it whenever the parser is more lenient than the renderer (`UUID.fromString` accepts uppercase, `toString` renders lowercase). `StandardCodecs` (#671) solves this internally with a private guard: parse, then require `render(value).equals(source)`, rejecting non-canonical spellings with the located `FieldError`. Users writing their own codecs face the identical trap with no help: the guard is private, so they either copy it or ship an unlawful leaf that fails `ValidatedPrismLaws` and quietly breaks the spec's `asValidatedPrism`/`asIso` tiers. Until recently the book actively taught the violation ("keep all normalising in `parse`", corrected in #671's review pass), which suggests how easily the trap is walked into.

```java
// An uppercase-UUID wire (SQL Server): the canonical form is THEIRS, lawfully
default ValidatedPrism<String, UUID> id() {
  return ValidatedPrism.canonical(
      "not an uppercase UUID",
      UUID::fromString,                                  // throwing parse, lenient is fine
      uuid -> uuid.toString().toUpperCase(Locale.ROOT)); // render defines the canon
}
```

- `ValidatedPrism.canonical(message, parse, render)`: the throwing parse and the per-value canonical check both guarded into one located `FieldError`; `render` is the total build direction. Both laws hold by construction for any injective render.
- Generic in `S`, not just `String` — the guard is `render(value).equals(source)` and nothing more.
- `StandardCodecs` delegates its private helper to the public factory, so there is exactly one implementation of the pattern.

- **Message vs FieldError.** `String message` (matching `StandardCodecs`' grammar) or a `FieldError` parameter (matching `fromPrism`'s `reason`); possibly both overloads.
- **The injectivity caveat.** The guard cannot catch a non-injective render (a two-digit-year formatter remaps 1926 to 2026 through a parseable rendering) — the javadoc must carry the same warning `localDate(DateTimeFormatter)` now does.
- **Error breadth.** The guard catches `RuntimeException` from parse and render; whether that breadth is right for a public factory (vs naming the expected exception types) deserves a look at the `fromPrism` precedent.
- **Book/laws.** One paragraph on the `ValidatedPrism` page beside the laws section, replacing the "copy the guard" gap; `ValidatedPrismLaws` coverage on a custom-canon example (the uppercase-UUID leaf the Record Mapping page now shows hand-written would become the worked example).

## Dependencies / scope

- Depends on: #671 (merged StandardCodecs supplies the delegating call site).
- In: the public factory on `ValidatedPrism`, javadoc with the injectivity caveat, `StandardCodecs` delegation, law tests, book paragraph.
- Out: any registry, implicit application, or defaulting — unchanged from #671's outs.

MagnusSmith:
Closes #690.

## What

`ValidatedPrism.canonical(message, parse, render)`, plus a `FieldError` overload matching `fromPrism`'s reason type: the public smart constructor that gives a user codec the build-parse section law by construction. The natural way to write a codec — wrap a throwing JDK parser, render on the way out — silently violates the law whenever the parser is more lenient than the renderer (`UUID.fromString` accepts uppercase; `toString` renders lowercase). The factory guards it per value, generic in `S`:

```mermaid
flowchart LR
    S([wire source s]) --> P["parse.apply(s)"]
    P -->|throws RuntimeException| R([Invalid: located reason])
    P -->|value a| G{"render.apply(a).equals(s)?"}
    G -->|no: a spelling render cannot reproduce| R
    G -->|render throws| R
    G -->|yes| V([Valid a])
```

The uppercase-UUID wire from the issue is the worked example throughout — javadoc, hkj-core law tests, the compile-verified book fixture, and the Record Mapping page, whose previously hand-written leaf (which referenced an undefined `isUuid` helper) becomes the factory call:

```java
default ValidatedPrism<String, UUID> id() {
  return ValidatedPrism.canonical(
      "not an uppercase UUID",
      UUID::fromString,                                  // throwing parse, lenient is fine
      uuid -> uuid.toString().toUpperCase(Locale.ROOT)); // render defines the canon
}
```

`StandardCodecs`' private guard is now a one-line delegation to the public factory, so the pattern has exactly one implementation (repo-swept: no other hand-rolled copies exist). Per the issue's outs: no registry, no implicit application, no defaulting.

## The issue's three explore points, resolved

- **Message vs FieldError**: both overloads. The `FieldError` overload is not speculative — a pre-located reason (`FieldError.of(...).at("id")`) is inexpressible through the String overload, and tests pin that it is carried through unchanged.
- **Error breadth**: `RuntimeException` retained and documented as doctrine — any runtime failure inside the guard (parse on malformed input, render on a value the lenient parse produced, a null-returning parse) is the located rejection, never an exception on wire input. `Error`s still propagate; the outward `build` stays total by contract.
- **Injectivity caveat**: carried in the javadoc and book as one of two explicit obligations (below).

## Review pass

The pre-PR five-lens panel (correctness, type safety, immutability, ergonomics, architecture) found no code defects — variance matches `of`, delegation is semantics-identical to the old guard (verified against origin/main), the shared `FieldError` reason is deeply immutable, ArchUnit green, `Valid(null)` provably cannot leak. The survivors were contract-prose findings, all folded in:

- **"Lawful by construction" overstated sufficiency** (demonstrated in jshell): an injective render is necessary but not sufficient — the parse must also accept what the render produces. That coherence failure is *loud* (rejections, never a silently wrong value); non-injectivity remains the *silent* trap (a two-digit-year format re-parses 1926 as 2026). Javadoc, book, and glossary now state both obligations and point at `ValidatedPrismLaws`.
- **The guard is `equals`-based** (demonstrated with `byte[]`): an array-typed `S` degenerates to identity and rejects every parse — documented as a value-equality requirement on `S`.
- **Purity note**: the render runs on every parse, so the prism is only as thread-safe as the functions it wraps (a captured `SimpleDateFormat` would race).
- Convention fixes: explicit `message` null-guard; the `fromPrism` parenthetical names the shared reason type rather than a parameter shape it does not have.

## Coverage and docs

- Six new hkj-core tests via `ValidatedPrismLaws`: the uppercase-UUID canon (lowercase rejected as non-canonical, not normalised), a non-String source (`ValidatedPrism<Integer, Month>`), render-failure locatedness, the `FieldError` overload's carried reason, and both overloads' null contracts. `ValidatedPrism` and `StandardCodecs` at 100% instructions and branches.
- Book: a `canonical` passage beside the laws section on the ValidatedPrism page, compile-verified via anchored fixtures (`ValidatedPrismBook.UPPER_UUID` + its `ValidatedPrismLaws` test); the Record Mapping escape-hatch paragraph rewritten around the factory; glossary, release-history entry, and the hkj-optics/hkj-bridge/hkj-mapping skills swept for the new vocabulary.

## Summary by CodeRabbit

- **New Features**
  - Added canonical validation for parsing and rendering values.
  - Supports custom parsers, canonical output formats, and clear validation errors.
  - Rejects noncanonical representations that cannot be reproduced consistently.

- **Documentation**
  - Updated guidance, glossary content, examples, and release history with canonical validation usage.
  - Added an uppercase UUID example demonstrating canonical formatting.

- **Tests**
  - Added coverage for parser and renderer failures, custom errors, generic inputs, null validation, canonical forms, and validation laws.


MagnusSmith:
Addressed in 6b9017347: the three compressed surfaces (hkj-mapping skill, hkj-bridge member table, glossary) now say "every spelling the render cannot reproduce" and carry the injectivity obligation, matching the qualification the javadoc and the ValidatedPrism book page already state. (The remaining "reject every other spelling" on the laws page is the discipline asked of a lawful codec author, not a factory guarantee, so it stands.)

Issue #680, opened 2026-07-28
[DESIGN] UpdateSpec: tri-state PATCH (absent / explicit-null / value) interop


MagnusSmith:
## The gap

`updateFrom` reads a bean property as two-state: null means absent (keep), non-null means set/parse. JSON Merge Patch and the OpenAPI-generator ecosystem (`JsonNullable<T>`) distinguish three states: property absent (keep), property explicitly `null` (clear), property present (set). A team whose wire contract includes null-clears cannot express it with `UpdateSpec` today; the hkj-book's answer (model a sentinel) changes the wire contract, which is problematic for adoption.

## What "good" looks like (user perspective)

An opt-in property shape that carries the third state, with the rest of the sparse contract (validation of present values, accumulation, laws) intact:

```java
public class UserPatchDto {
  private JsonNullable<String> nickname = JsonNullable.undefined();  // or an hkj Updatable<T>
  ...
}

// updateFrom: undefined -> keep; JsonNullable.of(null) -> clear (component must be clearable);
// JsonNullable.of(v) -> set/parse v, located error if invalid.
```

## Things to explore / constraints

- **Carrier type.** Direct `JsonNullable` support (a soft dependency or duck-typed detection) vs an hkj-owned `Updatable<T>` with a Jackson module; the OpenAPI-generator interop argues for recognising `JsonNullable` itself.
- **What "clear" writes.** Only lawful for components that can hold a cleared state: `Optional` domain components (currently rejected in the sparse tier precisely because two-state null-as-absent could not encode "set empty" — the third state resolves that original objection) or nullable-by-domain-choice fields (which HKJ records typically do not have). The tier's laws gain a clear-idempotence clause.
- **Doctrine boundary.** The two-state default stays exactly as is; tri-state is opt-in per property by using the carrier type, so existing specs are untouched.
- **Rejections that flip.** `Optional<T>` domain components become mappable in the sparse tier when (and only when) the wire side is tri-state; the diagnostic wording must explain the pairing.

## Dependencies / scope

- Depends on: demand; this is the review's "who does not care" case in reverse (only JSON-Merge-Patch/OpenAPI shops need it).
- In (if pursued): design note first (carrier decision), then sparse-tier support, laws, book contrast update (the dense/sparse table becomes dense/sparse/tri-state).
- Out: dense tier and full-tier semantics (unchanged); JSON Patch (RFC 6902) operations.

PR #693, opened 2026-08-10
@GenerateMapping: one leaf vocabulary across tiers — sparse UpdateSpec element lifting (#672)

### LINKED ISSUE #672: [FEAT] @GenerateMapping: one leaf vocabulary across tiers (element vs container prisms)
## The gap

For a `List<PhoneNumber>` component named `phones`, the full tier's leaf convention is an
*element* prism (`ValidatedPrism<String, PhoneNumber> phones()`, lifted through the list), but
the sparse tier's container rule demands a *whole-container* prism under the same name
(`ValidatedPrism<List<String>, List<PhoneNumber>> phones()`). One shared mix-in vocabulary
therefore cannot serve a full spec and an update spec for the same domain: the comparator had to
demote the element prisms to interface statics and hand-write
`ValidatedPrism.of(PHONE::parseAll, PHONE::buildAll)` in the patch spec. The diagnostic explains
the mismatch well, but the design forces a restructure exactly where mix-ins promise reuse.

The same element leaf serves both tiers:

```java
public interface PhoneVocabulary {
  default ValidatedPrism<String, PhoneNumber> phones() { return PHONE; }
}

@GenerateMapping interface CustomerMapping
    extends PhoneVocabulary, MappingSpec<Customer, CustomerDto> {}     // lifts elementwise (today)

@GenerateMapping interface CustomerPatchMapping
    extends PhoneVocabulary, UpdateSpec<Customer, CustomerPatchDto> {} // present list: parse
                                                                       // elementwise, located
                                                                       // phones.1 errors (new)
```

- **Mechanism.** The sparse tier learns element lifting for present containers: a present
  `List<S>` wire property routes through the element leaf's `parseAll` (located, accumulating,
  consistent with the dense tiers) instead of requiring a whole-container prism. This also
  retires the "container element-lifting is deferred" carve-out in the update path.
- **Wholesale semantics kept.** The replace-wholesale contract is unchanged (a present container
  still replaces the component entirely); only the element validation and error location improve.
- **Precedence.** If both an element-shaped and a container-shaped leaf match the name, the
  existing explicit-leaf-wins grammar needs a stated tie-break (container prism wins, as the more
  specific declaration; diagnostic on genuine ambiguity).
- **Identity containers in the sparse tier.** Decide whether the identity-container null scan
  (dense tiers) extends to present identity containers in `updateFrom` for uniformity, or stays
  wholesale-as-sent (the current, documented position).

## Dependencies / scope

- Depends on: nothing hard; touches `classifyUpdate` and the sparse emission only.
- In: element lifting for present `List`/`Map`-value/`Optional` properties in the sparse tier,
  tie-break rule + diagnostic, laws (sparse validation law over a container element), goldens,
  book update to the UpdateSpec rules list.
- Out: tri-state semantics (#680), deep merge.

MagnusSmith:
Closes #672.

## What

The sparse `UpdateSpec` tier learns the dense tiers' element vocabulary: a present `List`, `Optional` or `Map`-valued PATCH property now routes through the element leaf named after the component, so the exact mix-in interface the issue asked for works — one `PhoneVocabulary` serving a full spec and its PATCH sibling:

```java
public interface PhoneVocabulary {
  default ValidatedPrism<String, PhoneNumber> phones() { return PHONE; }
}

@GenerateMapping interface RosterMapping
    extends PhoneVocabulary, MappingSpec<Roster, RosterDto> {}        // lifts elementwise (as before)

@GenerateMapping interface RosterPatchMapping
    extends PhoneVocabulary, UpdateSpec<Roster, RosterPatchBean> {}   // now lifts too: phones.1 errors
```

`classifyUpdate` resolves through the same `containerLeafCorrespondence` the dense tiers use, and the emission selects the parser by kind:

```mermaid
flowchart LR
    P([present wire property]) --> W{whole-component leaf?}
    W -->|yes: more specific, wins| L1["parseIfPresent(…, leaf::parse)"]
    W -->|no| E{element leaf over List / Optional / Map?}
    E -->|yes| L2["…leaf::parseAll / leaf::parseValues / element-of-Optional lambda"]
    E -->|no| I{identity match?}
    I -->|parameterised List/Map| L3["…Impl::hkj$allPresent / hkj$valuesPresent (null scan)"]
    I -->|other same type| L4["setIfPresent (wholesale, as before)"]
    I -->|no| N[nested spec wholesale, or located diagnostic]
```

Replacement stays wholesale — only element validation and error location improve (`phones.1: not a phone number`, accumulating). The `UpdateEdit` "element-lifting is deferred" carve-out retires, and `resolveNestedSpec`'s docs now name both call sites. Per the issue's outs: no tri-state (#680), no deep merge, no nested-spec-through-container lifting (the book prescribes an element leaf delegating to the nested Impl's `asValidatedPrism()` for that).

## The issue's explore points, resolved

- **Tie-break**: a whole-container leaf wins over the element interpretation as the more specific declaration — the dense tiers' existing order, now stated in the javadoc, book, glossary and skill, and pinned by a test. Genuine ambiguity is structurally impossible (one method name, one return type), so no ambiguity diagnostic exists; a Kind-canary test instead forces any future correspondence kind to choose its sparse emission before landing.
- **Identity containers**: extended to the dense tiers' null scan, deliberately — a present same-typed container with a null element/value is now a located rejection (`tags.1: must not be null`) instead of writing the null into the domain record. This reverses a documented 0.4.9 behaviour ("written as sent, unscanned") and the release-history entry names the reversal explicitly. The scan applies only where the emitted generic helper can type: raw and wildcard-argument containers stay plain identity writes (see review pass).
- **Diagnostics**: the no-update-source fix line offers both leaf forms for container pairs, element first — gated by the same `liftableMapPair` key check the resolver uses and a wildcard check, so it never suggests a leaf the resolver would refuse to consult. The Optional bridge rejection is scoped to non-Optional wires (an Optional-typed property *can* express "set to empty"), and its fix line names that alternative.

## Review pass

The pre-PR five-lens panel (correctness, type safety, semantics, ergonomics, architecture) verified the classification order, `Edit.parseIfPresent` contracts, law preservation (identity/idempotence/validation hold structurally — every edit is a pure set-shaped function of the wire), and reference semantics (identical aliasing posture to the dense tiers). Findings folded in:

- **Type safety (probe-confirmed regressions, fixed)**: routing wildcard-argument or raw same-typed containers through the generic scan helper broke compilation of the generated Impl (wildcard captures unify inconsistently; raw types cascade five unchecked warnings into `-Werror` builds). `sparseIdentityKind` now demotes those shapes to plain `setIfPresent` — the 0.4.9 behaviour — with a regression test compiling all four shapes.
- **Correctness (probe-confirmed, fixed)**: the suggestion helper offered dead-end element leaves for mismatched-key Map pairs and wildcard element types — exactly the leaves the resolver never consults. Both gated, with covering diagnostic tests.
- **Semantics**: the 0.4.9 reversal disclosure, plus Jackson caveats on the Optional bullet (an explicit JSON `null` binds to `Optional.empty()` and clears; the bean field must default to `null`, not the idiomatic `Optional.empty()`, or omitted fields clear).
- **Architecture/ergonomics**: the element-of-Optional lambda is now one shared fragment across tiers; the Map-pair gate is one shared helper; javadoc contradictions and roadmap narration cleaned; dash styles, glossary wording, and the release-history bullet order aligned.

A post-open review round tightened the proof obligations and the docs:

- **Accumulation proven, not assumed**: the list and map element tests now supply two invalid entries each and assert both located errors, and the identity-scan test applies a null `Map` value at runtime, asserting `tags.1` and `labels.tier` accumulate across both container edits.
- **Bean-Optional rule scoped**: the full tier's "optionality bridges through `null`" rule is scoped to full mappings and cross-references the sparse tier's deliberate exception (an Optional-typed PATCH property encodes "set to empty"), so the two book sections no longer read as contradicting.
- **Jackson caveat names the behaviour**: Jackson 3 binds an explicit JSON `null` on an Optional-typed property to `Optional.empty()` regardless of JDK version — the caveat no longer attributes this to the retired jdk8 datatype module.

## Coverage and docs

- Fifteen new processor tests (emission, runtime for all three container kinds, tie-break, identity scans, raw/wildcard regression, Kind canary, five diagnostic shapes) plus a new golden locking every emission shape; `MappingProcessor` stays at 100% line and branch coverage.
- The book's Roster fixture law-checks the sparse laws over container elements through `MappingLaws` (compile-verified anchors); the UpdateSpec rules list is rewritten; glossary, hkj-mapping skill, and release-history aligned.

## Summary by CodeRabbit

* **New Features**
  * PATCH mappings support element-level validation for lists, maps, and optional values.
  * Optional wire properties distinguish absent fields from explicitly empty or null values.
  * Container updates support whole-container replacement or element-level parsing with indexed and keyed error paths.
  * Nested mappings and inherited rules are handled consistently.
  * Identity containers scan for null values, while unsupported or incompatible shapes remain rejected.
  * Added examples for validated list replacement and located element errors.

* **Documentation**
  * Expanded guidance on sparse updates, containers, optional values, and validation semantics.


MagnusSmith:
Review comments addressed in bbc7ca8d5:

- **Accumulation over container elements**: the list test now feeds two invalid elements and asserts `phones.0` **and** `phones.2`; the map test feeds two invalid values (LinkedHashMap, so entry order is deterministic) and asserts both key-located errors — proving the lifted parsers accumulate rather than short-circuiting.
- **Map identity scan at runtime**: the identity-scan test now applies a null map value and asserts `tags.1` and `labels.tier` accumulate across both container edits, not just the emitted source.
- **Bean-Optional rule scope**: the full tier's "optionality bridges through `null`" bullet is now scoped to full mappings and cross-references the sparse tier's deliberate exception, so the two sections no longer read as contradicting.

On the JDK8 wording: the *behaviour* is still real on Java 25 — Jackson 3 folded the old `jackson-datatype-jdk8` module's `Optional` handling into core databind, and an explicit JSON `null` still binds an `Optional`-typed property to `Optional.empty()`. What was wrong was naming the retired module; the caveat now names the behaviour instead.

PR #716, opened 2026-08-23
hkj-book: apply the exemplar treatment to the optics Integration and Recipes sub-chapter

MagnusSmith:
Applies the exemplar treatment (#704, #707, #708, #709, #710, #713) to the optics **Integration and Recipes** sub-chapter: nine pages treated. Stacked on #713.

## Why this chapter needed it most

It had roughly 100 java fences and **one** of them compile-verified. Nothing here had ever been through a compiler. Adding a single gated payoff to the chapter intro found a phantom in the first minute, and the accuracy lens then compile-proved seven more.

## The sub-chapter now opens like the exemplar

- **A verified payoff before any theory**: `ch5_intro` shows the capstone in miniature, a composed `Lens >>> Prism >>> Traversal >>> Lens` path with an accumulating validation, under the gate via a new `optics_ch5_intro` fixture (ratchet 306 → 307).
- **The "Which Optic Do I Need?" flowchart is now mermaid, and correct.** The ASCII version had no `Affine` outcome at all and routed "Optional" to `Prism`, conflating *a field that may be absent* with *a value that may be another variant*.
- **Five dark-baked diagrams replaced.** The batching pages used PlantUML SVGs with a Catppuccin Macchiato ground compiled in, on a book whose `default-theme` is `latte` — they rendered as dark slabs for anyone who had not switched themes. Now four mermaid sequence diagrams (theme-managed by `mermaid-init.js`) and one flowchart on the house palette; the ten orphaned assets are deleted.
- **Key Takeaways on all eight content pages**, converting the prose closers; See Also webs; "Why this matters" beside three real differentiators; the batching-era pages moved to the two-line nav; `auditing_complex_data_example` gained the `Next` link it never had.

## What was actually broken

```mermaid
flowchart TD
    Q{"Chain two optics<br/>of different kinds"}
    Q -->|"Lens >>> Prism"| A(["Affine"])
    Q -->|"Prism >>> Lens"| A
    Q -->|"Iso >>> Traversal"| X(["no such overload:<br/>asTraversal() first"])
    Q -->|"the page said"| T(["Traversal"])

    class Q decision
    class A tier
    class X,T error
```

**Composition results were wrong in four places, always the same mistake**: a mixed chain settles at the *weakest* optic, not at `Traversal`. In the cookbook this inverted the advice outright — the block labelled "Preferred: type-safe, clearer intent" did not compile, while the "Fallback" beneath it did. Two prism-narrowed chains also kept using lenses on the *wide* type after the prism had narrowed to the subtype.

**Phantoms**: `Traversal.empty()` and `Traversal.getAll` (it does not extend `Fold`; `asFold()` is a conversion), `Validated.match` (it has `fold`), `IntValueLenses` (that record carries no `@GenerateLenses`), `settingWithKey`, and — found by the very first gated snippet — `PrincipalPrisms.userLogin()`, used **ten times**. That last one traced back to a global "user" → "userLogin" replacement that had also corrupted the prose and four variable names.

**`UserPrisms` on a record.** The advanced-composition section narrowed a `record User` with a prism; `PrismProcessor:64` accepts only interfaces and enums.

**A chained `.each()` cannot infer its element type** — `FocusPath.each()` takes `E` from the target, so only the *last* one in a chain has it. Two recipes needed witnesses, and both also composed lenses they never declared.

**`OpticOps` is source-first; `TraversalExtensions` is optic-first.** Seven capstone calls had them the wrong way round, including the one under its own "~60% less code" headline. Both conventions are now stated where the two families meet.

**Five code blocks on the audit page were physically corrupted** — collapsed onto single lines with literal `\_` escapes from a markdown round-trip.

**The same false performance claim, for the fourth chapter running**: "lazy evaluation" and "structural sharing" where `ListTraverse.traverse` loops every element with no early exit and rebuilds the container, plus `modifyAllEither` sold as a work-saving short-circuit on five more sites. It keeps the first error; it does not skip elements.

## Review

The standing three-lens panel ran. Every finding was verified against the implementation before any wording changed, and seven were compile-proved by a temporary fixture through `bookVerify`. The panel also caught **four errors in this pass's own new material**: two cookbook Key Takeaways contradicted by recipes on their own page, and two mistakes in the new mermaid diagrams (`Fetch.runCached` takes a `Function`, not a `BatchLoader`; `Guard.check` has three parameters, and the third is what makes `maxBackendCalls` work).

Deliberately left: `optics_extensions` still opens as API reference rather than with a payoff, and the audit page still puts its alternatives/pitfalls/benchmark sections ahead of the solution. Both are restructures rather than corrections, and both are noted for a follow-up.

## Verification

- `bookVerify` green with the ratchet at 307.
- 28 mermaid diagrams pass the parse gate; clean `mdbook build` under the pinned toolchain; no em dashes; every link and anchor in the nine pages resolves against the rendered HTML, inbound links included.

## Summary by CodeRabbit

- **Documentation**
  - Updated optics guides, cookbook recipes, and examples to match the current API and usage patterns.
  - Added a compilable Java capstone example covering validation, optics composition, and accumulated failures.
  - Replaced static batching and guardrail diagrams with clearer Mermaid diagrams.
  - Clarified batching, validation, performance, fail-fast behavior, and optic composition semantics.
  - Added “Key Takeaways,” “See Also,” and improved navigation throughout the documentation.
- **Tests**
  - Expanded documentation snippet verification to cover one additional example.


MagnusSmith:
@coderabbit full review

PR #717, opened 2026-08-23
hkj-book: apply the exemplar treatment to the optics Advanced Optics sub-chapter

MagnusSmith:
Applies the exemplar treatment to the optics **Advanced Optics** sub-chapter: three pages. Stacked on #716.

## The finding that matters

**`validating()` executes your program. It is not a dry run.** The chapter was built on the opposite belief, in thirteen places across all three pages.

The library says so itself. `ValidationOpticInterpreter`'s class javadoc: *"this interpreter **executes** optic operations while collecting validation warnings and errors"*. Inside `validate()`: *"We must execute operations to get proper results for flatMap chaining"*. And a `modify` modifier runs **twice** — once in `validateModify` to check it, once in `safeExecuteModify` to perform it.

```mermaid
flowchart TD
    Q{"What do you want<br/>from the program?"}
    Q -->|"the result"| D(["direct(): runs it"])
    Q -->|"the result and a trail"| L(["logging(): runs it"])
    Q -->|"a report instead<br/>of the result"| V(["validating(): also runs it"])
    Q -->|"nothing executed"| P(["ProgramAnalyser.analyse<br/>structural walk"])

    class Q decision
    class D,L,V tier
    class P wire
```

Five of those thirteen sites were written by **this pass**, in the commit immediately before the fix: the payoff comment "do not run it at all", the mermaid node "inspect, never run", the note "choosing it is choosing not to execute", a Why-this-matters claim about dry runs, and a Key Takeaway stating flatly that `validating()` does not run your program. The API name is a very convincing trap. The pages now describe a *checked run* — safe for pure modifiers over immutable data, unsafe for anything with a side effect — and point at `ProgramAnalyser.analyse`, whose traversal is documented as "structural and does not execute the program".

## The rest of what was broken

- **An interface that does not exist.** The page printed `public interface OpticInterpreter { <A> A run(...); }` and said all interpreters implement it. There is no such type: the three built-ins are independent `final` classes. That also made "custom constraints via custom interpreter subclass" impossible twice over.
- **Both custom-interpreter examples were broken.** They returned `Id.of(result)` where the fold requires `Id.of(Free.pure(result))` — a runtime `ClassCastException`; and the profiling interpreter's switch could not compile at all, because each mention of a wildcard pattern variable is a fresh capture, so `optic()` and `source()` never agreed.
- **Every printed audit-trail line was fabricated.** `LoggingOpticInterpreter` names optics by `getClass().getSimpleName()`, so generated and composed optics appear as `Lens$N`, and a `get` shows as an `OpticPrograms` lambda. The arithmetic was right; only the names were invented. Root cause is upstream — the library's own javadoc claims `GET: PersonLenses.age() -> 25`, which is worth its own issue.
- **`getLog()` is a live view, not a snapshot**, so the "Managing Logs" example did not hold what it claimed after `clearLog()`.
- A `flatMap` continuation that throws is not caught, so the migration loop died rather than collecting errors; `PROBATION` was used but absent from the enum the page defines; rollback and operation fusion were sold as capabilities nothing provides; and `direct()` was recommended for performance-critical paths against the sibling page's correct advice.

## The treatment

- **A verified payoff on the intro**, which had no code at all: one withdrawal described once, then run directly, run again through the logging interpreter, and finally run through the validating interpreter for a report. New `optics_ch6_intro` fixture; ratchet 307 → 308.
- **A "Which interpreter do you need?" mermaid flow**, with the fourth branch the chapter needed once the dry-run claim was corrected.
- Key Takeaways and See Also on both content pages, converting the "Next Steps" prose lists; subtitles; untyped admonitions typed; a duplicated Hands-On block removed; and `interpreters` gained the `Next` link into Reference it never had.
- `Free<OpticOpKind.Witness, A>`, `OpticPrograms` and `OpticOps` glossed once — the chapter stopped to explain `Either` and `Tuple`, which are incidental, and never explained its own vocabulary.

## Review

The standing panel ran. Every finding was verified against the implementation before any wording changed. Deliberately left for a follow-up, both being restructures rather than corrections: roughly 290 lines are duplicated between the two pages (the same three scenarios told twice with no cross-link), and `free_monad_dsl` buries its own "do you need this?" section at 79% of the page.

## Verification

`bookVerify` green at 308, 29 mermaid diagrams parse, clean `mdbook build`, no em dashes, every link and anchor resolves — and the panel confirmed no inbound link into these three pages uses an anchor, so the heading churn broke nothing.

## Summary by CodeRabbit

* **Documentation**
  * Added a worked example showing one optics program executed directly, with logging, and with validation.
  * Clarified that validation executes operations and reports problems; structural inspection is available separately.
  * Added guidance for choosing interpreters, updated comparisons, and expanded learning resources.
  * Improved examples and explanations for error handling, audit logs, profiling, mock interpreters, and validation rules.

* **Tests**
  * Added verification coverage for the new optics documentation example.
  * Increased the minimum required verified snippet count.

### DIFF
--- a/hkj-book/src/optics/ch6_intro.md
+++ b/hkj-book/src/optics/ch6_intro.md
@@ -10,6 +10,33 @@ Most optic work involves the everyday tools: lenses, prisms, traversals, and the
 
 This chapter is for those occasions. The Free Monad DSL turns optic operations into a value you can pass around, inspect, and execute under different strategies (production, audit, dry-run, mock). Interpreters are the strategies that turn descriptions into results.
 
+Here is the whole idea before any of the theory. One program, described once, run three different ways. Every line compiles against the real library on every build:
+
+<!-- verify -->
+```java
+// A description, not an action: nothing has touched the account yet
+Free<OpticOpKind.Witness, Account> withdrawal = Fixture.withdraw(Fixture.account, 30);
+
+// Run it for real
+DirectOpticInterpreter direct = OpticInterpreters.direct();
+Account settled = direct.run(withdrawal);
+// Account[id=ACC-1, balance=70]
+
+// Run the same value again, recording every optic operation on the way
+LoggingOpticInterpreter logging = OpticInterpreters.logging();
+Account audited = logging.run(withdrawal);
+List<String> trail = logging.getLog();
+
+// Or do not run it at all: inspect what it would do
+ValidationOpticInterpreter validator = OpticInterpreters.validating();
+ValidationOpticInterpreter.ValidationResult check = validator.validate(withdrawal);
+boolean safeToRun = check.isValid();
+```
+
+~~~admonish tip title="Why this matters"
+The three blocks differ by one line. `withdrawal` is an ordinary value: it can be stored in a field, passed to a method, returned from one, and run later or never. That is the property the rest of this chapter trades on. An audit trail stops being logging statements scattered through the code and becomes a second interpreter over the same description; a dry run stops being a boolean flag threaded through every method and becomes a decision not to call `run`.
+~~~
+
 If you have not yet hit a problem that needs this, you do not need this chapter. Come back when an audit requirement, a testability concern, or a multi-mode execution scenario forces the issue.
 
 ~~~admonish info title="In This Chapter"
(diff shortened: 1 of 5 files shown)


MagnusSmith:
@coderabbit full review

PR #720, opened 2026-08-24
hkj-book: apply the exemplar treatment to the optics Reference sub-chapter

MagnusSmith:
Applies the exemplar treatment to the optics **Reference** sub-chapter: six pages. Stacked on #717. **This completes the optics rollout** (#708, #709, #710, #713, #716, #717, and this).

## Why a reference chapter is the worst place for a wrong fact

A tutorial page with a bad claim costs you a confused reader. A lookup page with a bad claim costs you a reader who *checked*, and was told the wrong thing by the page whose entire job is to be checkable. Two findings here are of that kind.

**Six of the seven quoted processor diagnostics did not exist.** A repo-wide grep for `not allowed on this type`, `must return Iso`, `no copy strategy` and `not assignable to source` returns nothing in any `.java` or `.properties`. The reader's workflow on that page is pasting an error string into a search box; none of these would ever have matched. Every heading now carries what the processor actually emits, and several *causes* were wrong too: `@GenerateIsos` only counts type arguments and never checks the raw type is `Iso`, and `@MatchWhen` is not validated at all, so a typo surfaces as javac `cannot find symbol` inside generated code.

**The capability table was wrong in about fifteen cells.** I rebuilt it from `javap` against the built interfaces rather than from reading source, which is what exposed the scale:

| Claim | Reality |
|---|---|
| `Lens`/`Iso` have `getAll` and `getOptional` | Neither has either |
| `Iso` has `set` and `modify` | Only `get`, `reverseGet`, `modifyF`; `reverseGet` rebuilds |
| `Prism` has `set` | It does not |
| `foldMap` on six optics | Only `Fold` and `Getter`; it is not on `Optic` |
| `Setter` focuses "exactly one" | Zero-or-more, per its own javadoc |
| `Getter` blank for the query family | It extends `Fold` and inherits all of it |

The root cause was the preamble: a ✓ conflated "declares the method" with "can do it eventually". It now says which, and points at Conversions for the rest.

## Two broken snippets

The caching example composed `OrderFocus.items().each().price()` — a generated collection accessor is already element-level, and `TraversalPath` has no field accessors. Worse, the compiler-errors page's **own fix** for an inference failure added a surplus `.each()`, which *compiles* (it infers its element type from the assignment target through an unchecked cast) and then fails at runtime. A silently-compiling runtime bug, presented as the fix, on the page about diagnosing compile failures.

## The page called Decision Trees had three ASCII decision trees

Exactly the case the style guide assigns to mermaid. All three converted. Tree 1 was the same defective tree #716 fixed in `ch5_intro` — no Affine, "Optional" routed to Prism — and here it contradicted its own table directly below, which lists Affine correctly.

## Two reversals on production readiness

"Additional caching gains nothing" for a single generated lens is backwards: the accessor is a factory calling `Lens.of` on every invocation. And incremental compilation does not regenerate "just that record's companion classes" — every processor is registered *aggregating*, which the repo's own `incremental.annotation.processors` states outright.

## Review

The standing panel ran, and **caught five errors in this pass's own new material** — four cells in the table I had just rebuilt, and three Key Takeaways contradicting their own pages. That is the recurring lesson of this rollout: the closers written during a treatment pass are the most defect-prone thing in it, and every chapter has needed the panel pointed at its own additions.

One result worth recording as a positive: `production_readiness`'s short-circuit wording was checked and found **correct** — the first chapter in six where that claim was already right.

Deferred, and worth a follow-up issue rather than this PR: `ch5_intro` still carries a byte-identical copy of Tree 1, and `focus_reference.md`'s Troubleshooting section duplicates four of this chapter's entries while contradicting one of them on `traverseOver`.

## Verification

`bookVerify` green at 309, 32 mermaid diagrams parse, clean `mdbook build`, no em dashes, every link and anchor resolves — and the panel confirmed all seven inbound links into these pages are file-level, so the heading churn broke nothing.

## Summary by CodeRabbit

* **Documentation**
  * Added an introduction to optic operation support, including a Java example for converting a `Lens` to a `Fold`.
  * Clarified optic capabilities, conversions, traversal options, runtime costs, caching, and compilation guidance.
  * Reworked compiler-error guidance with expanded diagnostics and troubleshooting advice.
  * Replaced decision-tree ASCII diagrams with clearer Mermaid diagrams and restored navigation links.
  * Added Key Takeaways and See Also sections across relevant pages.
* **Tests**
  * Added verification for a new optics documentation snippet and increased the verified snippet count.

### DIFF
--- a/hkj-book/src/optics/ch7_intro.md
+++ b/hkj-book/src/optics/ch7_intro.md
@@ -10,6 +10,25 @@ The reference cluster collects the lookup material that returning readers come b
 
 Each page in this cluster is structured as a quick-scan reference, not a tutorial. If you need the conceptual material, the earlier chapters cover it in depth.
 
+Most lookups here resolve to one distinction, so it is worth stating before the tables do. An optic either declares an operation or it does not, and when it does not, a conversion usually reaches it anyway:
+
+<!-- verify -->
+```java
+Lens<Order, String> customer = OrderLenses.customer();
+
+String name = customer.get(Fixture.order);
+// "Alice": get is declared on Lens
+
+// customer.getAll(Fixture.order);
+// will not compile: getAll is on Fold, not Lens
+
+Fold<Order, String> asFold = customer.asFold();
+List<String> all = asFold.getAll(Fixture.order);
+// ["Alice"]: the same access, one conversion later
+```
+
+[Optic Capabilities](optic_capabilities.md) is the table of what each optic declares; [Conversions](conversions.md) is the table of how to get from one to another. Between them they answer most of what brings a returning reader back.
+
 ~~~admonish info title="In This Chapter"
 - **Optic Capabilities** – A unified table showing which operations (`get`, `set`, `modify`, `modifyF`, `getAll`, `preview`, `foldMap`, `matches`, `build`) each optic type supports, including the asymmetric specialists (`Getter`, `Setter`, `Fold`).
 - **Conversions** – The methods for converting between optic types (`asTraversal`, `asFold`, `asLens`, `andThen`) and the rules governing what type results from composing two optics.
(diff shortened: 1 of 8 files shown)


MagnusSmith:
@coderabbit full review

MagnusSmith:
@CodeRabbit full review

Issue #721, opened 2026-08-24
[BUG] @GenerateTraversals: a wildcard type argument generates uncompilable source


MagnusSmith:
`@GenerateTraversals` copies a record component's element type argument straight into the `Traversal` it declares. When that argument is a wildcard, the generated source names `Traversal<Holder, ? extends Leaf>` — legal as a return type, but not as the anonymous class the method returns, and not as anything `Traversals.traverseList` can be applied to.

Unlike #718, this is not about optic-instance inference and it is not confined to SPI containers: the hardcoded `List` fails too.

## Reproduction

```java
@GenerateTraversals
public record Holder(List<? extends Leaf> items) {}
```

The generated `HolderTraversals` does not compile:

```
HolderTraversals.java:30: error: unexpected type
    return new Traversal<Holder, ? extends Leaf>() {
HolderTraversals.java:30: error: no interface expected here
HolderTraversals.java:31: error: method does not override or implement a method from a supertype
HolderTraversals.java:35: error: method traverseList in class Traversals cannot be applied to given types
HolderTraversals.java:36: error: incompatible types: Object cannot be converted to List<? extends Leaf>
```

The emitted shape:

```java
public static Traversal<Holder, ? extends Leaf> items() {
  return new Traversal<Holder, ? extends Leaf>() {   // <- illegal
```

## The mechanism

```mermaid
flowchart TD
    F["a container-typed record component"] --> X["read the type argument<br/>at the generator's focus index"]
    X --> W{"is it a wildcard?"}
    W -->|no| OK["Traversal&lt;S, Leaf&gt;<br/>compiles"]
    W -->|yes| E["Traversal&lt;S, ? extends Leaf&gt;<br/>illegal as an anonymous class"]

    classDef wire fill:#8caaee,stroke:#1e66f5,color:#232634
    classDef decision fill:#e5c890,stroke:#df8e1d,color:#232634
    classDef error fill:#e78284,stroke:#d20f39,color:#232634
    classDef domain fill:#a6d189,stroke:#40a02b,color:#232634
    class F wire
    class X wire
    class W decision
    class E error
    class OK domain
```

`TraversalProcessor` reads the focus type at `TraversalProcessor.java:148`:

```java
focusType = TypeName.get(declaredType.getTypeArguments().get(typeArgumentIndex)).box();
```

[...]

PR #722, opened 2026-08-24
@GenerateFocus: reject a container the widening cannot be written for (#718)

### LINKED ISSUE #718: [BUG] @GenerateFocus: a wildcard type argument in an SPI container generates uncompilable widening
`@GenerateFocus` widens an SPI container by handing the path an optic **instance**: `.some(Affines.eitherRight())` for a `ZERO_OR_ONE` generator, `.each(EachInstances.mapValuesEach())` for `ZERO_OR_MORE`. That instance's type arguments are inferred from the field type, and a field parameterised with a wildcard has no ground instantiation to infer them from. javac settles on `Object` and the generated call does not apply to the path.

## Reproduction

```java
@GenerateFocus
public record Holder(Either<String, ? extends Leaf> boundedEither) {}
```

The generated `HolderFocus` does not compile:

```
error: no suitable method found for some(
    org.higherkindedj.optics.Affine<org.higherkindedj.hkt.either.Either<java.lang.Object,java.lang.Object>,java.lang.Object>)
```

## The mechanism

```mermaid
flowchart TD
    F["a container-typed record component"] --> H{"hardcoded container?<br/>Optional, List, Set, Collection"}
    H -->|yes| B[".some() / .each()<br/>built-in, free type variable"]
    H -->|no: SPI generator| O[".some(optic) / .each(optic)<br/>optic instance, inferred from the field type"]
    B --> OK["compiles"]
    O --> W{"is a type argument<br/>a wildcard?"}
    W -->|no| OK
    W -->|yes| E["no ground instantiation<br/>javac infers Object<br/>method does not apply"]

    class F wire
    class H,W decision
    class E error
    class B,O,OK domain
```

The hardcoded containers escape because `.each()` and `.some()` are built-in methods with a free type variable and no optic argument to unify.

## What is affected

Measured with the real processor, one field per compilation:

| field | widening emitted | generated source compiles |
|---|---|---|
| `Either<String, ? extends Leaf>` | `.some(Affines.eitherRight())` | no |
| `Either<String, ?>` | `.some(Affines.eitherRight())` | no |
| `Map<String, ? extends Leaf>`, `generateNavigators = true` | `.each(EachInstances.mapValuesEach())` | no |
| `Map<String, ? extends Leaf>`, default | none | yes |
| `List<? extends Leaf>` | `.each()` | yes |
| `Optional<? extends Leaf>` | `.some()` | yes |
| `Either<String, Leaf>`, `Map<String, Leaf>` | as above | yes |

A wildcard-valued `Map` compiles by default only because the static Focus method leaves every `ZERO_OR_MORE` SPI field a plain `FocusPath` (the `^1` footnote in the container-types reference). Turning navigators on emits the widening, and the failure appears.

## Why the tests do not catch it

Every SPI container in the processor tests is parameterised with a concrete type, so no test ever asks the generator to widen a wildcard.

## Options

1. **Diagnose at the declaration.** Reject a wildcard type argument in an SPI container with a message naming the component and the concrete alternative (`Either<String, Leaf>`). This matches the generator's existing habit of failing where the mistake is written rather than emitting source that cannot compile.
2. **Leave the field un-widened**, as the static method already does for every `ZERO_OR_MORE` SPI field, and let the reader widen it explicitly.
3. **Emit through a capturing helper.** `Each<Map<K, V>, V>` cannot be denoted for `Map<String, ? extends Leaf>` at all, so this needs a raw bridge and a justified `@SuppressWarnings`; it buys generality at a cost the first two options avoid.

Option 1 is the cheapest honest answer and option 2 the most permissive; both are preferable to the present behaviour, which is a compile error inside generated code.

---

Found while reviewing #714. It predates that change and is independent of it: the same reproduction fails on `main`.

MagnusSmith:
Fixes #718.

## What was broken

An SPI container widens by receiving an optic **instance** — `.some(Affines.eitherRight())`, `.each(EachInstances.mapValuesEach())` — whose own type arguments javac infers from the field type. A wildcard has no ground instantiation to infer them from, so the call did not apply to the path and the failure surfaced *inside generated source* rather than at the declaration that caused it:

```
error: no suitable method found for some(Affine<Either<Object,Object>,Object>)
```

Measured with the real processor, one field per compilation. Two rows the issue did not record:

| field | before | note |
|---|---|---|
| `Either<String, ? extends Leaf>` | uncompilable | the issue's reproduction |
| `Either<?, Leaf>` | uncompilable | **wildcard in the type argument the generator does not focus on** — `Affines.eitherRight()` must infer `L` as well as `R` |
| `Either raw` | uncompilable | **raw containers fail identically**; pre-existing, same root cause |
| `Map<String, ? extends Leaf>`, `generateNavigators = true` | uncompilable | |
| `Either<String, List<? extends Leaf>>` | compiles | the wildcard belongs to the `List`; `Either` is still ground |
| `List<? extends Leaf>`, `Optional<? extends Leaf>` | compiles | `.some()`/`.each()` are free type variables with no optic to unify |

## The mechanism

```mermaid
flowchart TD
    F["a container-typed record component"] --> H{"hardcoded container?<br/>Optional, List, Set, Collection"}
    H -->|yes| B[".some() / .each()<br/>free type variable, nothing to unify"]
    H -->|"no: SPI generator"| N{"does the generator<br/>name an optic expression?"}
    N -->|no| B
    N -->|yes| W{"are the container's own<br/>type arguments denotable?"}
    W -->|yes| O[".some(optic) / .each(optic)"]
    W -->|"no: raw or wildcard"| G{"would this layer<br/>widen at all?"}
    G -->|no| B
    G -->|yes| E["reject the component<br/>at the declaration"]
    B --> OK["compiles"]
    O --> OK

    class F wire
    class H,N,W,G decision
    class E error
    class B,O,OK domain
```

## The fix

Option 1 of the issue: reject at the declaration, in the shared what/why/fix shape, naming the concrete alternative.

```
error: @GenerateFocus: record component 'Holder.boundedEither' has a wildcard type argument in
Either<String, ? extends Leaf>. The optic instance that widens that container is inferred from the
field type, and a wildcard has no ground instantiation to infer it from. Declare the component with
concrete type arguments, such as Either<String, Leaf>.
```

The container is also left **un-widened**, so the generated file still compiles and the user sees exactly one diagnostic rather than ours plus the cryptic one.

Three conditions keep the rule off code that works today:

- **The generator must name an optic expression.** One with none widens through `.nullable()`/`.each()`, whose free type variable takes a raw or wildcard argument without complaint. Every shipped generator names one; `SpiGeneratorConflictTest`'s `Solo`/`Dup` do not, and regressed until this condition existed.
- **The layer must actually widen.** A `ZERO_OR_MORE` SPI container stays a plain `FocusPath` unless collections are widened or a navigator takes it, so `Map<String, ? extends Leaf>` at the default settings is untouched.
- **…and so must everything beneath it.** The analysis neither widens an un-widened container nor looks inside it, so `Map<String, Either<String, ? extends Leaf>>` compiles by default.

Every widening site reads its generator through one guard (`wideningGenerator`), so the path kind, the navigator class and the composition call cannot disagree about which containers widen.

## How this has been tested

`gradle :hkj-processor:clean build` green, `spotlessCheck` clean, and a forced `--rerun-tasks` recompile of all five modules that run the processor — nothing in the repo's own sources trips the new rejection.

The issue's post-mortem was *"every SPI container in the processor tests is parameterised with a concrete type"*, so the new guards are built on the shape of the failure rather than on its mechanism:

**`GeneratedFocusSourceCompilesTest`** — seven container shapes (concrete, wildcard-focused, wildcard-unfocused, unbounded, `? super`, raw, wildcard-nested) × four settings. Two assertions: no error is ever located in generated output, and every error carries the `@GenerateFocus` tag, which only a diagnostic reported against a declaration does. Because a guard that says only *where* errors land is satisfied by rejecting everything, a second half pins a corpus of records that must keep compiling.

Both halves are non-vacuous, checked against `main`'s sources: 22 of the 28 shape cases fail there, and **zero** corpus cases do.

**An architecture rule** holds the chokepoint: only `wideningGenerator` and the diagnostic walk may call `findSpiGenerator`, so a new widening site cannot silently skip the guard. Verified by pointing one site back at the raw lookup and watching it fail.

## Review

A pre-merge review caught a regression this PR had introduced: the diagnostic walk descended unconditionally, so `Map<String, Either<String, ? extends Leaf>>` — which compiles on `main` — was rejected. Reproduced, fixed (the walk now stops at any layer the analysis does not widen), and the "still compiles" corpus above exists because the shape sweep alone did not catch it.

## Related

#721 — `@GenerateTraversals` is broken by wildcards far more broadly, including `List<? extends Leaf>`. Different root cause (`TraversalProcessor` copies the type argument verbatim into `new Traversal<Holder, ? extends Leaf>() {`, which is illegal as an anonymous class), so it is filed separately rather than folded in.

---

Type of change: bug fix (non-breaking), plus documentation and tests. The container-types reference in `hkj-book` and in the `hkj-optics` skill both gained the rule.

https://claude.ai/code/session_0149w8faEHS2icbw48epbUJA

## Summary by CodeRabbit

- **Documentation**
  - Added guidance on raw and wildcard container type arguments in Focus DSL usage.
  - Clarified which nested wildcard patterns are supported and when concrete type arguments are required.

- **Bug Fixes**
  - Improved handling of SPI containers with raw or wildcard types during focus generation.
  - Added diagnostics and actionable suggestions for unsupported widening scenarios.
  - Preserved valid non-widened and built-in container cases.

- **Tests**
  - Added comprehensive coverage for wildcard, raw, nested, navigated, and widened containers.
  - Verified generated focus sources compile successfully across supported configurations.

### DIFF
--- a/.claude/skills/hkj-optics/reference/container-types.md
+++ b/.claude/skills/hkj-optics/reference/container-types.md
@@ -117,6 +117,26 @@ public static TraversalPath<Employee, Integer> scores() {
 
 Patterns like `Optional<List<String>>` or `Either<E, Map<K, V>>` are detected automatically. The processor generates composed widening chains (e.g., `.some().each()`).
 
+## Raw and Wildcard Container Type Arguments
+
+An SPI container widens by receiving an optic instance — `.some(Affines.eitherRight())`, `.each(EachInstances.mapValuesEach())` — whose own type arguments javac infers from the field type. A raw container offers none to infer from, and a wildcard has no ground instantiation, so `@GenerateFocus` rejects the component rather than emitting a call that cannot compile.
+
+```java
+// Rejected: no Affine can be denoted for a wildcard type argument
+@GenerateFocus
+record Holder(Either<String, ? extends Leaf> boundedEither) {}
+
+// Accepted
+@GenerateFocus
+record Holder(Either<String, Leaf> either) {}
+```
+
+- Both of the container's own type arguments count, focused or not: `Either<?, Leaf>` is rejected too.
+- A wildcard nested *inside* an argument is fine: `Either<String, List<? extends Leaf>>` has a ground instantiation and widens to `.some(Affines.eitherRight()).each()`.
+- The built-in `Optional`, `List` and `Set` widenings take a wildcard without complaint, because `.some()` and `.each()` are methods with a free type variable and no optic argument to unify.
+- A `ZERO_OR_MORE` SPI container is rejected only when something actually widens it: `widenCollections = true`, or a navigator taking it. At the default settings it stays a `FocusPath`, and so does everything beneath it — `Map<String, Either<String, ? extends Leaf>>` compiles by default, because the un-widened `Map` means the `Either` is never asked for an optic.
+- A custom generator that names no optic expression is exempt: it widens through `.nullable()` or `.each()`, whose free type variable takes a raw or wildcard argument without complaint. Every generator shipped with HKJ names one.
+
 ## ZERO_OR_MORE Manual Widening
 
 For SPI `ZERO_OR_MORE` types, static Focus methods return `FocusPath`. Widen manually:
--- a/hkj-book/src/optics/focus_containers.md
+++ b/hkj-book/src/optics/focus_containers.md
@@ -134,6 +134,28 @@ AssetClass rebalanced =
         .modifyAll(w -> w * 1.05, assetClass);
 ```
 
+~~~admonish warning title="Raw and wildcard container type arguments"
+An SPI container widens by receiving an optic **instance** — `.some(Affines.eitherRight())`, `.each(EachInstances.mapValuesEach())` — whose own type arguments javac infers from the field type. A raw container offers none to infer from, and a wildcard has no ground instantiation, so `@GenerateFocus` rejects the component rather than emitting a call that cannot compile:
+
+```java
+// Rejected: no Affine can be denoted for a wildcard type argument
+@GenerateFocus
+public record Holder(Either<String, ? extends Leaf> boundedEither) {}
+
+// Accepted
+@GenerateFocus
+public record Holder(Either<String, Leaf> either) {}
+```
+
+Both of the container's own type arguments count, focused or not, so `Either<?, Leaf>` is rejected too. A wildcard nested *inside* an argument is fine: `Either<String, List<? extends Leaf>>` still has a ground instantiation and widens to `.some(Affines.eitherRight()).each()`.
+
+The built-in `Optional`, `List` and `Set` widenings take a wildcard without complaint, because `.some()` and `.each()` are methods with a free type variable and no optic argument to unify. `List<? extends Leaf>` widens to `TraversalPath<Holder, Leaf>` as usual.
+
+A `ZERO_OR_MORE` SPI container is rejected only when something actually widens it — `widenCollections = true`, or a navigator taking it. At the default settings it stays a `FocusPath`, and the wildcard costs it nothing. Nor does one beneath it: `Map<String, Either<String, ? extends Leaf>>` compiles at the default settings, because the `Map` is never widened and so the `Either` inside it is never asked for an optic.
+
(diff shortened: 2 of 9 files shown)


MagnusSmith:
@coderabbit review

PR #724, opened 2026-08-24
@GenerateFocus: one widening answer per component, whichever way it is reached (#719)

### LINKED ISSUE #719: [BUG] Navigator methods and static Focus methods disagree about container widening
The same record component gets a different path type depending on which way you reach it: `ShapesFocus.arrayList()` hands back a `FocusPath<Shapes, ArrayList<String>>`, while `OuterFocus.shapes().arrayList()` hands back a `TraversalPath<S, String>`. Both are generated from the same declaration by the same processor run.

## What diverges

Measured with the real processor on one record, reached both ways (after #714):

| component | static `ShapesFocus` method | navigator method | agree |
|---|---|---|---|
| `ArrayList<String>` | `FocusPath<Shapes, ArrayList<String>>` | `TraversalPath<S, String>` | no |
| `Map<String, String>` | `FocusPath<Shapes, Map<String, String>>` | `TraversalPath<S, String>` | no |
| `List<Optional<String>>` | `TraversalPath<Shapes, String>` | `TraversalPath<S, Optional<String>>` | no |
| `List<List<List<String>>>` | `TraversalPath<Shapes, String>` | `TraversalPath<S, List<List<String>>>` | no |
| `TreeSet<String>` | `FocusPath<Shapes, TreeSet<String>>` | `FocusPath<S, TreeSet<String>>` | yes |
| `Optional<List<String>>` | `TraversalPath<Shapes, String>` | `TraversalPath<S, String>` | yes (#714) |

`TreeSet` agrees only because neither side recognises it: it is a `Set` by inheritance, and no analysis follows inherited supertypes.

## Why

Two independent analyses of the same question:

```mermaid
flowchart TD
    D["a container-typed record component"] --> S["FocusProcessor.analyseFieldType"]
    D --> N["NavigatorClassGenerator.getFieldPathKindRecursive"]

    S --> S1["Optional / Collection by name"]
    S --> S2["Kind and SPI generators"]
    S --> S3["nested chain, composed to the leaf"]

    N --> N1["Optional / Collection by name"]
    N --> N2["Collection subtypes, direct interfaces only"]
    N --> N3["SPI generators"]
    N --> N4["nesting capped at 3, chained to the declared kind"]

    S3 --> R1["static method's path type"]
    N4 --> R2["navigator method's path type"]
    R1 -.->|"differ per shape"| R2

    class D wire
    class S,S1,S2,S3,N,N1,N2,N3,N4 domain
    class R1,R2 error
```

Each analysis is defensible on its own; nothing reconciles them. Three concrete differences fall out:

- the navigator walks a field type's **direct interfaces** for `Collection` subtypes, and the static analysis does not walk them at all;
- the static method leaves every SPI `ZERO_OR_MORE` field a `FocusPath` (the documented `^1` footnote), while the navigator applies the generator's optic;
- the static analysis composes a nested chain all the way to the leaf, while the navigator chains only as far as the kind it declares (which is what #714 had to reach to stop generating uncompilable source).

## Why it matters

The book and the skills teach the static shape, so a reader who navigates to the same field through a navigator gets a different type from the one the page documents, with no diagnostic. `TreeSet` is the case a reviewer raised on #714: deepening the navigator's interface walk on its own would move `TreeSet` from "consistently unsupported" into the same disagreement `ArrayList` already sits in, so it wants deciding here rather than there.

## What a fix looks like

1. **One analysis.** Extract the widening analysis (recognised containers, SPI consultation, nesting) into a single component both generators consult, so a shape cannot be classified twice.
2. **Then decide each shape once**, with the answers written down: are `Collection` subtypes recognised (and by inherited supertypes, or direct interfaces only)? Do SPI `ZERO_OR_MORE` fields widen in static methods? Does the chain always compose to the leaf?

Aligning the static methods upward changes their return types, which is source-breaking for anyone holding the current type, so it belongs at a release boundary with a release-note entry and a migration line. Aligning the navigators downward is equally breaking for navigator users, and loses the more useful shape.

---

Split out of #714, which removed this divergence for the `@Nullable` corner (the navigator applied `.nullable()` to a container while the static method widened by its contents) and for nested containers. The rest is unchanged behaviour on `main`.

MagnusSmith:
Fixes #719.

`ShapesFocus.arrayList()` and `OuterFocus.shapes().arrayList()` are generated from one declaration by one processor run, and disagreed about it. Two analyses answered the same question — `FocusProcessor.analyseFieldType` and the navigator's own `getFieldPathKind` walk — and nothing reconciled them.

## Measured, before and after

One record per shape, reached both ways, real processor:

| component | static `ShapesFocus.f()` | navigator, before | navigator, after |
|---|---|---|---|
| `ArrayList<String>` | `FocusPath<_, ArrayList<String>>` | `TraversalPath<_, String>` | `FocusPath<_, ArrayList<String>>` |
| `Map<String, String>` (default) | `FocusPath<_, Map<String, String>>` | `TraversalPath<_, String>` | `FocusPath<_, Map<String, String>>` |
| `List<Optional<String>>` | `TraversalPath<_, String>` | `TraversalPath<_, Optional<String>>` | `TraversalPath<_, String>` |
| `List<List<List<String>>>` | `TraversalPath<_, String>` | `TraversalPath<_, List<List<String>>>` | `TraversalPath<_, String>` |
| `Kind<ListKind.Witness, String>` | `TraversalPath<_, String>` | **did not compile** | `TraversalPath<_, String>` |
| `Either<String, Leaf>` | `EitherLeafNavigator<_>` | `AffinePath<_, Leaf>` | `EitherLeafNavigator<_>` |
| `TreeSet<String>` | `FocusPath<_, TreeSet<String>>` | `FocusPath<_, TreeSet<String>>` | unchanged |

Every row of the corpus now agrees, at both `widenCollections` settings, and `NavigatorAgreesWithStaticFocusTest` is the guard: it compares the two routes field by field rather than pinning any one shape, so a new container, a new setting, or a second analysis creeping back in fails without anyone having to predict which shape would diverge first.

## What changed

```mermaid
flowchart TD
    D["a container-typed record component"] --> A["WideningAnalysis"]
    A --> S["static Focus method<br/>FocusPath.of(lens, &quot;f&quot;).each()"]
    A -.->|"return type only"| N["navigation method"]
    S ==>|"composed by"| N
    N --> R["delegate.via(ShapesFocus.f())"]

    class D wire
    class A,S,N,R domain
```

The widening analysis (recognised containers, Kind fields, SPI consultation, nesting, `@Nullable` precedence) moves to one `WideningAnalysis` that both generators consult. A navigation method then stops re-deriving anything: it composes the static Focus method for the field it reaches. All nine delegate-tier × field-tier `via` overloads already return exactly the composed tier, so the two cannot drift apart, and the analysis is consulted only for the return type the method must declare.

That deletes the navigator's duplicate container tables, its Collection-subtype interface walk, its peel-to-declared-kind chain and `buildViaStatement` — a net ~330 lines off the two generators.

## Shapes that generated source javac rejected, now compiling

- a `Kind<F, A>` component on a navigator-reachable record: the navigator knew nothing of Kind fields, so it asked a widened path for its `toLens()`;
- a navigator over an SPI container of records whose target record has a plain navigable field of its own (`Map<String, Mid>` where `Mid` holds a `Leaf`): the composed path is a `TraversalPath` and `MidFocus.LeafNavigator` holds a `FocusPath`, so the path is now handed back rather than passed to it, and the chain stops one hop early instead of failing to compile;
- a record whose `targetPackage` sends its companion elsewhere: navigators looked for it where the record was declared.

A navigated path also carries its field-name segments now, so `pathString()` locates it the way a static path does (#592).

## Source-breaking for navigator callers

Three return types move to what the static method has always reported.

- **A `Collection` subtype** (`ArrayList`, `LinkedList`) is a `FocusPath` over the container. The traversal the navigator used to hand back rebuilds a plain `List`, which the component's declared type will not accept — measured `ClassCastException: ImmutableCollections$List12 cannot be cast to ArrayList` on a one-element list, working only from two elements up, by accident. Declare the component as `List<E>`, or register a `TraversableGenerator` that knows how to rebuild the concrete type.
- **An SPI `ZERO_OR_MORE` container of a non-navigable element** (`Map<String, String>`) stops at the container until the record that *declares* it sets `widenCollections = true`. Set the flag, or widen at the call site with `.each(EachInstances.mapValuesEach())`. A container whose element is itself a `@GenerateFocus` record is stepped into either way, flag or no flag, because that is how its navigator reaches the element.
- **A nested container composes to the leaf**: `List<Optional<String>>` reached through a navigator is `TraversalPath<S, String>`, so drop the `.some()` that used to follow it.

Release notes, the two book pages and the `hkj-optics` skill are updated with the rule; `NavigatorExample`'s javadoc claimed the divergent shape and now states the real one.

## Verification

- `gradle build` green across every module, from a clean `hkj-processor`.
- `WideningAnalysis` joins the CLASS-level 100% JaCoCo rule (line, branch, instruction); `NavigatorClassGenerator` holds its.
- `GeneratedFocusSourceCompilesTest` (the #718 corpus) unchanged and green, so nothing that compiled before regressed.

Found while measuring: `Set` and `Collection` components widen through a `List`-only `.each()` and throw at runtime, on both routes — not a divergence, so filed separately rather than folded in here.

https://claude.ai/code/session_01KfKediRuS6K1vWr2rVUyGm

## Summary by CodeRabbit

* **Bug Fixes**
  * Navigator-generated paths now match their corresponding static Focus paths, including collection widening and nested containers.
  * Preserved field-name path segments and improved navigation across redirected packages.
  * Fixed generated-source compilation issues involving nested container types and explicit type information.

* **Documentation**
  * Clarified when containers produce Focus, Affine, or Traversal paths.
  * Updated examples and release notes for collection-widening behavior and related source compatibility changes.

* **Tests**
  * Added comprehensive coverage confirming consistent static and navigator path types across supported containers.


MagnusSmith:
@coderabbit review

MagnusSmith:
Half of this is a real miss, and I've fixed it and swept for siblings. The other half would introduce an error, so I've left that part alone — details below.

**Taken: the stale `ZERO_OR_MORE` claim.** `NavigatorExample` still said "`ZERO_OR_MORE` (Map, arrays, third-party collections) → navigator returns `TraversalPath`" and "navigators correctly handle SPI-registered types without falling back to `FocusPath`" — three lines above the code example I *had* updated, which shows `WarehouseFocus.inventory()` returning a `FocusPath`. It contradicted itself. The claim turned out to live in six more places, so 5976f90 states it the same way in all seven:

> `ZERO_OR_ONE` → `AffinePath`, always; `ZERO_OR_MORE` → `TraversalPath` when the container is widened, which it is under `widenCollections`, or when its element is itself a navigable record and the container has to be stepped into to reach it.

Swept: `NavigatorExample`'s SPI javadoc and its printed summary, the inline tier comment on `PortfolioRiskExample.AssetClass` (which contradicted the correct one 40 lines below it), the generator-plugins takeaway and its quoted copy of the SPI javadoc, the Focus DSL journey key insight, the portfolio-risk container table, and the PCollections page. `TraversableGenerator.getCardinality()`'s own javadoc gains the condition too, since that is the contract an SPI implementor reads. The v0.4.0 release note keeps its wording — it records what shipped then.

**Not taken: "navigators apply to Collection types including List and Set".** They do not, and the SKILL.md sentence is correct as written. `NavigatorClassGenerator.navigatorTarget` reaches a navigator two ways — the field type is itself a `@GenerateFocus` record, or `spiNavigable(fieldType)` — and `spiNavigable` returns `null` for anything `WideningAnalysis.recognisedContainer` claims, which is exactly `Optional`, `Maybe`, `List`, `Set` and `Collection`. Those five widen through their own path instead and never produce a navigator class.

Measured on this branch, from the corpus in `NavigatorAgreesWithStaticFocusTest` (`Leaf` being a navigable record):

| component | generated |
|---|---|
| `List<Leaf> leaves` | `TraversalPath<_, Leaf>` — **not** a navigator |
| `Map<String, Leaf> leafMap` | `LeafMapNavigator<_>` |
| `Either<String, Leaf> eitherLeaf` | `EitherLeafNavigator<_>` |

So `List`/`Set` fields really are the case where you chain with `.via(...)`, and documenting them as navigator-producing would send readers at an API that isn't generated. Happy to reword for clarity if the current phrasing reads ambiguously, but the rule itself stands.

PR #726, opened 2026-08-25
@GenerateTraversals: read the type a wildcard stands for (#721)

### LINKED ISSUE #721: [BUG] @GenerateTraversals: a wildcard type argument generates uncompilable source

MagnusSmith:
Closes #721.

`@GenerateTraversals` copied a record component's type argument straight into the optic it declares, so a wildcard landed where javac wants a class or interface:

```java
@GenerateTraversals
public record Holder(List<? extends Leaf> items) {}
```

```
HolderTraversals.java:30: error: unexpected type
    return new Traversal<Holder, ? extends Leaf>() {
```

Legal as a return type, illegal as an anonymous class — and the error landed inside the generated file rather than at the declaration that caused it.

## The rule, and where it was not applied

```mermaid
flowchart TD
    C["a container-typed record component"] --> R["read the type argument<br/>at the generator's focus index"]
    R --> W{"is it a wildcard?"}
    W -->|no| N["name it as written"]
    W -->|"? extends T"| B["name T"]
    W -->|"? / ? super T"| O["name Object"]
    N --> D["Traversal&lt;S, A&gt;, denotable"]
    B --> D
    O --> D

    class C,R wire
    class W decision
    class N,B,O,D domain
```

`ProcessorUtils.resolveWildcard` has stated that rule since 0.4.0 and `@GenerateFocus` has always applied it; three sites in the traversal path did not. All three do now:

| site | what it names |
|---|---|
| `TraversalProcessor` | the `Traversal<S, A>` supertype and the `modifyF` signature |
| `BaseTraversableGenerator.getTypeArgumentName` | one reader, replacing the five near-identical ones on the base, Either, Validated and the two map-value generators |
| the container-typed locals | the component's **own** type, wildcards and all — that is what the accessor returns |

Rejecting the declaration, the answer #718 chose, is not the shape here: there the optic instance had no ground instantiation, while here the element type only needs a denotable name, and the bound is one.

Measured with the real processor before the fix, every wildcard form failed — `List<? extends Leaf>`, `Map<String, ? extends Leaf>`, `Either<String, ? extends Leaf>`, their unbounded and super-bounded forms, and `Map<?, Leaf>`, where the wildcard sits in an argument the generator does not even focus but the generated body still named.

## Four more shapes the sweep found

Building the sweep turned up four shapes that emitted uncompilable source and predate this fix — none of them a wildcard problem. Fixed here rather than filed:

| component | before | now |
|---|---|---|
| `List<Leaf>[]`, `Leaf[][]` | `new List<Leaf>[size]` — generic array creation; `new Leaf[][size]` — syntax error | copied to length with `Arrays.copyOf`, which needs no name and keeps the runtime component type |
| `int[]` — and all eight primitives | `Arrays.stream` has an overload for three of them, and those return a stream that cannot collect to a `List` | walked and boxed an element at a time, unboxed back into a fresh array |
| `record Holder<T>(List<T> items)` | `Traversal<Holder, T>` in a static method that never declared `T` | the method carries the record's type variables, as `@GenerateLenses` and `@GenerateFocus` already did |
| `record Claimed<F>(List<F> items)` | the effect variable `F` shadowed by the record's own | the effect takes the first free name, `F1` |

A plain class is still created by name, so an array stays wide enough for whatever the traversal produced and the common shape's generated source is unchanged.

## Two things that fell out of it

**The map-value generators delegate.** The entry list they built could not be typed at the resolved element type at all, so both hand the field to `Traversals.traverseMapValues`, which already takes its function as `? super V`. Four statements become one:

```diff
-final var sourceEntries = new ArrayList<>(source.index().entrySet());
-final Function<Map.Entry<String, List<Integer>>, Kind<F, Map.Entry<String, List<Integer>>>> entryF =
-        entry -> applicative.map(newValue -> Map.entry(entry.getKey(), newValue), f.apply(entry.getValue()));
-final var effectOfEntries = Traversals.traverseList(sourceEntries, entryF, applicative);
-final var effectOfMap = applicative.map(
-        newEntries -> newEntries.stream().collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue)),
-        effectOfEntries);
+final var effectOfMap = Traversals.traverseMapValues(source.index(), f, applicative);
 return applicative.map(newMap -> new NestedGenerics(newMap, ...), effectOfMap);
```

**That helper was quadratic.** It copied the map it was building on every entry. It traverses the values and zips the keys back now:

| entries | before | after |
|---|---|---|
| 1 000 | 8.9 ms | 0.6 ms |
| 4 000 | 61.9 ms | 1.6 ms |
| 16 000 | 851.3 ms | 1.3 ms |
| 64 000 | — | 2.5 ms |

`Traversals.forMapValues()` was already on that path, so it gets the same fix. `traverseList` now takes `List<? extends A>` and `Function<? super A, ? extends Kind<F, ? extends B>>`, matching `traverseSet`, `traverseArray` and `traverseMapValues` — which lets the first two drop an unchecked cast each — and the container half is what lets a wildcard-typed field be handed over directly.

## Why the tests did not catch it

Every container in the `@GenerateTraversals` tests was parameterised with a concrete type. `GeneratedTraversalSourceCompilesTest` now compiles every shape through the real processor — JDK, HKJ and third-party containers, wildcards in and outside the focused argument, raw, nested, arrays, generic records — and asserts three things of what comes back:

- no error is reported;
- a method is generated for every component that has one to generate;
- each method focuses the type that component's argument resolves to.

The last two matter because emitting no method at all compiles perfectly well, and so does a traversal that has quietly widened its focus to `Object`. Both assertions were mutation-checked against a deliberately wrong expectation.

`GeneratedTraversalBehaviourTest` runs the generated classes through a classloader — compiling is not traversing — over a wildcard list, a wildcard map, a concrete map, a primitive array, an array of parameterised elements and a generic record.

## Behaviour change, narrow

A `Map` component's traversal no longer throws on a null **key** in the source map, nor on a null the function hands back. The old body rebuilt the map through `Map.entry` and `Collectors.toMap`, which reject both — the key too, though a values traversal never touches it, and both only at the rebuild, after the function had already run over every entry.

Measured on a `Map<String, String>` component, old body against new:

| | old | new |
|---|---|---|
| null value in source, function returns non-null | `{a=x, b=x}` | `{a=x, b=x}` |
| null key in source | `NullPointerException` | `{null=x, b=x}` |
| function returns null | `NullPointerException` | `{a=null}` |

The `List`, `Set` and array siblings have always stored whatever the function returned, null included, so the map was the odd one out — and only because of how it rebuilt itself. Of the zero-or-one containers, `Optional`, `Maybe` and `Validated` reject a null replacement in their own constructors while `Either` and `Try` accept one; none of that is changed here.

## Verification

- `gradle clean build` and `gradle :hkj-examples:bookVerify` green from clean.
- hkj-processor-plugins at 100% line and branch; every file this touches fully covered in its own module.
- Docs: [Wildcard Element Types](https://higher-kinded-j.github.io/optics/traversals.html) and Generic Records added to the traversals chapter, both snippets under the book compile gate; the `@GenerateFocus` rejection rule now says whose rule it is, in the book and in the `hkj-optics` skill; SPI authors are told to read type arguments through `getTypeArgumentName` and to construct through `recordTypeName`.

https://claude.ai/code/session_01LWKYixRrJz8b5uCeeCs9id

## Summary by CodeRabbit

* **New Features**
  * Generated traversals now support wildcard and raw container types, generic records, primitive arrays, and parameterized arrays.
  * Traversals preserve generic record type information and correctly resolve wildcard focus types.
  * Map-value traversal supports null keys and values while rebuilding maps more efficiently.
  * List traversal accepts broader compatible list and function types.

* **Documentation**
  * Expanded guidance and examples for wildcard containers, arrays, generic records, map traversal, and custom generator plugins.
  * Updated release history and verified examples for the latest traversal improvements.


MagnusSmith:
@coderabbit review

PR #727, opened 2026-08-25
@ViaCopyAndSet: pass the source under the type the copy constructor takes (#723)

### LINKED ISSUE #723: [BUG] @ViaCopyAndSet: copyConstructor is read but never reaches the generated code
`@ViaCopyAndSet(copyConstructor = "...")` is read, threaded through the analyser, stored on `CopyStrategyInfo` — and then discarded. The generator's two branches emit identical code.

`CopyStrategyCodeGenerator.generateCopyAndSetSetter`:

```java
private CodeBlock generateCopyAndSetSetter(CopyStrategyInfo info, TypeMirror sourceType) {
  TypeName sourceTypeName = TypeName.get(sourceType);
  String setter = info.setter();

  // If copyConstructor is specified, use that type; otherwise use source type
  if (!info.copyConstructor().isEmpty()) {
    // Complex case with different copy constructor type - not commonly needed
    return CodeBlock.of(
        "(source, newValue) -> {\n  $T copy = new $T(source);\n  copy.$L(newValue);\n  return copy;\n}",
        sourceTypeName, sourceTypeName, setter);   // <-- sourceTypeName, not the configured type
  }

  return CodeBlock.of(
      "(source, newValue) -> {\n  $T copy = new $T(source);\n  copy.$L(newValue);\n  return copy;\n}",
      sourceTypeName, sourceTypeName, setter);
}
```

Both arms build from `sourceTypeName`, so the `if` is dead: `info.copyConstructor()` never reaches the emitted source. The comment above it states the opposite.

```mermaid
flowchart LR
    A["@ViaCopyAndSet(<br/>copyConstructor = &quot;Base&quot;)"] --> B["SpecInterfaceAnalyser<br/>reads the attribute"]
    B --> C["CopyStrategyInfo<br/>carries the value"]
    C --> D{"generateCopyAndSetSetter<br/>if (!copyConstructor.isEmpty())"}
    D -->|"configured"| E["new SourceType(source)"]
    D -->|"default"| F["new SourceType(source)"]
    E --> G["identical output:<br/>the value is dropped here"]
    F --> G

    class A,B,C wire
    class D,F domain
    class E,G error
```

## Why the coverage gate did not catch it

Two tests reach the branch, and neither asserts anything that distinguishes it from the default.

`CopyStrategyCodeGeneratorTest.shouldGenerateCopyAndSetSetterWithCopyConstructor` configures `"PersonCopy"` and then asserts the source type appears:

```java
CopyStrategyInfo info = CopyStrategyInfo.forCopyAndSet("PersonCopy", "setName");
...
// When copyConstructor is non-empty, the branch at line 164-175 is triggered
assertThat(code.setter()).contains("com.test.Person copy = new com.test.Person(source)");
```

It pins the buggy output as expected behaviour. `SpecMutationKillingTest.viaCopyAndSetWithSpecifiedCopyConstructor` is titled "uses it" but asserts only `contains("setValue(newValue)")`, which the default branch satisfies equally.

Both were written to enter the branch rather than to check what it does, so branch coverage stayed at 100% over a no-op.

## What the attribute is supposed to mean

Per the annotation javadoc, it names the copy constructor's **parameter** type:

> If empty (the default), assumes a copy constructor that takes an instance of the source type: `new SourceType(source)`. Specify a fully qualified class name if the copy constructor takes a different type (e.g., a base class or interface).

Worth noting before implementing: for the stated use case it is close to redundant. If `Custom` declares only `Custom(Base other)`, then `new Custom(source)` already compiles by widening. The value only does work when overloads must be disambiguated, where the emission would need a cast on the argument rather than a different constructed type:

```java
(source, newValue) -> {
  SourceType copy = new SourceType((Base) source);   // the configured type as a cast
  copy.setField(newValue);
  return copy;
}
```

Naming it as the *constructed* type cannot be right: the lens is `Lens<S, A>`, so the setter has to return `S`.

## Impact

Low in practice. The default path is correct and is what every in-repo usage takes; a caller setting `copyConstructor` gets working optics that quietly ignore the setting rather than broken ones. The cost is a documented public attribute that does nothing, plus two tests asserting that it does nothing.

## Options

1. **Implement it as an argument cast** (`new SourceType((ParamType) source)`) and rewrite both tests to assert the configured type actually appears. Keeps the published API.
2. **Remove `copyConstructor()`** from the annotation. It has shipped since v0.4.4, so this is a breaking change and would want to ride a `Target 0.5.0` cut alongside the other removals.
3. **Keep and document as inert.** What the book does today, but it leaves a trap in the public API.

Option 1 looks right if the overload-disambiguation case is real; option 2 if it is not. Either way the two tests need to assert behaviour rather than reach for the branch.

## Meanwhile

`optics/copy_strategies.md` no longer presents it as a working override. The parameter listing now reads:

```java
@ViaCopyAndSet(
    copyConstructor = "",     // not honoured yet: the source type's copy constructor is always used
    setter = "setHost")       // required
```

Found while addressing review feedback on #713.

MagnusSmith:
Closes #723.

`@ViaCopyAndSet(copyConstructor = ...)` was read, carried through the analyser and stored on `CopyStrategyInfo` — and then dropped. The generator's two arms built the same string:

```java
// If copyConstructor is specified, use that type; otherwise use source type
if (!info.copyConstructor().isEmpty()) {
  // Complex case with different copy constructor type - not commonly needed
  return CodeBlock.of(
      "(source, newValue) -> {\n  $T copy = new $T(source);\n  copy.$L(newValue);\n  return copy;\n}",
      sourceTypeName, sourceTypeName, setter);   // <-- sourceTypeName, not the configured type
}

return CodeBlock.of(
    "(source, newValue) -> {\n  $T copy = new $T(source);\n  copy.$L(newValue);\n  return copy;\n}",
    sourceTypeName, sourceTypeName, setter);
```

The `if` was dead, so a spec that named a copy constructor parameter type quietly got the default. The comment above it said the opposite.

Two tests reached the branch without asserting anything the default arm does not also satisfy — one pinned the *source* type as the expected output, the other checked only `setValue(newValue)` — so branch coverage stayed at 100% over a no-op.

## The rule

Per the annotation, the attribute names the copy constructor's **parameter** type. It cannot name the constructed type: the setter is `(S, A) -> S`, so what is built is always `S`.

```mermaid
flowchart TD
    A["@ViaCopyAndSet(copyConstructor = N)"] --> E{"N empty?"}
    E -->|yes| P["new S(source)"]
    E -->|no| R{"N resolves<br/>to a type?"}
    R -->|no| X1["rejected at<br/>the declaration"]
    R -->|yes| S{"does S extend or<br/>implement it?"}
    S -->|no| X2["rejected at<br/>the declaration"]
    S -->|yes| I["the supertype as instantiated,<br/>Holder&lt;String&gt; not raw Holder"]
    I --> Q{"same type as S?"}
    Q -->|yes| P
    Q -->|no| C["new S((Holder&lt;String&gt;) source)"]

    class A,I wire
    class E,R,S,Q decision
    class P,C domain
    class X1,X2 error
```

This is **option 1** of the three in the issue. Option 2 (remove the attribute) has to ride the 0.5.0 cut — it shipped in 0.4.4 and this is 0.4.10-SNAPSHOT.

## Where the cast earns its keep

The issue asked whether the overload-disambiguation case is real. It is, and it is the only one: a lone `Config(BaseConfig other)` already takes the source by widening, so the default emission covers it. What the default cannot express is a call that resolves to the wrong overload, or to none:

```java
public class Config extends BaseConfig implements Auditable {
    public Config(BaseConfig other) { ... }
    public Config(Auditable other) { ... }   // new Config(source) is ambiguous
    public void setHost(String host) { ... }
}
```

```java
@ViaCopyAndSet(copyConstructor = "com.example.BaseConfig", setter = "setHost")
Lens<Config, String> host();
```

```java
(source, newValue) -> {
  Config copy = new Config((BaseConfig) source);
  copy.setHost(newValue);
  return copy;
}
```

## Two things the naive emission would get wrong

| shape | naive `(N) source` | what is emitted |
|---|---|---|
| a generic base — `Labelled extends Holder<String>`, `Labelled(Holder<String> other)` | `(Holder) source` — a raw cast, so an unchecked conversion at the constructor call, which this repo compiles as an **error** under `-Xlint:unchecked,rawtypes -Werror` | the supertype as `S`'s own `extends` clause instantiates it: `(Holder<String>) source` |
| the attribute naming `S` itself | `(Config) source` — javac reports a redundant cast under `-Xlint:cast` | no cast; the value is still honoured, it just says nothing |

Both decisions are made with `Types` in the analyser, as every other same-type question in this processor is, which leaves the generator a single rule: a null parameter type passes the source unchanged.

## Rejected at the declaration, not in the generated file

A cast has to **compile**, not merely be legal. Four things are checked where the attribute is written, rather than left to javac to report inside a file the user did not author — the failure mode #721 and #718 were both about:

| written | rejected because |
|---|---|
| `copyConstructor = "BaseConfig"` | does not resolve — the attribute is a plain string, not read against the interface's imports, and takes no type arguments |
| `copyConstructor = "java.lang.Thread"` | `Node` does not extend or implement it |
| `copyConstructor = "com.external.Hidden"` | not public, so the generated class cannot name it — even though `new Node(source)`, which never names it, would have compiled |
| `copyConstructor = "java.io.Serializable"` | a genuine supertype, but no `Node` constructor accepts one; the message lists the single-argument constructors that do exist |

A varargs copy constructor counts as accepting it, since a varargs invocation with one argument is what the generated call is.

A fifth shape goes the other way. A hierarchy holding a type this round cannot resolve reads as having *no* supertypes, so a good name was rejected and that error landed beside javac's own `cannot find symbol`. The compiler is now asked directly before anything is rejected, so an unreadable hierarchy costs the instantiation rather than drawing a second, misleading error.

## One error per problem, for all three siblings

The missing-strategy diagnostic moved into `parseCopyStrategy` so a rejected value draws one error rather than also drawing "requires a copy strategy annotation". The same bug was live in both sibling parse methods and is fixed with it: a rejected `@InstanceOf` target and an undetectable `@ThroughField` container each drew their own error **and** a misleading "requires a … hint annotation" from the caller. Each parse method now reports where the problem is found, and all three assert `hadErrorCount(1)`.

## The hazard the processor cannot check

Naming a supertype selects the constructor that takes it, and that constructor can only copy what it can see. If `balance` is declared on `Ledger` and you name `LedgerBase`, every `set` returns a copy with `balance` back at its default — a perfectly *deep* copy of everything in scope. Nothing can detect that, so it is documented beside the other law hazards, in the annotation's warning list and the book's admonition, and the release note carries a `_Behaviour change, both ways:_` paragraph: a spec that already sets the attribute silently changes which constructor runs, and names that used to be discarded are now rejected.

## Tests

Both tests that entered the dead branch now assert the configured type reaches the emitted source, and the end-to-end set runs the real processor:

| test | pins |
|---|---|
| `shouldCastToNamedCopyConstructorParameterType` | `new Node((Base) source)`, against a source with both `Node(Base)` and `Node(Node)` |
| `shouldNameGenericCopyConstructorParameterType` | `new Labelled((Holder<String>) source)` — the instantiation, not the raw name |
| `shouldOmitCastWhenCopyConstructorNamesSourceType` | `new Node(source)`, and *not* `(Node) source` |
| `shouldAcceptVarargsCopyConstructor` | a varargs copy constructor is accepted, a non-matching one is not |
| four `shouldReject*` tests | each rejection, each asserting `hadErrorCount(1)` |
| `shouldNotRejectWhenASupertypeIsUnresolved` | the deferral — javac's error alone, not ours as well |
| `viaCopyAndSetWithSpecifiedCopyConstructor` | was titled "uses it" while asserting only `setValue(newValue)`; now asserts the cast |

The generator unit test had pinned output that is not legal Java — it cast a `record` to an interface the record does not implement, which JLS 5.5.1 forbids — so it asserted a string rather than behaviour. It now names a supertype the source really has.

`@ViaCopyAndSet` had no lens-law coverage at all, and the fixture's two constructors copied identically, so no test could see which one ran. `SpecGeneratedOpticLawsTest` now runs the laws against a total copy constructor, and a second spec pins the narrowing case — `balance` 100 → 0 when the named base cannot reach it.

`SpecInterfaceAnalyser` and `CopyStrategyCodeGenerator` are at 100% line, branch and instruction, pinned in `hkj-processor/build.gradle.kts`.

## Also in here

The shipped `hkj-optics` skill described the copy strategies with three uncompilable examples — including `@ViaCopyAndSet(copy = "clone")`, which is not an element and teaches the attribute as a method name when it names a type. It installs into consumer projects, so it is fixed here. The four rejections have `compiler_errors.md` entries, alongside the missing-strategy heading, which never matched the message emitted. The book's worked example is now compile-gated, which needed the example types in main sources and raises the snippet floor to 309.

Found while reviewing this: #728, a pre-existing `ClassCastException` on a generic spec interface. Not fixed here — line 115 is currently the only thing keeping a type variable out of the supertype walk.

https://claude.ai/code/session_01783mM7hfLGeUoULcFW1XtV

## Summary by CodeRabbit

- **New Features**
  - `@ViaCopyAndSet` now supports selecting among overloaded copy constructors by specifying the constructor’s parameter type.
  - Generated updates correctly cast source values when copying through a selected supertype.
  - Added validation for unresolved, invalid, inaccessible, or incompatible constructor types.

- **Documentation**
  - Expanded copy-strategy guidance, examples, parameter references, and troubleshooting diagnostics.
  - Added examples covering inherited state and potential data loss when using narrower supertypes.
  - Added release-history notes describing the updated behavior.


MagnusSmith:
@coderabbit review

MagnusSmith:
@coderabbit review

MagnusSmith:
@coderabbitai help

MagnusSmith:
@coderabbitai full review

MagnusSmith:
@coderabbitai full review

Issue #733, opened 2026-08-26
[BUG] An @InstanceOf prism onto a parameterised target cannot check its type argument


MagnusSmith:
An `@InstanceOf` prism whose target is parameterised narrows through a raw `instanceof`, which erasure leaves unable to test the type argument. The generated prism reports a match for *any* instantiation and hands it back under the one the caller asked for.

```java
public class Shape {}
public class Circle<T> extends Shape { public T tag() { return null; } }

@ImportOptics
public interface ShapeOpticsSpec<T> extends OpticsSpec<Shape> {
    @InstanceOf(Circle.class)
    Prism<Shape, Circle<T>> circle();
}
```

generates

```java
public static <T> Prism<Shape, Circle<T>> circle() {
  return Prism.of(
    source -> source instanceof Circle t ? Optional.of(t) : Optional.empty(),
    subtype -> subtype);
}
```

so

```java
ShapeOptics.<String>circle().getOptional(new Circle<Integer>(42))
```

returns a populated `Optional<Circle<String>>` holding a `Circle<Integer>`, and reading `.tag()` throws:

```
java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.String
```

## Why it is not simply fixable

`@InstanceOf` takes a `Class<?>` constant, which carries no type arguments — `Circle.class` is the raw class. Erasure means no generated implementation can check `T` at runtime. The optic's type says more than the test behind it can support.

```mermaid
flowchart TD
    A["@InstanceOf(Circle.class)<br/>Prism&lt;Shape, Circle&lt;T&gt;&gt;"] --> B["class constant carries<br/>no type arguments"]
    B --> C["source instanceof Circle<br/>raw test"]
    C --> D{"matches?"}
    D -->|"any instantiation"| E["returned as Circle&lt;T&gt;"]
    E --> F["ClassCastException<br/>at the first read"]

    classDef wire fill:#8caaee,stroke:#1e66f5,color:#232634
    classDef decision fill:#e5c890,stroke:#df8e1d,color:#232634
    classDef domain fill:#a6d189,stroke:#40a02b,color:#232634
    classDef error fill:#e78284,stroke:#d20f39,color:#232634
    class A wire
    class B,C,E domain
    class D decision
    class F error
```

## What already exists

[...]

Issue #732, opened 2026-08-26
[BUG] @ViaCopyAndSet(copyConstructor) is unusable on a spec interface with a generic source type


MagnusSmith:
`@ViaCopyAndSet(copyConstructor = ...)` cannot be used on a spec interface whose source type is generic. The constructor lookup compares a supertype derived from the **instantiated** source type against a constructor parameter read **as declared**, so the two never match and the spec is rejected outright.

```java
public class Base<X> { protected String v; }

public class Holder<X> extends Base<X> {
    public Holder() {}
    public Holder(Base<X> other) { this.v = other.v; }
    public String getV() { return v; }
    public void setV(String v) { this.v = v; }
}

@ImportOptics
public interface HolderOpticsSpec<U> extends OpticsSpec<Holder<U>> {
    @ViaCopyAndSet(setter = "setV", copyConstructor = "com.external.Base")
    Lens<Holder<U>, String> v();
}
```

```
@ViaCopyAndSet: copyConstructor names 'com.external.Base', which no constructor of
'com.external.Holder<U>' accepts. ... Found single-argument constructors taking
[com.external.Base<X>].
```

The diagnostic displays the mismatch itself — `Holder<U>` against `Base<X>` — and reads to the user as a puzzle, because the constructor it says is missing is right there.

## Where

`SpecInterfaceAnalyser.hasConstructorAccepting`, around `:712-731`, reached from `:596`.

`resolveCopyConstructorParameterType` correctly derives the cast target from the instantiated `sourceType`: the supertype walk substitutes, giving `Base<U>`. The lookup then reads the constructor as declared:

```java
TypeMirror parameterType = parameters.getFirst().asType();   // Base<X> — the declaration's variable
if (typeUtils.isAssignable(argument, parameterType)) {       // isAssignable(Base<U>, Base<X>) is false
```

`argument` speaks the spec's variables; `parameterType` speaks the source type's. They cannot agree.

```mermaid
flowchart TD
    S["OpticsSpec&lt;Holder&lt;U&gt;&gt;"] --> A["supertype walk substitutes<br/>argument = Base&lt;U&gt;"]
    H["class Holder&lt;X&gt;<br/>Holder(Base&lt;X&gt; other)"] --> P["parameter read as declared<br/>parameterType = Base&lt;X&gt;"]
    A --> C{"isAssignable?"}
    P --> C
    C -->|"never"| R["rejected: no constructor accepts it"]

[...]

Issue #731, opened 2026-08-26
[BUG] @GenerateIsos on a generic method generates source naming an undeclared type variable


MagnusSmith:
`@GenerateIsos` on a generic method generates source that names a type variable it never declares. The compile fails inside the generated file with a bare javac error, and nothing points at the annotated method.

```java
public record Box<T>(T value) {}

public final class BoxConverters {
    @GenerateIsos
    public static <T> Iso<Box<T>, T> boxToValue() {
        return Iso.of(Box::value, Box::new);
    }
}
```

```
error: cannot find symbol
  symbol:   class T
  location: class com.example.BoxConvertersIsos
```

## Where

`IsoProcessor.java:86-96`. The processor reads the two `Iso` type arguments off the method's return type and emits them into a `public static final` **field**. `method.getTypeParameters()` is never consulted — `getTypeParameters()` does not appear anywhere in the file.

A field cannot declare type variables, so unlike its siblings there is nothing to resolve to.

```mermaid
flowchart TD
    M["@GenerateIsos<br/>static &lt;T&gt; Iso&lt;Box&lt;T&gt;, T&gt; boxToValue()"] --> R["read the Iso type arguments<br/>Box&lt;T&gt; and T"]
    R --> F["emit public static final<br/>Iso&lt;Box&lt;T&gt;, T&gt; BOX_TO_VALUE"]
    F --> X["a field cannot declare &lt;T&gt;<br/>cannot find symbol"]

    classDef wire fill:#8caaee,stroke:#1e66f5,color:#232634
    classDef domain fill:#a6d189,stroke:#40a02b,color:#232634
    classDef error fill:#e78284,stroke:#d20f39,color:#232634
    class M wire
    class R,F domain
    class X error
```

## The family this belongs to

Every other processor that generates parameterised code either declares the parameters or rejects them at the declaration. `IsoProcessor` does neither:

| processor | on a generic declaration |
|---|---|
| `LensProcessor`, `TraversalProcessor`, `FocusProcessor`, `NavigatorClassGenerator`, `MappingProcessor` | declares them |
| `MergeProcessor:112`, `AssemblyProcessor:87`, `ErrorEnvelopeProcessor:121` | rejects at the declaration |
| `@ImportOptics` spec interfaces | declares them ([#730](https://github.com/higher-kinded-j/higher-kinded-j/issues/730)) |
| **`IsoProcessor`** | **neither — emits uncompilable source** |

## Suggested fix

[...]

Issue #728, opened 2026-08-25
[BUG] @ImportOptics on a generic spec interface crashes with ClassCastException instead of reporting


MagnusSmith:
`@ImportOptics` on a generic spec interface crashes the processor with a `ClassCastException` instead of reporting a diagnostic. The compile fails with a stack trace naming javac internals, and nothing points at the declaration that caused it.

```java
@ImportOptics
public interface BoxOpticsSpec<S extends Box> extends OpticsSpec<S> {

    @ViaCopyAndSet(setter = "setV")
    Lens<S, String> v();
}
```

```
java.lang.ClassCastException: class com.sun.tools.javac.code.Symbol$TypeVariableSymbol
cannot be cast to class javax.lang.model.element.TypeElement
```

## Where

`SpecInterfaceAnalyser.analyse` reads the source type out of `OpticsSpec<S>` and casts its element straight to `TypeElement`:

```java
TypeMirror sourceType = extractSourceType(specInterface);      // a TypeVariable here
...
TypeElement sourceTypeElement = (TypeElement) typeUtils.asElement(sourceType);   // <-- line 115
if (sourceTypeElement == null) {
  error("Source type " + sourceType + " is not a valid type element", specInterface);
  return Optional.empty();
}
```

The null guard below it is written for the array/primitive case, where `Types.asElement` returns null. A type variable is the third case: `asElement` returns a `TypeParameterElement`, which is **not** a `TypeElement`, so the cast throws before the guard is reached.

```mermaid
flowchart TD
    S["OpticsSpec&lt;S&gt; type argument"] --> K{"what kind?"}
    K -->|"declared type"| E["TypeElement — analysis proceeds"]
    K -->|"array, primitive"| N["asElement returns null<br/>→ guarded, reports a diagnostic"]
    K -->|"type variable"| T["asElement returns TypeParameterElement<br/>→ ClassCastException"]

    classDef wire fill:#8caaee,stroke:#1e66f5,color:#232634
    classDef decision fill:#e5c890,stroke:#df8e1d,color:#232634
    classDef domain fill:#a6d189,stroke:#40a02b,color:#232634
    classDef error fill:#e78284,stroke:#d20f39,color:#232634
    class S wire
    class K decision
    class E,N domain
    class T error
```

## Scope

[...]

PR #734, opened 2026-08-26
@ImportOptics: declare the type parameters the generated signature names (#730)

### LINKED ISSUE #730: [BUG] A generic OpticsSpec source type generates method type parameters from the source type's declaration, not the spec's
A spec interface whose source type is generic generates a method signature whose type parameters come from the **source type's own declaration**, not from the variables free in the source type. When the two happen to agree it works; when they do not, the generated code either fails to compile or silently carries an uninferable parameter.

Found by the review panel on #729, which needed to document what "a generic source type is supported" actually means. Pre-existing, and independent of the `@ImportOptics` guard that PR adds.

## Where

`SpecInterfaceGenerator.java:154`

```java
// Add type parameters if source type has them
for (TypeParameterElement typeParam : sourceTypeElement.getTypeParameters()) {
  methodBuilder.addTypeVariable(TypeVariableName.get(typeParam));
}
```

`sourceTypeElement` is `Box` — so the loop copies `Box`'s declared `<T>`, whatever the spec wrote. The rest of the signature is built from `sourceType`, the *instantiated* `Box<U>`. The two only line up by name coincidence.

```mermaid
flowchart TD
    D["interface BoxOpticsSpec&lt;U&gt;<br/>extends OpticsSpec&lt;Box&lt;U&gt;&gt;"] --> S["sourceType = Box&lt;U&gt;<br/>the instantiated type"]
    D --> E["sourceTypeElement = Box<br/>the declaration"]
    S --> R["return + receiver types<br/>Lens&lt;Box&lt;U&gt;, String&gt;"]
    E --> P["method type parameters<br/>&lt;T&gt;, copied from Box"]
    R --> G["static &lt;T&gt; Lens&lt;Box&lt;U&gt;, String&gt; label()"]
    P --> G
    G --> X["U is not in scope<br/>cannot find symbol"]

    class D wire
    class S,E,R,P,G domain
    class X error
```

## Two symptoms

Both reproduced by compiling them against `public class Box<T>` / `public class Pair<A, B>`.

### 1. Renamed parameter — generated source does not compile

```java
@ImportOptics
public interface BoxOpticsSpec<U> extends OpticsSpec<Box<U>> {
    @Wither("withLabel")
    Lens<Box<U>, String> label();
}
```

`BoxOptics` is written, then javac fails **inside the generated file**:

```
error: cannot find symbol
  symbol: class U
  location: class com.myapp.BoxOptics
```

generated: `public static <T> Lens<Box<U>, String> label()`

The control, `interface BoxOpticsSpec<T> extends OpticsSpec<Box<T>>`, succeeds — the only difference is the letter.

### 2. Arity mismatch — compiles, and is silently wrong

```java
@ImportOptics
public interface PairOpticsSpec<A> extends OpticsSpec<Pair<A, String>> {
    @Wither("withLeft")
    Lens<Pair<A, String>, A> left();
}
```

Compiles. Generates:

```java
public static <A, B> Lens<Pair<A, String>, A> left() { ... }
```

`B` is copied from `Pair`'s declaration, occurs nowhere in the signature, and is uninferable. Callers get `Object` or must write `PairOptics.<String, Object>left()`. Nothing reports anything.

This is the worse half: symptom 1 at least fails loudly.

## Why it matters beyond the two shapes

`TraverseWith.java:44` javadoc shows `interface TreeOptics<T> extends OpticsSpec<Tree<T>>` as the way to write a generic spec. It is correct as written, and one rename from symptom 1.

## Suggested fix

Read the parameters from the type the optics are actually generated against, rather than from its declaration — the free variables of `sourceType`, in the order the spec declares them. `interface PairOpticsSpec<A> extends OpticsSpec<Pair<A, String>>` should generate `static <A> Lens<Pair<A, String>, A> left()`, and the renamed case should just work.

Where that is not wanted, the alternative is to reject the mismatch at the declaration, in the what/why/fix shape the analyser now uses — the spec's type parameters must match the variables free in `sourceType`. That is the cheaper change but it is a **breaking one**: symptom 2 compiles today, so anyone relying on it would start failing. Substituting correctly only ever turns broken or useless output into working output, which is why it is the suggestion above.

Either way the guard belongs in `SpecInterfaceAnalyser` beside `reportUnusableSourceType`, so nothing reaches the generator that the generator cannot write.

## Scope note

Deliberately left out of #729, which fixes the `ClassCastException` on a *bare* type variable. That one crashes; this one has a different root cause in a different class, and the silent-wrong variant needs a compatibility decision rather than a bug fix. #729 narrows the book's claim to the name-and-arity-matched case and says the mismatch is not currently diagnosed.

MagnusSmith:
Fixes #730.

A generated optic method took its type parameters from the **source type's own declaration** while building the rest of the signature from the **instantiated** source type. The two agreed only by name coincidence.

```java
// SpecInterfaceGenerator, before
for (TypeParameterElement typeParam : sourceTypeElement.getTypeParameters()) {
  methodBuilder.addTypeVariable(TypeVariableName.get(typeParam));
}
```

`sourceTypeElement` is `Box` — so this copies `Box`'s declared `<T>`, whatever the spec wrote. Everything else is built from `sourceType`, the instantiated `Box<U>`.

```mermaid
flowchart TD
    D["interface BoxOpticsSpec&lt;U&gt;<br/>extends OpticsSpec&lt;Box&lt;U&gt;&gt;"] --> S["sourceType = Box&lt;U&gt;<br/>instantiated"]
    D --> E["sourceTypeElement = Box<br/>the declaration"]
    S --> R["return and receiver types<br/>Lens&lt;Box&lt;U&gt;, String&gt;"]
    E --> P["type parameters<br/>&lt;T&gt;, copied from Box"]
    R --> G["static &lt;T&gt; Lens&lt;Box&lt;U&gt;, String&gt; label()"]
    P --> G
    G --> X["U is not in scope<br/>cannot find symbol"]

    class D wire
    class S,E,R,P,G domain
    class X error
```

## Four broken shapes, not the two on the issue

I wrote #730 from two symptoms. Probing every shape turned up four, and the two I had not seen are the ones most likely to be hit:

| spec | before | after |
|---|---|---|
| `<T>` / `OpticsSpec<Box<T>>` | `<T>` | unchanged — it was right by coincidence |
| `<U>` / `OpticsSpec<Box<U>>` | `cannot find symbol` | `static <U> Lens<Box<U>, String>` |
| `<A>` / `OpticsSpec<Pair<A,String>>` | `<A, B>`, `B` uninferable | `static <A> Lens<Pair<A,String>, A>` |
| **`OpticsSpec<Box<String>>`** | `<T>` uninferable | `static Lens<Box<String>, String>` |
| **`<T>` / `OpticsSpec<Shape>`, `Prism<Shape, Circle<T>>`** | `cannot find symbol` | `static <T> Prism<Shape, Circle<T>>` |

Row four is a concrete instantiation of a generic type — an entirely ordinary thing to write, silently broken. Row five is not about a generic *source* at all: a prism whose focus names a parameter the source does not, where reading the source's declaration finds nothing to declare.

## The fix

Parameters come from the **spec**, filtered to those the signature reaches, in declaration order. One subtlety: a parameter reached only through a kept parameter's bound has to come too, or `<T, V extends List<T>>` focused through `V` leaves `T` undeclared — the same fault one level down.

The "does this type name that parameter" walk is the one #728 added to keep a circular bound out of a diagnostic, so it moved to `ProcessorUtils` rather than being written twice. It gained wildcard bounds on the way, which closed a hole on the #728 side: `<S extends List<? extends S>>` was answered with `OpticsSpec<List<? extends S>>`, as circular as the enclosing-type case fixed there.

## Second commit: what the panel found

A five-lens review (correctness, type safety, immutability, ergonomics, architecture) ran after the first commit. It found the same defect one step further in, and a shape the fix newly enabled that a consuming build cannot compile. Every case below was reproduced by compiling it.

**The fix was incomplete.** A parameter declared on the optic *method* was never a candidate, so `<X> Lens<Box<String>, X> content()` still emitted `X` into a signature that never declared it. Declaring it only moves the failure to `incompatible types` — the source type is fixed by `OpticsSpec<S>`, so a method-level parameter appears only in the focus and nothing can ever bind it. It is now rejected at the declaration, beside the no-value-parameters rule.

**The fix introduced a regression.** An `@InstanceOf` prism onto a parameterised target — row five above, which did not compile at all before — narrows through a raw `instanceof`. It compiled only where `-Werror` was off:

```
warning: [rawtypes] found raw type: com.external.Circle
warning: [unchecked] unchecked conversion
warning: [unchecked] unchecked method invocation: method of in interface Prism
```

Both lints are now answered where they arise, as the `@ThroughField` traversal already did. `@MatchWhen` narrows through the source type's own getter, needs none of it, and a test pins that the suppression stays precise rather than blanket.

**Also fixed**

- The **class javadoc** was the last read of the source type's own declaration, rendering `{@link Box<T>}` beside methods that say `<U>`; doclint rejects the reference. It goes through `ClassName`, which every other generator in the package already does and which the `PairLenses` golden pins.
- `mentions` gained the **intersection arm** its javadoc already promised — javac's intersection implements `DeclaredType`, so the declared arm swallowed it and missed every bound it names. It also takes an `Element`, so the one cast the first commit added at the call site goes.
- `mentions` and `hasTypeArguments` are pinned by a **contract test against real mirrors**, rather than only through their callers.
- The book stated a rule the bound closure deliberately breaks ("declares exactly the parameters it names" — `<T, V extends List<T>>` keeps `T`, which the signature does not name).
- What a generated method declares now has a **section in the chapter that teaches spec interfaces**, rather than a paragraph trailing an unrelated compiler-error entry.

## Behaviour change

`<A, B>` becoming `<A>`, and `<T>` becoming none, is an **arity change**. Both compiled in 0.4.9, so a call site that supplied the witnesses explicitly — `PairOptics.<String, Object>left()` — stops compiling. Dropping the explicit arguments restores it; the inferred result is unchanged. A spec that renamed the source type's parameters never compiled at all, so nothing can depend on that.

This has a `release-history.md` entry with the `_Behaviour change, source-breaking_` call-out, matching what #723, #721, #719 and #711 each did in their own commits.

## Split out rather than folded in

Three pre-existing bugs the sweep confirmed, each verified by compiling it:

- **#731** — `@GenerateIsos` on a generic method emits an undeclared type variable. The last processor in the family that neither declares parameters nor rejects them.
- **#732** — `@ViaCopyAndSet(copyConstructor)` is unusable on any generic spec: it compares an instantiated supertype against a declared parameter, so they never match. Relevant here because this PR documents generic source types as supported, which is true of every strategy except that one.
- **#733** — an `@InstanceOf` prism onto a parameterised target cannot check its type argument; demonstrated `ClassCastException`. Filed precisely because the suppression above makes the hazard silent rather than warned.

`MappingProcessor`, `MergeProcessor`, `AssemblyProcessor`, `ErrorEnvelopeProcessor` and `EffectAlgebraProcessor` were each probed and are clean — `MappingProcessor` already reads off the spec, which is direct precedent for this fix's direction.

## Verification

All 47 `getTypeParameters()` / `addTypeVariable` sites in the processor were swept. `ProcessorUtils`, `SpecInterfaceAnalyser` and `SpecInterfaceGenerator` are all at zero uncovered lines and branches. `clean build` across all modules: BUILD SUCCESSFUL, EXIT=0.

## Summary by CodeRabbit

* **New Features**
  * Improved support for generic optics specifications, including propagation of relevant type parameters into generated optic methods.
  * Added support and guidance for parameterized source types such as `Box<U>`.
  * Added clearer handling of wildcard bounds and generic focus types.

* **Bug Fixes**
  * Optic methods that declare their own type parameters now produce a clear validation error.

* **Documentation**
  * Expanded guidance and release notes covering generic type inference, parameter resolution, and `@ImportOptics` behavior.


MagnusSmith:
@coderabbit full review

PR #735, opened 2026-08-26
@ViaCopyAndSet: read the copy constructor under the source type's instantiation (#732)

### LINKED ISSUE #732: [BUG] @ViaCopyAndSet(copyConstructor) is unusable on a spec interface with a generic source type

MagnusSmith:
Fixes #732. Rebased on `main` after #734 (#730) merged.

`@ViaCopyAndSet(copyConstructor = ...)` was unusable on any spec interface with a generic source type. The check derived its argument from the **instantiated** source type — a supertype walk that substitutes — and compared it against a constructor parameter read **as declared**:

```
argument       Base<U>   the spec's variable, from the walk over Node<U>
parameterType  Base<X>   Node's own variable, read off the constructor
```

They speak different declarations, so `isAssignable` was never true and every generic spec was rejected. Reading the parameter through `asMemberOf` rewrites it under the source type's own instantiation, which is the question actually being asked: the emitted call is `new Node<U>((Base<U>) source)`, so what matters is the parameter as `Node<U>` sees it.

**Same name is not the same variable** — a spec that reuses the source type's parameter name was rejected too, since those are distinct type-variable elements. That is what the first test uses, deliberately: holding the name fixed leaves the constructor match as the only thing under test.

`@Wither`, `@ViaConstructor` and `@ViaBuilder` carry names rather than types, so none was affected.

### The rejection message could not be followed

A reviewer followed the fix sentence step by step and got nowhere, twice. It said *"Name a type one of those constructors takes"* and listed `com.external.Other<U>`:

| attempt | result |
|---|---|
| paste `com.external.Other<U>` | rejected — the attribute takes no type arguments |
| strip to `com.external.Other` | rejected — `Node<U>` does not extend it |

`Other` is a **sibling** of the source type, so the attribute can never name it. The sentence now says what the attribute can actually accept — a supertype, named as the class alone — and the list is documented as something to recognise a name in rather than copy from.

### The message misreported the code it generates

It claimed `new Node((Base) source)` while the emitted code is `new Node<U>((Base<U>) source)` — contradicting #723's own release note, which says the cast is parameterised "rather than raw". Both the snippet and the list now go through `ProcessorUtils.simpleTypeName`, the helper written for exactly this. Two sibling `@ViaCopyAndSet` diagnostics rendered their snippets the same way and move with it.

Where `S`'s `extends` clause **pins** the supertype's arguments — `class PNode<X> extends PBase<String>` — naming the supertype was rejected with no way to see why, because the message showed only the name given and never the type `S` reaches it as. It now names both.

### A regression the fix would have shipped

Reading the constructor under the instantiation had been intercepting a wildcard source type **by accident**. With that gone, `OpticsSpec<Node<?>>` was accepted and javac failed inside the generated file with `unexpected type … found: ?` — an error with no position in anything the author wrote, which is precisely what these checks exist to prevent.

The underlying hole is older and wider than `copyConstructor`: both arms of `@ViaCopyAndSet` and `@ViaConstructor` fail the same way, with or without the attribute. So the guard replaces an accident with a diagnostic, and covers the arms that never had one.

It is asked **per strategy**, not of the source type as a whole — `@Wither` rebuilds through a method, needs no constructor, and a wildcard source type works fine there. A test pins that. The same guard rejects an inner-class source type, whose call needs an enclosing instance the generated class cannot reach; a *static* nested type is accepted, and that is pinned too.

### A cast justified by an invariant that does not exist

My comment said `analyse()` "rejects a source type that is not a declared type". It does not: its guard is `asElement(sourceType) instanceof TypeElement`, which admits `DECLARED`, `ERROR` **and** `INTERSECTION` — all three `DeclaredType` on javac. The cast is total, but for a different reason than the one I wrote, and "is a declared type" is exactly the mistake #728 was. The comment now states what the guard establishes, the narrowing happens once, and `hasConstructorAccepting` no longer carries the instantiated type and its element side by side — the shape this whole bug class comes from.

Two branches that no test could reach were removed rather than waived, per the rule the coverage list states.

## What this is one of

`@ViaCopyAndSet` had **no** generic-spec test before this PR, and neither did any other copy strategy except `@Wither` — which is structurally blind to this fault, because it reads no member of the source type. One `@ViaCopyAndSet` case on `OpticsSpec<Node<X>>` would have caught #732 the day #727 landed.

The panel's sweep found three more instances of the same declaration-vs-instantiation confusion, each verified by compiling it. They are filed separately rather than folded in here, along with the missing test axis.

## Verification

Both original tests confirmed as real regression guards — stash the fix, both fail; restore, both pass. `SpecInterfaceAnalyser` stays at 100% line, branch and instruction. `clean build` across all modules: BUILD SUCCESSFUL, EXIT=0.

## Summary by CodeRabbit

- **Documentation**
  - Clarified compiler errors for generic supertypes, constructor compatibility, wildcard source types, and available alternatives.

- **Bug Fixes**
  - Improved handling of generic type substitutions when validating copy constructors.
  - Added clearer diagnostics for incompatible, inaccessible, wildcard, and non-static inner-class source types.
  - Preserved support for compatible varargs constructors and static nested types.

- **Tests**
  - Expanded coverage for generic parameters, wildcard handling, constructor matching, nested classes, and diagnostic messages.

### DIFF
--- a/hkj-book/src/optics/compiler_errors.md
+++ b/hkj-book/src/optics/compiler_errors.md
@@ -87,11 +87,17 @@ A source type that is itself generic is supported, and the spec names its own ty
 
 **Fix.** Name a public supertype, generate into that package with `@ImportOptics(targetPackage = ...)`, or drop the attribute.
 
-### "@ViaCopyAndSet: copyConstructor names '...', which no constructor of '...' accepts"
+### "@ViaCopyAndSet: copyConstructor names '...', which '...' reaches as '...', and no constructor accepts"
 
-**Cause.** The name is a genuine supertype, but no single-argument constructor of `S` takes it — `java.lang.Object` and marker interfaces such as `Serializable` reach this often. The message lists the single-argument constructors that do exist.
+**Cause.** The name is a genuine supertype, but no single-argument constructor of `S` takes the type `S` actually reaches it as — `java.lang.Object` and marker interfaces such as `Serializable` reach this often. The message names both the type you gave and the one `S` reaches, which differ when `S`'s own `extends` clause pins the arguments: `class PNode<X> extends PBase<String>` reaches `PBase` as `PBase<String>`, whatever `X` is.
 
-**Fix.** Name one of the types those constructors take, or drop the attribute. The attribute is only needed when the copy constructor is overloaded — see [Copy Strategies](copy_strategies.md#viacopyandset-legacy-types-with-a-copy-constructor-and-setters).
+**Fix.** Name a supertype of `S` that one of the listed constructors takes, as the class alone without type arguments, or drop the attribute. The list carries type arguments and the attribute does not, so read it to recognise your supertype in it rather than to copy from it — and a listed type that is not a supertype of `S` cannot be named at all. The attribute is only needed when the copy constructor is overloaded — see [Copy Strategies](copy_strategies.md#viacopyandset-legacy-types-with-a-copy-constructor-and-setters).
+
+### "@ViaCopyAndSet: '...' is written with a wildcard type argument"
+
+**Cause.** The source type carries a wildcard, `OpticsSpec<Node<?>>`, and the strategy rebuilds it through a constructor. `new Node<?>(...)` is not something that can be written, whatever the arguments. `@ViaConstructor` reports the same thing for the same reason. An inner class draws the sibling message, because its constructor call needs an enclosing instance the generated class has no way to reach.
+
+**Fix.** Name the type the wildcard stands for, or switch to `@Wither`, which rebuilds through a method and names no constructor — a wildcard source type is no obstacle there.
 
 ### "@MatchWhen: predicate / getter method not found on source"
 
--- a/hkj-processor/src/main/java/org/higherkindedj/optics/processing/external/SpecInterfaceAnalyser.java
+++ b/hkj-processor/src/main/java/org/higherkindedj/optics/processing/external/SpecInterfaceAnalyser.java
@@ -11,6 +11,7 @@ import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.Modifier;
 import javax.lang.model.element.Name;
+import javax.lang.model.element.NestingKind;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.element.VariableElement;
 import javax.lang.model.type.ArrayType;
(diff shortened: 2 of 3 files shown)


MagnusSmith:
@Coderabbit full review

PR #738, opened 2026-08-26
Declaration-vs-instantiation: fix the three remaining sites (#736)

### LINKED ISSUE #736: [BUG] Declaration-vs-instantiation: three more sites read a member as declared and use it against an instantiated type
Three PRs in a row — [#729](https://github.com/higher-kinded-j/higher-kinded-j/pull/729) (#728), [#734](https://github.com/higher-kinded-j/higher-kinded-j/pull/734) (#730), [#735](https://github.com/higher-kinded-j/higher-kinded-j/pull/735) (#732) — have fixed the same mistake in three different classes:

> **A member or type is read *as declared*, then compared with, or emitted alongside, something derived from an *instantiated* type.**

The two vocabularies only agree by name coincidence. A review panel sweeping every `getTypeParameters()` and member-reading site in the processor found three more instances, each confirmed by compiling it.

```mermaid
flowchart TD
    D["the declaration<br/>class Node&lt;X&gt;"] --> P["read as declared<br/>Base&lt;X&gt;, &lt;X&gt;, T"]
    I["the instantiation<br/>Node&lt;U&gt; / Inner&lt;String&gt;"] --> Q["derived by substitution<br/>Base&lt;U&gt;, &lt;U&gt;, List&lt;String&gt;"]
    P --> C{"compared or<br/>emitted together"}
    Q --> C
    C -->|"names agree"| OK["works, by coincidence"]
    C -->|"names differ"| X["cannot find symbol,<br/>or a wrong rejection"]

    class D,I wire
    class P,Q domain
    class C decision
    class OK domain
    class X error
```

### 1. `NavigatorClassGenerator:193` — generated source does not compile

`TypeName.get(targetRecord.asType())` reads the navigable target's **declaration**, while the component reaching it is instantiated. The navigator declares only `<S>`, and `addNavigationMethods:544` reads `targetRecord.getRecordComponents()` as declared too.

```java
@GenerateFocus(generateNavigators = true) record Inner<T>(T value, String label) {}
@GenerateFocus(generateNavigators = true) record Outer(Inner<String> inner, String tag) {}
```

→ 9 × `cannot find symbol: class T, location: class OuterFocus.InnerNavigator<S>`.

`component.asType()` already holds the instantiated type; `targetRecord.asType()` is the wrong thing to reach for. `NavigatorClassGenerator` is on the 100% coverage list, which tells you no test carries a generic navigable target.

### 2. `ExternalLensGenerator:347,388` — generated source does not compile

`createTraversalMethod` uses the raw `ClassName.get(recordElement)` and never calls `addTypeVariable`, while `getFocusType:408` reads the component's declared arguments. Its two siblings in the same class, `createRecordLensMethod:198` and `createWitherLensMethod:243`, both declare the parameters correctly.

```java
public record Bag<T>(String name, List<T> items) {}

@ImportOptics({com.external.Bag.class})
package com.myapp.optics;
```

→ `cannot find symbol: class T, location: class com.myapp.optics.BagLenses`.

This is a **sibling-sweep miss**: [#726](https://github.com/higher-kinded-j/higher-kinded-j/pull/726) fixed exactly this in `TraversalProcessor` (`recordTypeName`, `addTypeVariable`, `effectVariableName`) and left the `@ImportOptics` mirror of the same generator untouched.

### 3. `SpecInterfaceAnalyser.findFieldType:1004` — rejects where resolution is possible

`autoDetectTraversalForField:955` passes only `sourceTypeElement`, and `findFieldType` returns `component.asType()` / `method.getReturnType()` / `field.asType()` as declared. `parseTraversalHint:899` does not even receive `sourceType`, so the signature makes the right answer unreachable.

```java
public class Holder<T> { public T items() { ... } }

interface HolderOpticsSpec extends OpticsSpec<Holder<List<String>>> {
    Traversal<Holder<List<String>>, String> eachItem();
}
```

→ *"Cannot auto-detect traversal for field 'items' of type **'T'**"*

Two faults in one: a rejection where `asMemberOf` resolves the field to `List<String>` and detection succeeds, and a diagnostic naming a variable the author's spec never wrote — the same defect #732's second test exists to prevent for constructors.

## The systemic finding

**No copy strategy had a generic-*spec* test except `@Wither` — and `@Wither` is structurally blind to this fault**, because it rebuilds through a method and reads no member of the source type. That is why #732 shipped in #727 and was found only by a panel on a later PR.

Generic spec interfaces became first-class in #730, which turns a set of latent faults into reachable ones. The suite has no axis that would catch any of them.

**Recommendation:** make "generic spec" a first-class axis — one case per strategy and hint (`@ViaBuilder`, `@ViaConstructor`, `@ViaCopyAndSet`, `@Wither`, `@InstanceOf`, `@MatchWhen`, `@TraverseWith`, `@ThroughField`) against a generic source type, and one per optic processor against a generic record. It earns its keep immediately: instance 3 above is exactly such a case, and it fails today.

## Suggested shape for the fix

Where the member's owner is instantiated, read it with `types.asMemberOf(instantiatedOwner, member)` rather than `member.asType()`. `MappingProcessor.componentType:682` and `SpecInterfaceAnalyser.constructorParameterType` (added by #735) are the two worked examples in tree. When a third *processor* needs it, the three should fold into one shared helper rather than a fourth copy.

Where the parameters are being *declared* rather than compared — instances 1 and 2 — the answer is #730's: declare the parameters the signature actually names, taken from the right element.

MagnusSmith:
Fixes the three remaining sites in #736 — one commit each, since the right answer differs per site.

The family: **a member or type read *as declared*, then used against something *instantiated*.** The two vocabularies agree only by name coincidence, which is how each of these survived.

## 1. `@ThroughField` — rejected where the container was there to find

`SpecInterfaceAnalyser.findFieldType` read the record component, accessor return or public field straight off the element:

```java
public class Holder<T> { public T items() { ... } }

interface HolderOpticsSpec extends OpticsSpec<Holder<List<String>>> {
    Traversal<Holder<List<String>>, String> eachItem();
}
```

```
Cannot auto-detect traversal for field 'items' of type 'T'.
```

Two faults in one. `T` is `List<String>` under that instantiation, so a container *was* there; and the message named a variable the spec never wrote — the same defect #732's rejection had.

Reading the member through `asMemberOf` answers both. The record branch goes via the component's **accessor**, since that is the member javac substitutes. The one diagnostic that named the source type used its element's qualified name, which cannot say which instantiation was searched; it names the type now.

## 2. `@ImportOptics` — a generated traversal named a parameter it never declared

`ExternalLensGenerator.createTraversalMethod` named the record raw and declared no type parameters, while the focus type it computed read the component's own arguments:

```java
public record Bag<T>(String name, List<T> items) {}

@ImportOptics({com.external.Bag.class})
package com.myapp.optics;
```

→ `cannot find symbol: class T, location: class com.myapp.optics.BagLenses`

The two lens methods **in the same class** declared the parameters correctly. This was a sibling-sweep miss: #726 fixed exactly this in `TraversalProcessor` and left the `@ImportOptics` mirror of the same generator untouched. The fix applies the same three things #726 did — the instantiated record name in the signature, the record's own parameters on the method, and the effect variable read from `ProcessorUtils.effectVariableName` so a record claiming `F` for a parameter of its own does not collide with it.

Now generates `public static <T> Traversal<Bag<T>, T> itemsTraversal()`.

## 3. `@GenerateFocus` navigators — resolved by *not* offering one

This one I deliberately did **not** fix the way the other two were, and the reasoning is the substance of the commit.

A record holding an `Inner<String>` generated nine `cannot find symbol: class T` errors inside its own Focus class. I threaded the instantiation through the navigator; that fixed the class's own type name and left one error — the components its navigation methods read, which would mean substituting inside `WideningAnalysis`. That is a feature, not a fix, and no caller can be waiting for it, because the shape has never compiled.

So the component keeps its **plain path method**, which carries the instantiation and composes the same way. That is the branch a component with no navigable target already takes, so nothing new is generated — the navigator is simply not offered.

`@GenerateFocus` on a generic *record* is untouched; it was only the navigator for a generic *target* that had no way to name it. Supporting one is worth its own issue.

## Fourth commit: what the panel found

The panel caught a regression, a half-fix, and a wrong justification — all three in my own work.

**A raw source type erased every member.** javac's `memberType` returns `erasure(sym.type)` whenever the site carries no arguments, whatever the member declares — so reading `List<String> items()` under a raw `Holder` came back as `List` and matched no container. That is a spec which compiled before this branch, and the message it produced contradicted itself: *"of type 'List'. Supported types: List, …"*. There is nothing to substitute when the site has no arguments, so the member is read as declared — the guard `MappingProcessor.componentType` already carries.

**The navigator skip was half a fix, and the missing half was older than this PR.** `widensContainers` still asked whether the element is *navigable*, while `navigatorTarget` now answers a narrower question, so a record navigating **into** a container of a generic element declared a traversal and returned a focus. It asks `navigatorTarget` now, restoring what that method's javadoc claims — *"every site that asks reads the answer from here, so they cannot disagree"* — which also closed two triggers that predate this work entirely: a container whose element record turns navigators off, and one whose field the record excludes. Both are now pinned.

**The reason I gave for not resolving the navigator was wrong.** A reviewer neutered the guard and read the nine errors: every one is written in the *target's own* variables, which `WideningAnalysis` already returns. Substitution inside it was never the obstacle — the parameters could be declared on the navigator class and passed at its two construction sites. What is genuinely unsettled is a component written `Inner<?>` or raw, whose arguments cannot be written as class type arguments. Filed as #739 with the real reason.

**Declining silently was the wrong half of the precedent.** The same file reports a navigator declined for a name collision with a note naming the chain to write instead. A generic target now draws the same note — and `@GenerateFocus`'s own javadoc and the book's decision tree, both of which promised a navigator for any annotated field type, now say what is actually generated.

Also: the stale javadoc block orphaned above `memberTypeOf` — the same slip #729's review caught in `ProcessorUtils`, twice now.

## Split out rather than folded in

- **#739** — generate navigators for a generic target, with the corrected reasoning and the wildcard/raw decision that is actually unsettled.
- **#740** — a sixth and seventh instance of this family, both across a *supertype* boundary (`BeanPropertyAnalyser` reading an inherited getter, `MappingProcessor.checkMixins` missing a transitive generic mix-in). Both confirmed by compiling. That issue also carries the recommendation to consolidate the three near-copies of this call into `ProcessorUtils`, with the two subtleties each learned separately.

## Verification

Each fix has a regression test in the file that owns that behaviour — `ThroughFieldAutoDetectTest`, `ImportOpticsProcessorTest`, `NavigatorCoverageTest`.

Two of my first attempts at the navigator tests were wrong in a way worth recording: `List<Inner>` does not reach the SPI navigable path at all — a hardcoded collection widens through `.each()` instead — so the test passed for the wrong reason. `Map<String, Inner<String>>` is the shape that actually exercises it, which is what the committed test uses. The panel found a third of the same kind: a test asserting only the navigator's *absence* passes vacuously if navigators are ever disabled, so it now pins what is emitted instead.

`clean build` across all modules: BUILD SUCCESSFUL, EXIT=0. `NavigatorClassGenerator` and `SpecInterfaceAnalyser` both stay at 100% line, branch and instruction.

Closes #736 once #737 (the test axis) also lands.

## Summary by CodeRabbit

* **New Features**
  * Improved support for generic records and fields when generating traversals and focus paths.
  * Generic targets now retain standard `.via(...)` paths when navigator wrappers cannot be generated.
  * Added clearer diagnostics explaining navigator limitations and recommended chaining.

* **Bug Fixes**
  * Corrected type resolution for instantiated generic fields, accessors, containers, and external traversals.
  * Improved generated traversal signatures for generic record types.

* **Documentation**
  * Clarified navigator generation rules and generic-type behavior.

### DIFF
--- a/hkj-annotations/src/main/java/org/higherkindedj/optics/annotations/GenerateFocus.java
+++ b/hkj-annotations/src/main/java/org/higherkindedj/optics/annotations/GenerateFocus.java
@@ -75,7 +75,8 @@ public @interface GenerateFocus {
    *
    * <p>This enables patterns like {@code PersonFocus.address().city()} without explicit {@code
    * .via()} calls. Navigator classes are generated for fields whose types are also annotated with
-   * {@code @GenerateFocus}.
+   * {@code @GenerateFocus} and declare no type parameters of their own; a field whose type is
+   * generic keeps its plain {@code FocusPath} method, composed with {@code .via()}.
    *
    * <p>The navigator classes:
    *
--- a/hkj-book/src/optics/focus_navigation.md
+++ b/hkj-book/src/optics/focus_navigation.md
@@ -215,9 +215,10 @@ Not every field does, and knowing which is the difference between a chain that c
 ```mermaid
 flowchart TD
     F{"The field's type is..."}
-    F -->|"a record annotated<br/>@GenerateFocus"| N(["Navigator<br/>chain with a method call"])
+    F -->|"a non-generic record<br/>annotated @GenerateFocus"| N(["Navigator<br/>chain with a method call"])
     F -->|"Optional, List, Set,<br/>Collection"| W(["Widened path<br/>chain with .via()"])
-    F -->|"an SPI container whose<br/>element type is annotated"| N
+    F -->|"an SPI container whose<br/>element type is annotated<br/>and non-generic"| N
+    F -->|"a generic record annotated<br/>@GenerateFocus"| P
     F -->|"anything else"| P(["Plain path<br/>chain with .via()"])
 
     classDef decision fill:#e5c890,stroke:#df8e1d,color:#232634
@@ -227,6 +228,8 @@ flowchart TD
 ```
 
 The middle branch is the one that surprises people. `Optional`, `List`, `Set` and `Collection` are widened by the processor before navigators are considered, so a `List<Department> departments` field gives you a `TraversalPath<Company, Department>` and never a `DepartmentsNavigator`. Container types that arrive through the SPI (a `Map`, an Eclipse Collections `ImmutableList`, an `Either`) *are* eligible, and get a navigator when their element type is itself annotated.
+
+A target that declares type parameters of its own does not. A navigator is an inner class parameterised by the source type alone, so it has no way to name them — `Inner<String> inner` and `Map<String, Inner<String>> inners` both keep the plain path, chained with `.via()`. The processor says so as a note against the field, naming the chain to write instead.
 
 <!-- verify -->
 ```java
--- a/hkj-processor/src/main/java/org/higherkindedj/optics/processing/NavigatorClassGenerator.java
+++ b/hkj-processor/src/main/java/org/higherkindedj/optics/processing/NavigatorClassGenerator.java
@@ -101,11 +101,19 @@ public class NavigatorClassGenerator {
   /**
    * Whether the Focus method for this component steps into a {@code ZERO_OR_MORE} container.
    *
-   * <p>A container holding a navigable element always does — reaching that element is what its
-   * navigator exists for — and otherwise the declaring record's {@code widenCollections} decides.
+   * <p>A container whose element reaches a navigator always does — reaching that element is what
+   * the navigator exists for — and otherwise the declaring record's {@code widenCollections}
+   * decides.
+   *
+   * <p>The question is the one {@link #navigatorTarget} answers, not merely whether the element is
+   * navigable. A container the widening steps into but no navigator is offered for would leave the
+   * navigation method declaring a traversal and returning a focus, and every reason a navigator is
+   * declined — the element is generic, the record turned navigators off, the field is filtered out
+   * — reaches that same disagreement.
    */
   private boolean widensContainers(TypeElement record, RecordComponentElement component) {
-    return spiNavigable(component.asType()) != null || focusSettings(record).widenCollections();
+    return (spiNavigable(component.asType()) != null && navigatorTarget(record, component) != null)
+        || focusSettings(record).widenCollections();
   }
 
   /**
@@ -141,6 +149,7 @@ public class NavigatorClassGenerator {
     for (RecordComponentElement component : recordElement.getRecordComponents()) {
       TypeElement target = navigatorTarget(recordElement, component);
       if (target == null) {
+        reportGenericTargetSkipped(recordElement, component);
         continue;
       }
       focusClassBuilder.addType(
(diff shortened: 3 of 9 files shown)


MagnusSmith:
@Coderabbit full review

MagnusSmith:

PR #741, opened 2026-08-26
@GenerateMapping: read an inherited member under the type that inherits it (#740)

### LINKED ISSUE #740: [BUG] Declaration-vs-instantiation across a supertype: inherited bean and mix-in members
Two more sites of the [#736](https://github.com/higher-kinded-j/higher-kinded-j/issues/736) family — a member read **as declared** and used against an **instantiated** type — this time where the member is *inherited*, so the declaring element and the instantiating one are different types. Both confirmed by compiling.

Every previously-found instance was within one type. These are the cross-supertype cell, which is where `asMemberOf` earns its keep most.

## 1. `BeanPropertyAnalyser` — a bean inheriting from a generic base

`analyse()` gathers members via `Elements#getAllMembers` — deliberately, the class javadoc says *"a bean inherits properties from its superclasses (as JAXB-generated beans do)"* — then reads `getReturnType()` / `getParameters().getFirst().asType()` off the **declaring** element.

```java
public class BaseDto<T> { public T getId() {...} public void setId(T id) {...} }
public class UserDto extends BaseDto<String> { ... }

public record User(String id, String name) {}

@GenerateMapping
public interface UserMapping extends MappingSpec<User, UserDto> {}
```

```
@GenerateMapping: target field 'UserDto.id' has no usable source. The types differ
(T vs java.lang.String) and no matching leaf method was found.
Add 'default ValidatedPrism<T, java.lang.String> id()' to the spec
```

Both harms this family always has: a mapping that was there to find is rejected, and the remedy names `T` — a variable in scope nowhere the author can write it, so the suggested fix is literally untypeable.

`MappingProcessor.checkNotGeneric` only inspects `wireBean.getTypeParameters()`, so a bean that is itself non-generic passes the gate. JAXB/XJC-style DTO hierarchies with a generic base are the ordinary case for bean wires.

Fix: `asMemberOf(bean.asType(), member)`. `checkNotGeneric` runs first, so the bean type is always closed and the call is total.

## 2. `MappingProcessor.checkMixins` — a non-generic mix-in extending a generic one

`checkMixins` rejects a generic mix-in with a well-judged message, but walks only `spec.getInterfaces()` — the direct parents — while `specMembers` uses `getAllMembers`, which walks the whole ancestry. One level of indirection slips the gate:

```java
interface BaseEmails<T> { default ValidatedPrism<String, T> email() {...} }
interface Emails extends BaseEmails<EmailAddress> {}

@GenerateMapping
interface UserMapping extends MappingSpec<User, UserDto>, Emails {}
```

```
A default method 'email()' exists but returns
'ValidatedPrism<java.lang.String,T>' ...
Add 'default ValidatedPrism<java.lang.String, EmailAddress> email()' to the spec
```

The free `T` leaks into the diagnostic, and the author is told to write what they already wrote one interface along.

Worth noting the gate itself is arguably the non-conforming part. Its message says a generic mix-in *"would need substitution under its instantiation, which is not supported yet"* — but `asMemberOf(spec.asType(), method)` resolves both this transitive case and the direct one the gate currently rejects. That is the "resolve rather than reject where resolution is possible" arm the recent hardening run established.

## Suggested shape

Both want the same call, which now exists in three near-copies:

- `MappingProcessor.componentType`
- `SpecInterfaceAnalyser.constructorParameterType` (#735)
- `SpecInterfaceAnalyser.memberTypeOf` (#738)

Two of them are in one class. This is the moment to lift one documented helper into `ProcessorUtils`, taking `Types` as a first parameter so the class stays stateless, and to fold in the two subtleties each copy learned separately:

- skip `asMemberOf` when the site carries no type arguments — under a **raw** site javac erases every member, even one whose declared type never mentions the class's variable (this regressed `@ThroughField` in #738 before it was caught);
- ~~go through a record component's **accessor** rather than the component itself, which is what javac substitutes.~~ **Struck.** Measured on JDK 25: `asMemberOf` on a `RecordComponentElement` works and agrees with the accessor route — `T` → `String`, `List<T>` → `List<String>`, and both erase alike under a raw owner. Worth doing for portability; nothing is wrong today.

`ProcessorUtils` is not in the 100%-pinned class list in `hkj-processor/build.gradle.kts`; if the helper lands there, it should be added.

## Test axis

[#737](https://github.com/higher-kinded-j/higher-kinded-j/pull/737) added a generic-spec axis for `@ImportOptics` strategies and hints. Neither of these is covered by it, and neither is the `@ImportOptics({Class...})` auto-detection path. The rows worth adding:

- `@GenerateMapping` × (record wire / bean wire / sealed wire) × (own generics / inherited from a generic base).
- A `GenericImportedTypeAxisTest` for the `@ImportOptics({Class...})` path: (record / wither class / sealed interface / enum) × (non-generic / generic). It would pay for itself immediately — the sealed × generic cell emits raw prisms (`ExternalPrismGenerator`, matching `PrismProcessor`) — *confirmed, and fixed as #742* — ~~and the wither × generic cell emits no traversal at all~~. **Struck:** *no* wither class gets a `Traversals` class through `@ImportOptics({Class...})`, generic or not, so it is not a generics-specific defect. If the auto-detection path should emit them at all, that is its own question and not this axis's.

MagnusSmith:
Fixes #740. Independent of #738 — different files, no conflict, either order.

The declaration-vs-instantiation family, in the cell every earlier sweep missed: **the member and the type it is read on are different types.** Every previous instance was within one type, which is why five sweeps went past these.

## 1. A bean inheriting a property from a generic base

`BeanPropertyAnalyser` gathers properties with `getAllMembers` — deliberately, since a JAXB-generated wire inherits them — then read each getter's return and each setter's parameter off the element that *declared* them.

```java
public class BaseDto<T> { public T getId() {...} public void setId(T id) {...} }
public class UserDto extends BaseDto<String> { ... }

public record User(String id, String name) {}

@GenerateMapping
public interface UserMapping extends MappingSpec<User, UserDto> {}
```

```
@GenerateMapping: target field 'UserDto.id' has no usable source. The types differ
(T vs java.lang.String) and no matching leaf method was found.
Add 'default ValidatedPrism<T, java.lang.String> id()' to the spec
```

Both harms this family always has: a mapping that was there to find is rejected, and the remedy names `T` — a variable in scope nowhere the author can write it, so it **cannot be followed**. `checkNotGeneric` only inspects the bean's own type parameters, so a bean that is itself non-generic sails through.

Read under `UserDto`, the member is `String` and the field maps.

## 2. A generic mix-in reached through a non-generic one

`checkMixins` walked only `spec.getInterfaces()` — the direct parents — while `specMembers` uses `getAllMembers`, which walks the whole ancestry. One level of indirection slipped the gate:

```java
interface BaseEmails<T> { default ValidatedPrism<String, T> email() {...} }
interface Emails extends BaseEmails<EmailAddress> {}

@GenerateMapping
interface UserMapping extends MappingSpec<User, UserDto>, Emails {}
```

The free `T` reached the diagnostics the same way, and the author was told to write what they had already written one interface along.

It walks the ancestry now, and names the ancestor that is **actually** generic rather than the one the spec lists. Supporting a generic mix-in is a separate question — the gate's own wording says so — and this only makes the refusal honest.

## The reading moves to `ProcessorUtils`

This is the fourth site to need "read this member under the owner's instantiation", and two of the others sit in one class. It carries the two subtleties the earlier sites each learned separately:

- a **record component** is read through its **accessor**, which is the member javac substitutes;
- a **raw** owner is read as declared, because javac's `memberType` returns `erasure(sym.type)` for a raw site — erasing even a member that names no variable at all. That one regressed `@ThroughField` in #738 before a reviewer caught it.

The distinction that matters here, and that neither earlier copy had to make: **an owner that merely declares no parameters is not raw.** `UserDto` supplies no type arguments of its own, yet its inherited members are exactly what the walk is for. Raw is "declares parameters, supplied none".

```java
boolean raw = !((TypeElement) owner.asElement()).getTypeParameters().isEmpty()
    && owner.getTypeArguments().isEmpty();
```

`SpecInterfaceAnalyser`'s two copies fold into this once #738 lands; they are left alone here to keep the branches independent.

## Verification

Both fixes have a regression test in the file that owns that behaviour — `MappingProcessorBeanTest` and `MappingProcessorUpdateTest`. `clean build` across all modules: BUILD SUCCESSFUL, EXIT=0, with the 100% gates on `BeanPropertyAnalyser*` and `MappingProcessor*` intact.

Not addressed here, and still open on #740: the test-axis recommendation (a `@GenerateMapping` × wire-shape × inherited-generics axis, and a `GenericImportedTypeAxisTest` for the `@ImportOptics({Class...})` path). Those are worth their own change — the axis added in #737 paid for itself on its first run, and the two cells that issue names are still unexercised.

### DIFF
--- a/hkj-processor/src/main/java/org/higherkindedj/optics/processing/BeanPropertyAnalyser.java
+++ b/hkj-processor/src/main/java/org/higherkindedj/optics/processing/BeanPropertyAnalyser.java
@@ -18,6 +18,7 @@ import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
 import javax.lang.model.util.ElementFilter;
 import org.higherkindedj.optics.processing.util.Diagnostics;
+import org.higherkindedj.optics.processing.util.ProcessorUtils;
 
 /**
  * Discovers the JavaBeans property model of a bean-shaped wire type for {@code @GenerateMapping} .
@@ -53,6 +54,9 @@ final class BeanPropertyAnalyser {
    * neither a mutable JavaBean nor a builder-based bean with a readable property.
    */
   WireShape.BeanShape analyse(TypeElement spec, TypeElement bean, String tag) {
+    // A bean inherits properties from its superclasses, so a member read off its declaring element
+    // speaks that element's variables: 'T getId()' on BaseDto<T> is String on UserDto.
+    DeclaredType beanType = (DeclaredType) bean.asType();
     Map<String, ExecutableElement> getters = collectGetters(bean);
 
     if (hasUsableNoArgsConstructor(spec, bean)) {
@@ -60,11 +64,12 @@ final class BeanPropertyAnalyser {
       List<WireShape.BeanProperty> properties = new ArrayList<>();
       for (Map.Entry<String, ExecutableElement> entry : getters.entrySet()) {
         String name = entry.getKey();
-        TypeMirror getterType = entry.getValue().getReturnType();
+        TypeMirror getterType =
+            ProcessorUtils.returnTypeIn(env.getTypeUtils(), beanType, entry.getValue());
         String getter = entry.getValue().getSimpleName().toString();
         ExecutableElement setter = setters.get(name);
         if (setter != null) {
-          if (typesDiffer(spec, bean, tag, name, getterType, paramType(setter))) {
+          if (typesDiffer(spec, bean, tag, name, getterType, paramType(beanType, setter))) {
             return null;
           }
           properties.add(
@@ -87,6 +92,7 @@ final class BeanPropertyAnalyser {
 
     BuilderModel builder = findBuilderModel(bean);
     if (builder != null) {
+      DeclaredType builderType = (DeclaredType) builder.builderType().asType();
       Map<String, ExecutableElement> builderSetters = collectBuilderSetters(builder.builderType());
       List<WireShape.BeanProperty> properties = new ArrayList<>();
       for (Map.Entry<String, ExecutableElement> entry : getters.entrySet()) {
@@ -95,8 +101,9 @@ final class BeanPropertyAnalyser {
         if (builderSetter == null) {
           continue;
         }
-        TypeMirror getterType = entry.getValue().getReturnType();
-        if (typesDiffer(spec, bean, tag, name, getterType, paramType(builderSetter))) {
+        TypeMirror getterType =
+            ProcessorUtils.returnTypeIn(env.getTypeUtils(), beanType, entry.getValue());
+        if (typesDiffer(spec, bean, tag, name, getterType, paramType(builderType, builderSetter))) {
           return null;
         }
         properties.add(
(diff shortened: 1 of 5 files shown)


MagnusSmith:

Issue #742, opened 2026-08-26
[BUG] @ImportOptics generates raw prisms for a generic sealed hierarchy


MagnusSmith:
A generic sealed hierarchy imported with `@ImportOptics({Shape.class})` generates prisms naming both sides **raw**, so they do not compose with any optic that carries the hierarchy's type argument.

```java
public sealed interface GenShape<T> permits GenCircle {}
public record GenCircle<T>(T tag) implements GenShape<T> {}

@ImportOptics({com.external.GenShape.class})
package com.myapp.optics;
```

generates

```java
public static Prism<GenShape, GenCircle> genCircle() { ... }
```

It compiles — nothing fails a build — but `GenShapePrisms.genCircle()` cannot be composed with a `Lens<GenShape<String>, …>`, and the raw types leak into whatever the caller writes next.

## Where

`ExternalPrismGenerator.createPrismMethodForSubtype`, which uses `ClassName.get(sumTypeElement)` / `ClassName.get(subtype)` — the raw names — and never calls `addTypeVariable`.

`PrismProcessor` (the `@GeneratePrisms` path) does the same thing, so the two prism generators are at least consistent with each other. Both want the same change.

## The shape of the fix

This is the `@GeneratePrisms`/`@ImportOptics` cell of what [#736](https://github.com/higher-kinded-j/higher-kinded-j/issues/736) fixed for records: declare the hierarchy's type parameters on the generated method and name both sides instantiated.

```java
public static <T> Prism<GenShape<T>, GenCircle<T>> genCircle() { ... }
```

The lens generators one file over already do exactly this — `ExternalLensGenerator.createRecordLensMethod` and `createWitherLensMethod` both loop `recordElement.getTypeParameters()` and `addTypeVariable`. `createTraversalMethod` was the odd one out until [#738](https://github.com/higher-kinded-j/higher-kinded-j/pull/738).

Two things a fix has to settle, both in the family [#718](https://github.com/higher-kinded-j/higher-kinded-j/issues/718) and [#721](https://github.com/higher-kinded-j/higher-kinded-j/issues/721) established:

- a permitted subtype that **re-parameterises**, `record Pair<A, B>(A a, B b) implements GenShape<A>` — the prism's own parameters are the subtype's, not the hierarchy's, and only some of them are bound by the source;
- a permitted subtype that **pins** its argument, `record Tagged() implements GenShape<String>` — the method takes no parameter at all, like the concrete-instantiation case in #730.

[...]

Issue #739, opened 2026-08-26
Generate navigators for a generic navigable target


MagnusSmith:
`@GenerateFocus(generateNavigators = true)` declines to generate a navigator when the navigable target declares type parameters of its own. The component keeps its plain `FocusPath` method, which composes the same way with `.via()`, and the processor says so as a note against the field.

```java
@GenerateFocus(generateNavigators = true) public record Inner<T>(T value, String label) {}
@GenerateFocus(generateNavigators = true) public record Outer(Inner<String> inner, String tag) {}
```

gives `OuterFocus.inner()` returning `FocusPath<Outer, Inner<String>>` rather than an `InnerNavigator`, so the chain is `OuterFocus.inner().via(InnerFocus.value())` rather than `OuterFocus.inner().value()`.

## Why it is declined rather than generated

The navigator is an inner class parameterised by the source type alone:

```java
public static final class InnerNavigator<S> { ... }
```

and its navigation methods read the target's components. Both name the target's own type variables, which are in scope on neither.

Worth recording precisely, because the first explanation given was wrong: this does **not** need substitution inside `WideningAnalysis`. Neutering the guard produces nine errors, and every one is written in the target's own variables — exactly what `WideningAnalysis.analyse(component)` already returns for a component of `Inner`. The parameters could be declared on the navigator class, `InnerNavigator<S, T>`, and the component's type arguments passed at the two construction sites, `createNavigatorMethod` and `addNavigatorComposition`, straight off `component.asType()`.

## What is actually unsettled

A component whose arguments cannot be written as class type arguments:

```java
record Outer(Inner<?> inner, String tag) {}     // a wildcard
record Outer(Inner inner, String tag) {}        // raw
```

`InnerNavigator<S, ?>` is not something that can be written, and a raw target supplies nothing to pass. So a resolution needs its own resolve-or-reject decision in the same family as [#718](https://github.com/higher-kinded-j/higher-kinded-j/issues/718) and [#721](https://github.com/higher-kinded-j/higher-kinded-j/issues/721): resolve where the component pins its arguments, reject at the declaration where it does not.

```mermaid
flowchart TD
    C["a component whose type is a<br/>generic @GenerateFocus record"] --> Q{"are its arguments<br/>writable as class<br/>type arguments?"}
    Q -->|"Inner&lt;String&gt;, Inner&lt;T&gt;"| R["declare them on the navigator<br/>InnerNavigator&lt;S, String&gt;"]
[...]

PR #743, opened 2026-08-26
Two more generic axes: imported types, and mapping wire shapes (#740)

MagnusSmith:
Adds the two test axes recommended in #740, and files the gap they found.

> **Base note.** Contains merges of both open branches it depends on — #738 (the generic-record traversal) and #741 (the inherited bean property). Merge those first; this then reduces to the two new test files. Without either, three of the nine cells fail — which is the point.

## Why two more axes

#737 added a generic-spec axis for `@ImportOptics` **spec interfaces**, and it found a bug on its first run. The other two paths into the processor had none, and both had a live defect in the generic half at the time of writing.

### `GenericImportedTypeAxisTest` — `@ImportOptics({Class...})`

Every shape the auto-detection path accepts, in a non-generic and a generic form:

| shape | non-generic | generic |
|---|---|---|
| record | ✔ | ✔ — the traversal that named the record raw (#738) |
| wither class | ✔ | ✔ |
| sealed interface | ✔ | ⚠ raw prisms — **#742** |
| enum | ✔ | *n/a — an enum cannot declare parameters* |

The generic half is where a generator naming the imported type **raw** and one reading a component's own type arguments disagree. Every non-generic form agrees whichever way it is read — which is exactly how a generic record's traversal named a parameter it never declared, in a class whose lens methods declared theirs correctly.

### `GenericMappingWireAxisTest` — `@GenerateMapping`

Every wire shape, with its type parameters supplied three ways:

| wire | none | declared by the spec | inherited from a generic base |
|---|---|---|---|
| record | ✔ | ✔ | *n/a — a record cannot extend* |
| bean | ✔ | — | ✔ — the property read as declared (#741) |
| mix-in ancestry | — | — | ✔ rejected, naming the ancestor that is generic |

**The inherited column is the one that hid a defect.** A wire declaring no parameters of its own still reaches members that do — `class UserDto extends BaseDto<String>` answers `T getId()` — and a gate asking only whether the wire itself is generic let those through unread. Every other column agrees whether members are read as declared or under the wire.

## Confirmed as real guards

Reverting the three main-source files to their `origin/main` state fails exactly three of the nine cells, one per bug:

```
a bean wire inheriting a property from a generic base            FAILED
a record, and a generic record whose parameters the methods declare  FAILED
a mix-in carrying a generic ancestor's members is rejected by name   FAILED
9 tests completed, 3 failed
```

## The gap it found: #742

`@ImportOptics` on a generic sealed hierarchy generates `Prism<GenShape, GenCircle>` — **raw on both sides**. It compiles, so nothing fails a build, but the prism will not compose with any optic carrying the hierarchy's argument. `PrismProcessor` does the same, so the two prism generators are consistent with each other and both want the same change.

The cell asserts what **is** emitted, with a comment naming the issue. Pinning current behaviour rather than omitting the row means closing #742 shows up here as a test to update, instead of as a cell nobody was looking at.

## Verification

`clean build` across all modules: BUILD SUCCESSFUL, EXIT=0.

Still open on #740 after this: nothing. #742 and #739 carry what is left.

### DIFF
--- a/hkj-annotations/src/main/java/org/higherkindedj/optics/annotations/GenerateFocus.java
+++ b/hkj-annotations/src/main/java/org/higherkindedj/optics/annotations/GenerateFocus.java
@@ -75,7 +75,8 @@ public @interface GenerateFocus {
    *
    * <p>This enables patterns like {@code PersonFocus.address().city()} without explicit {@code
    * .via()} calls. Navigator classes are generated for fields whose types are also annotated with
-   * {@code @GenerateFocus}.
+   * {@code @GenerateFocus} and declare no type parameters of their own; a field whose type is
+   * generic keeps its plain {@code FocusPath} method, composed with {@code .via()}.
    *
    * <p>The navigator classes:
    *
--- a/hkj-book/src/optics/focus_navigation.md
+++ b/hkj-book/src/optics/focus_navigation.md
@@ -215,9 +215,10 @@ Not every field does, and knowing which is the difference between a chain that c
 ```mermaid
 flowchart TD
     F{"The field's type is..."}
-    F -->|"a record annotated<br/>@GenerateFocus"| N(["Navigator<br/>chain with a method call"])
+    F -->|"a non-generic record<br/>annotated @GenerateFocus"| N(["Navigator<br/>chain with a method call"])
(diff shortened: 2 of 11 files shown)


MagnusSmith:

PR #744, opened 2026-08-26
@GeneratePrisms: name both sides of a prism for a generic hierarchy (#742)

### LINKED ISSUE #742: [BUG] @ImportOptics generates raw prisms for a generic sealed hierarchy

MagnusSmith:
Fixes #742.

> **Base note.** Stacked on #743, which is itself stacked on #738 and #741 — the axis cell this changes was added there. Merge in that order; this then reduces to the two generators plus the axis update.

A prism for a permitted subtype named both the sum type and the subtype **raw**, and declared no type parameters:

```java
public sealed interface GenShape<T> permits GenCircle {}
public record GenCircle<T>(T tag) implements GenShape<T> {}
```
```java
public static Prism<GenShape, GenCircle> genCircle() { ... }   // before
```

It compiled — which is why nothing caught it for so long — but a raw prism composes with nothing carrying the hierarchy's argument, and the rawness spreads into whatever the caller writes next.

## The reading that answers all three shapes

The prism is written in the **subtype's** vocabulary: its own parameters declared on the method, and the sum type as the subtype's own `implements` clause instantiates it.

| the subtype declares | generated |
|---|---|
| `Circle<T> implements Shape<T>` | `static <T> Prism<Shape<T>, Circle<T>> circle()` |
| `Tagged implements Shape<String>` | `static Prism<Shape<String>, Tagged> tagged()` |
| `Pair<A, B> implements Shape<A>` | `static <A, B> Prism<Shape<A>, Pair<A, B>> pair()` |
| `Leaf implements Plain` | unchanged |

**`Pair` is the one worth stating.** `B` is the subtype's own and appears in the focus type, so it is declared even though the sum type does not bind it. That follows what #730 settled: a signature declares the parameters it **names**, and `B` is one the author wrote — unlike #730's `B`, which was invented for them from a declaration they never referenced.

## The unchecked cast

`instanceof` tests an erasure, so a parameterised subtype narrows through the raw name. Here the sealed hierarchy is what makes that sound — a `Shape<T>` that is a `Circle` can only be a `Circle<T>` — but javac cannot see it, so the warning is answered where it arises rather than left for a consuming build under `-Werror`. Verified by compiling the generated output with `-Xlint:unchecked,rawtypes -Werror`.

This is the *sound* sibling of #733, where an `@InstanceOf` target carries no such guarantee and the same cast genuinely can throw.

## Both generators

`PrismProcessor` (`@GeneratePrisms`) and `ExternalPrismGenerator` (`@ImportOptics`) had the same code. Both change, so the two paths keep agreeing — the thing #726 got wrong by fixing one of a pair and leaving the mirror.

## How it was found, and what that says

The axis added in #743 pinned the raw output **deliberately**, with a comment naming this issue. Running the full suite after the fix produced exactly one failure out of 2088 — that cell, and nothing else. No other test had pinned the raw prisms.

That is what pinning a known-wrong behaviour is for: closing the gap shows up as a test to update rather than as a cell nobody was looking at. The cell now pins what is generated, and two more join it for the pinned and extra-parameter subtypes.

## Verification

`clean build` across all modules: BUILD SUCCESSFUL, EXIT=0, with the 100% gates intact.

### DIFF
--- a/hkj-processor/src/main/java/org/higherkindedj/optics/processing/PrismProcessor.java
+++ b/hkj-processor/src/main/java/org/higherkindedj/optics/processing/PrismProcessor.java
@@ -3,11 +3,14 @@
 package org.higherkindedj.optics.processing;
 
 import com.google.auto.service.AutoService;
+import com.palantir.javapoet.AnnotationSpec;
 import com.palantir.javapoet.ClassName;
 import com.palantir.javapoet.JavaFile;
 import com.palantir.javapoet.MethodSpec;
 import com.palantir.javapoet.ParameterizedTypeName;
+import com.palantir.javapoet.TypeName;
 import com.palantir.javapoet.TypeSpec;
+import com.palantir.javapoet.TypeVariableName;
 import java.io.IOException;
 import java.util.List;
 import java.util.Optional;
@@ -22,7 +25,9 @@ import javax.lang.model.element.Element;
 import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.Modifier;
 import javax.lang.model.element.TypeElement;
+import javax.lang.model.element.TypeParameterElement;
 import javax.lang.model.element.VariableElement;
+import javax.lang.model.type.DeclaredType;
 import javax.tools.Diagnostic;
 import org.higherkindedj.optics.Prism;
 import org.higherkindedj.optics.annotations.GeneratePrisms;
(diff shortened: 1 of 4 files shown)


MagnusSmith:

PR #745, opened 2026-08-26
Declaration-vs-instantiation: the remaining sites, plus the axes that find them (#736, #740, #742)

### LINKED ISSUE #736: [BUG] Declaration-vs-instantiation: three more sites read a member as declared and use it against an instantiated type

MagnusSmith:
Closes #736, #740, #742. Replaces the #738 → #741 → #743 → #744 stack with one linear branch off `main` — same tree, seven commits, no merges.

## One mistake, seven places

> **A member or type is read *as declared*, then compared with — or emitted alongside — something derived from an *instantiated* type.**

The two vocabularies agree by name coincidence, which is why every instance survived so long and why five separate sweeps each found a different subset. #728, #730 and #732 fixed the first three; this branch fixes the rest and adds the axes that make the generic half visible.

```mermaid
flowchart TD
    D["the declaration<br/>class Node&lt;X&gt;"] --> P["read as declared<br/>Base&lt;X&gt;, T, raw Bag"]
    I["the instantiation<br/>Node&lt;U&gt;, Box&lt;List&lt;String&gt;&gt;"] --> Q["derived by substitution<br/>Base&lt;U&gt;, List&lt;String&gt;"]
    P --> C{"compared, or<br/>emitted together"}
    Q --> C
    C -->|"names agree"| OK["works, by coincidence"]
    C -->|"names differ"| X["cannot find symbol,<br/>or a wrong rejection"]

    class D,I wire
    class P,Q domain
    class C decision
    class OK domain
    class X error
```

## The commits

| # | what | why it mattered |
|---|---|---|
| 1 | **`@ThroughField`** reads the member under the instantiation | `T items()` on `Holder<List<String>>` was rejected as *"of type 'T'"* — naming a variable the spec never wrote, and refusing a container that was there to find |
| 2 | **`@ImportOptics`** declares the record's parameters on a generated traversal | a generic record's traversal named `T` undeclared, while the lens methods beside it declared theirs correctly — the mirror #726 fixed and left |
| 3 | **`@GenerateFocus`** leaves a generic navigable target its plain path method | nine `cannot find symbol` inside the generated Focus class |
| 4 | **Panel fixes** to 1–3 | see below |
| 5 | **`@GenerateMapping`** reads an inherited member under the type that inherits it | a JAXB-style `UserDto extends BaseDto<String>` was told to add `ValidatedPrism<T, String>` — a remedy that cannot be written |
| 6 | **Two more axes** — imported types, mapping wire shapes | the two paths into the processor that had none |
| 7 | **`@GeneratePrisms`** names both sides for a generic hierarchy | `Prism<GenShape, GenCircle>`, raw — compiles, composes with nothing |

## What the panel caught in commits 1–3

Worth reading, because all three were in my own work:

- **A regression.** javac's `memberType` returns `erasure(sym.type)` for a *raw* site — whatever the member declares. Reading `List<String> items()` under a raw `Holder` came back as `List` and matched no container, breaking a spec that compiled before. Two lenses found it independently.
- **A half-fix, whose missing half predates all of this.** `widensContainers` asked whether an element is *navigable* while `navigatorTarget` now answers something narrower, so a record navigating **into** such a container declared a traversal and returned a focus. Making it ask `navigatorTarget` restored the invariant that method's own javadoc claims — and closed two triggers with nothing to do with generics: `generateNavigators = false`, and `excludeFields`.
- **A justification that was simply wrong.** I said resolving the navigator needed substitution inside `WideningAnalysis`. A reviewer neutered the guard and read the errors: every one is in the *target's own* variables, which `WideningAnalysis` already returns. The real obstacle is a component written `Inner<?>` or raw. #739 carries the corrected reasoning.

Commit 3 also stopped declining **silently** — three places promised a navigator for any annotated field type (`@GenerateFocus`'s javadoc, the book's decision tree, and the generated class javadoc). It now emits the same note the file already uses for a name collision, and the docs say what is generated.

## The shared reading

Four sites needed *"read this member under the owner's instantiation"*, two of them in one class, so it lives in `ProcessorUtils` carrying every subtlety each site learned separately:

- a **record component** is read through its **accessor** — the member javac substitutes;
- a **raw** owner is read as declared, because javac erases every member of a raw type, including one naming no variable at all;
- an owner that merely **declares no parameters is not raw** — `UserDto` supplies none of its own, yet its inherited members are exactly what the walk is for.

That last distinction is the one neither earlier site had to make, and getting it wrong would have silently reintroduced instance five.

## Three axes, three bugs found

Each axis found a defect the moment it was written — which is the argument for them better than anything I could write here:

| axis | covers | found |
|---|---|---|
| #737 `GenericSpecInterfaceAxisTest` | `@ImportOptics` spec interfaces × 8 strategies and hints | `@InstanceOf` rejected every generic sealed hierarchy |
| `GenericImportedTypeAxisTest` | `@ImportOptics({Class...})` × record / wither / sealed / enum | raw prisms — commit 7 |
| `GenericMappingWireAxisTest` | `@GenerateMapping` × record / bean / mix-in × none / declared / inherited | the inherited bean property — commit 5 |

Every case holds the spec's parameter at `U` while the source declares `X`. **Where the two letters agree the fault is invisible** — which is exactly how #732 survived #727.

Commit 6 pinned the raw-prism cell *deliberately*, naming #742. After commit 7 the full suite produced **one failure out of 2088** — that cell, and nothing else. That is what pinning a known-wrong behaviour is for.

## Left open, deliberately

- **#739** — generate navigators for a generic target. The wildcard/raw component needs its own resolve-or-reject decision.
- **#733** — an `@InstanceOf` prism cannot check its type argument. The *unsound* sibling of commit 7, where the sealed hierarchy gives no guarantee.
- **#731** — `@GenerateIsos` on a generic method. A static field has nowhere to put `<T>`.

## Verification

`clean build` across all modules: BUILD SUCCESSFUL, EXIT=0. Every 100%-pinned class holds — `SpecInterfaceAnalyser`, `NavigatorClassGenerator`, `WideningAnalysis`, `BeanPropertyAnalyser`, `MappingProcessor`. Each fix has a regression test in the file that owns that behaviour, and every one was confirmed to fail against the unfixed source.

## Summary by CodeRabbit

* **New Features**
  * Improved support for generic records, beans, sealed hierarchies, and imported optics.
  * Generated lenses, traversals, prisms, and mappings now preserve applicable type parameters.
  * Generic inherited bean properties and instantiated field types are resolved more accurately.

* **Bug Fixes**
  * Prevented unsafe navigator generation for generic targets and clarified when `.via()` is required.
  * Improved diagnostics for unsupported generic hierarchies and mix-ins.

* **Documentation**
  * Updated Focus DSL guidance on navigator eligibility and generic types.


MagnusSmith:
@coderabbit full review

Issue #746, opened 2026-08-26
[BUG] Declaration-vs-instantiation: generic builders, @GenerateIsos and @GeneratePathBridge


MagnusSmith:
The declaration-vs-instantiation family, swept a sixth way — from the **generated output backwards**, asking of each generator "what does it emit that names a type, and where does that name come from?" That angle found three sites no earlier sweep reached. All three confirmed by compiling.

Two of them are in annotations no panel in this run has ever looked at, and both emit source that does not compile.

## 1. `BeanPropertyAnalyser` — a generic builder, one line from the line #745 changed

`findBuilderModel` reads the **instantiated** builder off `factory.getReturnType()`, then immediately discards the instantiation with `.asElement()`; `analyse` re-derives the owner from that element's own declaration:

```java
TypeElement builderType = (TypeElement) ((DeclaredType) factory.getReturnType()).asElement();  // Builder<String> → Builder
...
DeclaredType builderType = (DeclaredType) builder.builderType().asType();                      // Builder<T>
```

```java
public static Builder<String> builder() { ... }
public static final class Builder<T> { public Builder<T> name(T n) { ... } }
```

→ `@GenerateMapping: bean property 'name' on 'UserDto' is read and written at different types (java.lang.String vs T).`

The getter reads `String`; the setter parameter is read under `Builder<T>` as `T`. Naming a variable the author never wrote, in the diagnostic — the exact harm #740 fixed for the bean's own members, in the builder path beside it. A non-generic builder works; a builder *inheriting* from a generic base works.

Fix: carry `(DeclaredType) factory.getReturnType()` in `BuilderModel` rather than re-deriving from the element.

## 2. `@GenerateIsos` on a generic method — uncompilable output

`IsoProcessor:77` names the method's declared signature verbatim in a generated **static field**, which cannot declare type parameters.

```java
@GenerateIsos
public static <T> Iso<Box<T>, T> boxIso() { ... }
```
```java
public static final Iso<Box<T>, T> boxIso = Isos.boxIso();
//                          ^ error: cannot find symbol — symbol: class T
```

A `static final` field has nowhere to put `<T>`, so resolution is not available: the conforming call is **reject at the declaration**, as `MergeProcessor`, `AssemblyProcessor` and `ErrorEnvelopeProcessor` all do. `@GenerateIsos` is documented in `optics/iso.md` and used in `hkj-examples`.

[...]

PR #747, opened 2026-08-27
@GenerateIsos and @GeneratePathBridge: answer a generic declaration, one way or the other (#746)

MagnusSmith:
Closes the remaining two sites of #746 — the two annotations that accepted a generic declaration and then emitted source naming a type variable nothing declared, which the author met as `cannot find symbol` in a file they never wrote.

The two sites want opposite answers, and which one an annotation gets is decided by whether there is any correct source to emit:

```mermaid
flowchart TD
    A["annotation meets a<br/>generic declaration"] --> B{"can the generated<br/>shape name a type<br/>parameter?"}
    B -->|"yes — it emits a method<br/>or a class"| R["<b>resolve</b><br/>declare the parameters<br/>the signature names"]
    B -->|"no — it emits a<br/>static final field"| F["<b>refuse</b><br/>at the declaration,<br/>with a followable remedy"]
    B -->|"neither — emit it raw"| X["the bug<br/>#746 is about"]

    class B decision
    class R,F good
    class X bad
```

## `@GenerateIsos` — refuse

The iso is published as a `public static final` field initialised by calling the method. A field has nowhere to declare a `<T>`, so there is no correct source to emit; the method is refused where it is written, as `@GenerateMerge`, `@GenerateAssembly` and `@GenerateErrorEnvelope` already refuse theirs.

The gate asks what the **iso names**, not what the method declares — the distinction matters in both directions:

| declaration | before | now |
|---|---|---|
| `<T> Iso<Box<T>, T> boxIso()` | `cannot find symbol: class T` | refused: *the iso returned by 'boxIso' names a type variable* |
| `<T> Iso<Box, String> boxIso()` | generated fine | still generated — `T` is inferred at the call and never reaches the field |
| `Iso<Box<X>, X>` on an instance method of `Holder<X>` | `cannot find symbol` | refused as the type variable it is |

That last row is why the naive predicate is wrong: the method declares nothing, so a gate reading `getTypeParameters()` falls through and says *"make it static"* — a remedy this method cannot follow, because a static method of `Holder<X>` cannot name `X` either.

The same pass closes four more shapes the processor was silent on, each of which had been javac complaining about generated code:

- an **instance** method — the field's initialiser is a static call with no receiver
- a method **taking arguments** — the initialiser passes none
- a method or enclosing type the **generated package cannot see** (`private`, or package-private under `targetPackage`)
- a return type that is **not a two-argument `Iso`** — which for `void`, a primitive and an array had been a `ClassCastException` out of the processor, with no diagnostic at all

## `@GeneratePathBridge` — resolve

Here the generated shapes *can* name parameters, so they do. Two sets, landing in two places, and both are needed at once:

```java
@GeneratePathBridge
public interface BothRepo<T extends Comparable<T>> {
    @PathVia <R> Optional<R> convert(T from, Class<R> to);
}
```
```java
public final class BothRepoPaths<T extends Comparable<T>> {   // from the interface, bounds included
    private final BothRepo<T> delegate;                        // was raw
    public <R> OptionalPath<R> convert(T from, Class<R> to) {  // from the method
```

The raw delegate field was a `-Werror` build-breaker for *any* generic service interface, including ones whose `@PathVia` methods never mention `T`.

This also adds the mirror of the `@GenerateIsos` guard: a `@PathVia` method that is `static` or `private` is refused, because the bridge reaches its delegate and neither can be called that way.

## The axis

#746's real ask was the missing test. `GenericsAcceptanceAxisTest` holds one row per generating annotation, asserting which of the only two answers it gives, and compiles under the consuming build's own `-Xlint:unchecked,rawtypes -Werror` rather than javac's defaults — because on defaults, an annotation that drops its type parameters altogether looks exactly like success. Reverting the delegate field to raw passed all eleven rows *and all 2,079 tests in the module* before that change; two rows catch it now.

`@GeneratePrisms` was the **third** instance of this bug, not the second: on a generic hierarchy it emitted raw prisms, failing the same gate. #745 has since merged and settled it, so it now has a row like the rest — this branch is rebased onto that, and the row asserts the signature it produces.

## Verification

Every fix was confirmed by putting the defect back and watching the named test fail. `clean build` across all 190 tasks: BUILD SUCCESSFUL. Every line added is covered; two branches no test could reach were deleted rather than tested around.

A five-lens panel (correctness, type safety, ergonomics, architecture, test quality) ran against the first commit and found two real defects in the new guard plus the false `@GeneratePrisms` pointer — all addressed in the second commit. Defects it found in `PathProcessor` that are *not* about generics are filed separately rather than folded in here.

## Summary by CodeRabbit

* **Documentation**
  * Clarified requirements and troubleshooting guidance for generated isos and path bridges.
  * Expanded service wrapper mapping documentation, including supported effect types and generic bridges.

* **Bug Fixes**
  * Improved validation and diagnostics for invalid iso and path bridge declarations.
  * Generic type parameters are now preserved in generated path bridges.
  * Static and private path methods are rejected with clear compiler errors.

### DIFF
--- a/hkj-annotations/src/main/java/org/higherkindedj/hkt/effect/annotation/GeneratePathBridge.java
+++ b/hkj-annotations/src/main/java/org/higherkindedj/hkt/effect/annotation/GeneratePathBridge.java
@@ -67,6 +67,13 @@ import java.lang.annotation.Target;
  *   <li>{@code IO<T>} → {@code IOPath<T>}
  * </ul>
  *
+ * <h2>Type Parameters</h2>
+ *
+ * <p>The bridge holds one delegate of the annotated interface, so it declares whatever that
+ * interface declares, bounds and all: {@code Repo<T>} yields {@code RepoPaths<T>} wrapping a {@code
+ * Repo<T>}. A {@link PathVia} method that declares parameters of its own keeps them on the bridge
+ * method, where its arguments and return type name them.
+ *
  * @see PathVia
  */
 @Target(ElementType.TYPE)
--- a/hkj-annotations/src/main/java/org/higherkindedj/optics/annotations/GenerateIsos.java
+++ b/hkj-annotations/src/main/java/org/higherkindedj/optics/annotations/GenerateIsos.java
@@ -13,6 +13,13 @@ import java.lang.annotation.Target;
  * <p>This should be placed on a method that returns an Iso. The annotation processor will then
  * generate a static field containing the Iso instance.
  *
+ * <p>The method has to be {@code static} and declare no type parameters of its own, because the
+ * generated field is a {@code public static final} one initialised by calling it: a field has
+ * nowhere to declare type parameters, and a static initialiser has no instance to call an instance
+ * method on. A method that is either is refused at the declaration rather than generated for. Fix
+ * the type arguments where the method is declared - {@code Iso<Box<String>, String>} rather than
+ * {@code <T> Iso<Box<T>, T>} - or expose the method itself instead of a generated field.
+ *
  * <p>By default, the generated class is placed in the same package as the enclosing class. Use the
  * {@link #targetPackage()} element to specify a different package for the generated class.
  */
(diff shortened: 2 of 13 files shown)


MagnusSmith:
@coderabbit full review

PR #749, opened 2026-08-27
@GeneratePathBridge: emit a bridge the consuming build compiles (#748)

### LINKED ISSUE #748: [BUG] PathProcessor: IO returns, Validated wildcards and raw returns emit uncompilable or warning-laden bridges; inherited @PathVia is dropped silently
Five defects in `PathProcessor`, found by the review panel on #747 and each confirmed here by compiling. None is about generics, which is why they were left out of that PR rather than folded in.

They arise at three different points, and only the first is reported to the author at all:

```mermaid
flowchart LR
    A["@PathVia method"] --> B["determinePathType<br/>maps the return type"]
    B -->|"no case matches"| E["reported:<br/>Unsupported return type"]
    B -->|"case matches"| C["copy signature<br/>+ pick a Path factory"]
    C --> D["generated bridge"]
    D -.->|"1 IO: wrong factory<br/>2 wildcard: capture<br/>3 raw: unchecked"| X["javac, against<br/>generated source"]
    A -.->|"4 inherited: never seen"| S["silently absent"]
    C -.->|"5 varargs erased"| Q["compiles, call site worse"]

    classDef meh fill:#e5c890,stroke:#df8e1d,color:#232634
    class X,S bad
    class E ok
    class Q meh
```

## 1. Every `IO` return generates uncompilable source — HIGH

`PathProcessor:291` maps `org.higherkindedj.hkt.io.IO` to factory `"io"`, but the two are not compatible:

```java
public static <A> IOPath<A> io(Supplier<A> supplier)   // Path.java:434 — what is emitted
public static <A> IOPath<A> ioPath(IO<A> io)           // Path.java:475 — what is needed
```

`IO<A> extends IOKind<A>`, not `Supplier<A>`.

```java
@GeneratePathBridge public interface IoSvc { @PathVia IO<String> load(String key); }
```
```
IoSvcPaths.java:32: error: method io in class Path cannot be applied to given types;
    return Path.io(delegate.load(key));
  required: java.util.function.Supplier<A>
  found:    org.higherkindedj.hkt.io.IO<java.lang.String>
```

So the `IO` row is wholly non-functional, generic or not. It survives because no `@PathVia` in the tree returns `IO` and `PathProcessorIntegrationTest` has no `IO` case — while `GeneratePathBridge`'s javadoc advertises `IO<T> → IOPath<T>` as supported, and the processor's own rejection message lists `IO` among the supported types.

Fix: `"io"` → `"ioPath"`, plus a case in `PathProcessorIntegrationTest` that compiles the result.

## 2. A wildcard in a `Validated` return generates uncompilable source — HIGH

```java
@GeneratePathBridge public interface ValWild {
    @PathVia Validated<? extends CharSequence, String> check(String in);
}
```
```
ValWildPaths.java:35: error: method validated in class Path cannot be applied to given types;
  reason: inference variable E has incompatible equality constraints
          capture#2 of ? extends CharSequence, capture#3 of ? extends CharSequence
```

`errorType` is copied verbatim into the added `Semigroup<E>` parameter, so it captures separately from the delegate's own `Validated<capture, String>` and the two cannot unify. `Validated` is the only mapping affected — the same wildcards in `Optional` and `Either` returns compile clean.

## 3. A raw return type emits unchecked warnings into generated source — MEDIUM

`typeArgOrObject` (`PathProcessor:306`) substitutes `Object`, so the generated *type* is not raw, but the delegate call is an unchecked conversion:

```java
@PathVia Optional byId(Long id);   // →  public OptionalPath<Object> byId(Long id)
```
```
RawOptPaths.java:33: warning: [unchecked] unchecked method invocation: method optional ...
RawOptPaths.java:33: warning: [unchecked] unchecked conversion   (x2)
```

The build compiles with `-Xlint:unchecked,rawtypes -Werror` (`build.gradle.kts:58`), so this is a build failure in a file the author cannot edit — and the only place to put a suppression is generated, which the `hkj-checker` `@SuppressWarnings` gate would then object to in turn. Raw `Validated` is worse: three warnings plus a `Semigroup<Object>` no caller can satisfy soundly.

`PathProcessorCoverageTest`'s raw-type case asserts only that generation succeeds; it never compiles the result under lint.

## 4. An inherited `@PathVia` method is silently dropped — MEDIUM

```java
public interface Base<T> { @PathVia Optional<String> byId(T id); }

@GeneratePathBridge public interface Derived extends Base<String> {}
```

generates a `DerivedPaths` with a constructor and **no methods at all**, and reports nothing. `generateBridgeClass` walks `interfaceElement.getEnclosedElements()` rather than the members, so anything inherited is invisible.

Either read the members (`Elements.getAllMembers`, under the interface's instantiation — the reading this whole #746 family is about), or say why the annotation found nothing. Silence is the one answer that leaves the author with a bridge that compiles and does nothing.

## 5. Varargs are erased to an array — LOW

`Optional<String> first(String... candidates)` yields `public OptionalPath<String> first(String[] candidates)`. It compiles, and is arguably safer, but the bridge is meant to mirror the delegate and every existing call site has to change shape. `TypeName.get(param.asType())` loses the modifier; `MethodSpec.varargs(method.isVarArgs())` restores it.

## Also reported, not verified here

`TypeVariableName.get` drops a JSpecify bound — `interface NullSvc<T extends @Nullable Object>` generates `class NullSvcPaths<T>`. Nothing fails today (the generated file is not `@NullMarked`), but under a nullness checker the bridge would reject an instantiation the delegate accepts.

## Suggested split

1 and 2 are uncompilable output and worth doing first; 3 is a `-Werror` failure on a shape that at least has a workaround; 4 is a correctness/diagnostics question with a design choice in it; 5 is cosmetic.

Whatever the split, each fix wants a case in `PathProcessorIntegrationTest` that **compiles the generated source under the consuming build's flags** — `.withOptions("-Xlint:unchecked,rawtypes", "-Werror")`, as `GenericsAcceptanceAxisTest` and `SpecInterfaceProcessingTest` do. Plain `succeeded()` is what let 1 through 4 survive this long.

MagnusSmith:
Closes #748. Five defects in `PathProcessor`, all of one kind: the annotation accepted a declaration, and the file it wrote was one the consuming build refused, in a file the author cannot edit.

They arise at three different points, and only the first was reported to the author at all:

```mermaid
flowchart LR
    A["@PathVia method"] --> B["read the return type"]
    B -->|"no case matches"| E["reported:<br/>unsupported return type"]
    B -->|"case matches"| C["copy the signature<br/>+ pick a Path factory"]
    C --> D["generated bridge"]
    D -.->|"1 IO: wrong factory<br/>2 wildcard: capture<br/>3 raw: unchecked"| X["javac, against<br/>generated source"]
    A -.->|"4 inherited: never seen"| S["silently absent"]
    C -.->|"5 varargs erased"| Q["compiles, call site worse"]

    classDef meh fill:#e5c890,stroke:#df8e1d,color:#232634
    class X,S bad
    class E ok
    class Q meh
```

Each was reproduced before it was fixed, by compiling the generated source under the consuming build's own flags.

## The five

**`IO` was wholly non-functional.** The table mapped `IO` to `Path.io`, which takes a `Supplier`; `IO<A> extends IOKind<A>` is not one. `Path.ioPath(IO<A>)` is the factory that was wanted. Nothing in the tree returns `IO` from a `@PathVia`, and the coverage test asserted the failure as expected behaviour, so a row the javadoc advertised had never worked.

**A wildcard `Validated` error type is refused.** The bridge names that type twice, in the `ValidationPath` it returns and in the `Semigroup` it asks the caller for, and a wildcard captures separately at each mention, so no argument satisfies both. There is no signature that both mirrors the delegate and can be called, so it is refused at the declaration. Wildcards elsewhere are named once and carry through unchanged, `Validated`'s own value type included:

| return type | before | now |
|---|---|---|
| `Validated<? extends CharSequence, String>` | `inference variable E has incompatible equality constraints` | refused, at the declaration |
| `Validated<String, ? extends Number>` | compiled | compiles, unchanged |
| `Validated<List<? extends CharSequence>, String>` | compiled | compiles, unchanged |
| `Optional<? extends CharSequence>`, `Either<?, ?>`, `Maybe<?>`, `Try<?>` | compiled | compile, unchanged |

**A raw type is refused wherever the bridge writes it.** `typeArgOrObject` substituted `Object`, which made the delegate call an unchecked conversion three times over. The remedy is followable, and the `Object` fallback is gone rather than tested around.

**An inherited `@PathVia` is bridged.** The walk read `getEnclosedElements()`, so `Derived extends Base<String>` generated a bridge with a constructor and no methods, and said nothing. It reads members now, under the annotated interface's own instantiation, which is what the rest of the #746 family is about: an inherited `<R extends T>` is `<R extends String>` in the bridge, and `T` is a name the bridge cannot write.

**Varargs are kept.** Except where a `Semigroup` follows them, which leaves the array no longer last, or where the element type is non-reifiable. That second exclusion is not fastidiousness: `T...` on the bridge is a `[unchecked] Possible heap pollution` warning, so copying it naively would have manufactured the third defect again.

## What the review panel then found

A five-lens panel (correctness, type safety, ergonomics, architecture, test quality) ran against the first commit. Reading inherited members had opened three new ways to emit source javac refuses, and each of the five fixes had stopped one position short of the defect it named. The second commit answers all of it.

**Regressions from the members walk.** A method type parameter sharing a name with one of the interface's now sits in one scope with it, where the method's hides the interface's, so an inherited `<T extends U>` on a `Derived<T>` was written `<T extends T>` -- a cycle. It is refused, but only where the signature depends on what is hidden; a name shadowed and never needed still generates unchanged. Two unrelated superinterfaces declaring the same method override nothing, so `getAllMembers` returns both and the bridge declared it twice; members are deduplicated by erased signature. A diagnostic against a method inherited from a compiled dependency had no file and no line, so those are reported against the annotated interface, naming where the method was declared. And the `{@link}` on an inherited method is a real import, which `targetPackage` could make unreachable.

**The same defect, one position further out.** Raw types were caught in the return type's head only, so `Optional<List>` and `foo(List raw)` still wrote `[rawtypes]` into the generated file; every position the bridge copies is walked now. Reachability was gated on bounds only, so `targetPackage` still emitted unreachable parameter, return and delegate types.

**Pre-existing shapes the new gate exposed.** A `throws` clause was dropped, leaving callers an unreported checked exception. A self-referential bound such as `<R extends Enum<R>>` left the `Path` factory's argument position with nothing to infer from, so the delegate call carries an explicit type witness. A parameter named `delegate` shadowed the field and turned the call into one on itself; the body says `this.delegate`. A parameter named `semigroup` collided with the appended one, which now takes the first free name. A `@PathVia(name = ...)` that is not an identifier reached javapoet and threw, showing the author a processor stack trace instead of the attribute they mistyped.

**One panel claim was right about my own reasoning.** I had justified two refusals with "generated source cannot carry a suppression". That is false -- five sites in this repo emit `@SuppressWarnings` into generated code, and the 0.4.10 note for #730 says so out loud. The refusals stand, because `OptionalPath<Object>` loses the element type at every call site, but they now give their real reason.

Diagnostics move to the shared what/why/fix format, as `Diagnostics`' javadoc asks of a processor being touched, and the book quotes each message as emitted so the text a user saw is the text they can search for. `ProcessorUtils` gains `memberOf` and `boundsOf` rather than `PathProcessor` re-settling either privately.

## Verification

`clean build` across all 190 tasks: BUILD SUCCESSFUL. `PathProcessor` is at 100% line, branch and instruction, and joins the pinned list -- which then caught five new paths I had probed but not turned into tests, and they are tests now.

Every case in `PathProcessorIntegrationTest` compiles the generated source under `-Xlint:unchecked,rawtypes -Werror`, the existing cases as well as the new: plain `succeeded()` is what let four of the five original defects survive, and it is what the panel used to find the rest. The panel's mutation pass put each defect back one at a time and named the test that failed; the one line it found *no* test for -- the substituted read of the return type -- has the case that kills it now.

## Not addressed here

`TypeName.get` drops type-use annotations, so `interface NullSvc<T extends @Nullable Object>` generates `class NullSvcPaths<T>`. It is javapoet-wide rather than anything this processor does, and it affects all 26 generator classes equally, so it wants its own change rather than a corner of this one: filed as #750.

I had said here that nothing fails today because generated files are not `@NullMarked`. That is wrong, and #750 has the evidence: `@EffectAlgebra` and `@ComposeEffects` both stamp `@NullMarked` on what they emit, and inside such a file the dropped annotation does not merely go missing, it reads as an assertion of non-null. Nothing in `@GeneratePathBridge`'s own output is affected, since its files carry no nullness contract.


MagnusSmith:
@coderabbit full review

MagnusSmith:
All seven taken, in `969a83f3d`.

**The `firstRawIn` wildcard gap is real, and it went one further than reported.** Confirmed by compiling under the consuming build's flags before touching anything: `Optional<? extends List>` reached the generated file and put `found raw type: java.util.List` at `/SOURCE_OUTPUT/…Paths.java:31`, i.e. in the file the author cannot edit, which is exactly the failure the walk exists to prevent. The observation that `ProcessorUtils.firstUnreachableIn` already walks wildcard bounds is the useful part: the two walks agreed about visibility and disagreed about rawness for the same nesting, which is how the gap stayed invisible.

Fixed by descending into both bounds, and also into the enclosing link of a declared type, which `firstUnreachableIn` walks and I had not — `Outer.Inner` is raw in `Outer` even where `Inner` declares nothing of its own.

Probing it surfaced a fourth position not in the review: the **interface's own type parameter bounds** are repeated verbatim in the bridge's declaration and were not rawness-checked at all, so

```java
@GeneratePathBridge
public interface Repo<T extends List> { @PathVia Optional<String> byId(T id); }
```

generated `class RepoPaths<T extends List>` and warned in the generated file. Gated with the rest, and both new shapes have tests that compile the result under `-Xlint:unchecked,rawtypes -Werror`.

**Two of the documentation findings were wrong rather than unclear, and both are worth calling out.**

The unsupported-return-type entry read *"something outside the bridged set: `Optional`, `Maybe`, `Either`, `Try`, `Validated`, `IO`"* — the colon makes the list parse as examples of what is **not** bridged, which is the exact opposite of what those six are. Reworded to name the set as the set. I have not taken the suggested wording verbatim, because dropping the list loses the answer to the question the reader arrived with; it now reads *"The bridged set is … ; `CompletableFuture` is the type most often met outside it."*

The no-`@PathVia` entry claimed *"neither the interface nor anything it extends has a `@PathVia` method"*, which its own following paragraph contradicts — an unannotated override hides an annotated parent, so the parent does have one. Reworded to talk about what survives among the members.

The other three (the errors-versus-warning intro, the `IO`/`Path.ioPath` sentence, and the missing word in the cheatsheet) are taken as described, with wording adjusted to the page's own voice rather than pasted, and without the `</li>` the annotation-javadoc suggestion added, since that file's list items are written unclosed throughout.

`clean build` across all 190 tasks is green, and `PathProcessor` stays at 100% line, branch and instruction.

Issue #750, opened 2026-08-27
[BUG] Generated code drops every type-use annotation, so a @Nullable component reads as non-null inside an @NullMarked generated file


MagnusSmith:
`TypeName.get(TypeMirror)` never copies type-use annotations, and `TypeVariableName.get` reads bounds off the *declaration* and then strips `Object` from them. Between them, every annotation JSpecify uses is dropped on the way into generated source. All 26 generator classes in `hkj-processor` go through those two methods, so this is uniform rather than any one generator's bug.

For most of them it is latent, because their output carries no nullness contract. For `@EffectAlgebra` and `@ComposeEffects` it is live today, because those two are the only generators that stamp `@NullMarked` on what they emit — and inside a `@NullMarked` file, an unannotated type *means* non-null. The annotation is not merely lost there; its absence asserts the opposite of what the author wrote.

```mermaid
flowchart TD
    A["author writes<br/>@Nullable String note"] --> B["TypeName.get(TypeMirror)"]
    B -->|"type annotations<br/>not copied"| C["String"]
    C --> D{"does the generated file<br/>carry @NullMarked?"}
    D -->|"no — 24 generators"| L["latent:<br/>unannotated, no contract"]
    D -->|"yes — @EffectAlgebra,<br/>@ComposeEffects"| X["live:<br/>reads as non-null"]

    E["author writes<br/>&lt;T extends @Nullable Object&gt;"] --> F["TypeVariableName.get"]
    F -->|"reads the declaration's bounds,<br/>then strips Object"| G["&lt;T&gt;"]
    G --> D

    classDef bad fill:#e78284,stroke:#d20f39,color:#232634
    classDef meh fill:#e5c890,stroke:#df8e1d,color:#232634
    class X bad
    class L meh
```

## The live case

```java
@EffectAlgebra
public sealed interface NoteOp<A> permits NoteOp.Write {
    record Write<A>(@Nullable String note) implements NoteOp<A> {}
}
```

generates, in `NoteOpOps.java`:

```java
@NullMarked
@Generated
public final class NoteOpOps {

  public static <A> Free<NoteOpKind.Witness, A> write(String note) {
    NoteOp<A> op = new NoteOp.Write<>(note);
    return Free.liftF(NoteOpKindHelper.NOTE_OP.widen(op), functor());
  }
```

[...]

PR #754, opened 2026-08-27
@GenerateFocus: widen a Set or Collection through the Each that rebuilds it (#725)

### LINKED ISSUE #725: [BUG] A `Set` component's generated traversal throws at runtime
`Set` and `Collection` are two of the five containers `@GenerateFocus` recognises by name, and both widen through the no-arg `.each()`. That method is `List`-only, and says so:

```java
/** @throws ClassCastException if the focused type {@code A} is not a {@code List} */
@SuppressWarnings("unchecked")
default <E> TraversalPath<S, E> each() {
  return via((Traversal<A, E>) FocusPaths.listElements());
}
```

The cast is unchecked, so the generated call compiles. It fails on first use.

## Measured

On `main`, one path built by hand exactly as `SetHolderFocus.tags()` builds it:

```java
record SetHolder(Set<String> tags) {}

FocusPath.of(Lens.of(SetHolder::tags, (s, v) -> new SetHolder(v)), "tags")
    .each()
    .modifyAll(String::toUpperCase, new SetHolder(new HashSet<>(List.of("a", "b"))));
```

```
java.lang.ClassCastException: class java.util.HashSet cannot be cast to class java.util.List
```

A `Collection<A>` component fails the same way whenever the runtime value is not a `List`; when it is one, the traversal quietly rebuilds it as a `List`, which a `Collection`-typed component accepts.

## Why

```mermaid
flowchart TD
    D["Set&lt;A&gt; component"] --> R["recognised container table"]
    R --> E[".each() - no argument"]
    E --> L["FocusPaths.listElements()"]
    L --> C["(Traversal&lt;A, E&gt;) cast - unchecked"]
    C --> X["modifyF(f, List&lt;E&gt; source, app)"]
    X --> B["ClassCastException at the first traversal"]

    class D wire
    class R,E,L,C,X domain
    class B error
```

The recognised-container table pairs a type with a *tier*, and the expression that reaches that tier is assumed to be `.each()` for all three of `List`, `Set` and `Collection`. Only `List` can honour it.

## What a fix looks like

`EachInstances.setEach()` already exists, backed by `TraverseTraversals.forSet()`, so `Set` is one expression away:

```java
// today
FocusPath.of(...).each()
// wanted
FocusPath.of(...).each(EachInstances.setEach())
```

`Collection` has no stock `Each`. Either give it one (`fromIterableCollecting` with a list collector rebuilds a `List`, which a `Collection`-typed component accepts, at the cost of turning a `Set`-valued field into a `List`), or stop recognising it and let it reach a traversal through the SPI like every other container the processor cannot rebuild.

The return types do not change either way — this is the expression, not the tier — so the fix is not source-breaking.

## Notes

- Not a #719 divergence: both the static Focus method and the navigator method emit the same wrong expression, so they agree. #719 removed the navigator's Collection-**subtype** widening for exactly this reason (measured `ClassCastException` rebuilding an `ArrayList` as a plain `List`), leaving `Set` and `Collection`, which are recognised by name, still doing it.
- Found while addressing #719 and deliberately left out of that fix, which was about the two analyses disagreeing.
- A runtime law-check over the recognised-container table — modify through each generated traversal and read back — would have caught it; the compile-testing suite cannot, because the failure is a runtime cast.

MagnusSmith:
Closes #725.

`@GenerateFocus` recognises five containers by name and widened three of them — `List`, `Set` and `Collection` — through the same no-argument `.each()`. That method carries a `List` traversal behind an unchecked cast, so a `Set` component compiled and then threw on its first traversal, read or write. No `Set`-valued Focus path has ever worked.

```mermaid
flowchart TD
    D["Set&lt;A&gt; component"] --> R["recognised container table"]
    R --> B["before: one step for all three<br/>.each() → FocusPaths.listElements()"]
    R --> A["after: one step per type"]
    B --> X["ClassCastException at the first traversal"]
    A --> L["List → .each()"]
    A --> S["Set → .each(EachInstances.setEach())"]
    A --> C["Collection → .each(EachInstances.collectionEach())"]
    L --> OK["rebuilds what it was handed"]
    S --> OK
    C --> OK

    class D wire
    class R,A,L,S,C,OK domain
    class B,X error
```

The recognised-container table was a `Set<String>` paired with a tier; it is now a `Map<String, StepKind>` pairing each type with the step it widens through. Tiers are unchanged, so **no generated signature moves**.

`Collection` had no stock `Each`, so `Traversals.forCollection()` and `EachInstances.collectionEach()` are new. I did not take the issue's list-collector suggestion: rebuilding every source as a `List` breaks the traversal identity law for a set-valued component and lets a many-to-one modification leave duplicates in a collection that had none. It rebuilds the shape it was handed instead.

## Behaviour changes

**Source-breaking at the declaration.** Naming an `Each` gives the widening type arguments javac infers from the component, which is the #718 rule. A raw or wildcard-carrying `Set`/`Collection` is now rejected where it is declared, at every nesting depth the analysis reaches. Every one of those shapes compiled in 0.4.9 and threw `ClassCastException` on the first traversal, so nothing that worked stops working. `Optional`, `Maybe` and `List` are unaffected and still take a wildcard silently — deliberately, and the class javadoc now says why routing `List` through an instance too would be a regression.

**One set-rebuild policy.** `EachInstances.setEach()` traversed through `Set.copyOf`, whose iteration order the JDK randomises per JVM start and which rejects nulls. Six consecutive runs of an identity modification over the same `LinkedHashSet`:

```
[A, B, C, D, E, F, G]  [D, E, F, G, A, B, C]  [G, A, B, C, D, E, F]
[C, D, E, F, G, A, B]  [A, G, F, E, D, C, B]  [B, C, D, E, F, G, A]
```

`getAll` kept source order, so read and write disagreed; a `TreeSet` with a case-insensitive comparator came back a plain set, flipping `contains("alice")` from true to false under an identity modify. `setEach()` now uses `Traversals.forSet()`, which every other annotation already emitted. The policy is written once in `traverseSet` — order preserved, nulls carried through, and (as `traverseList` already did) an **unmodifiable** set handed back. `SetGenerator`'s generated `modifyF` calls it too, instead of `Collectors.toSet()`. Four routes to a set traversal, three of which disagreed, are now one.

`Traversals.forSet()`/`traverseSet` previously returned a mutable `LinkedHashSet`; code that mutated a set traversal's result in place now meets `UnsupportedOperationException`.

**A documented limit.** `forCollection` cannot be lawful over all of `Collection`: an `ArrayDeque` or `PriorityQueue` comes back a `List`, so identity does not return an *equal* collection — those types inherit identity `equals` and no rebuild could satisfy it. A `SortedSet` keeps its elements but not its comparator. Both are pinned by tests, the second as an explicit non-law.

## Tests

A compile-testing assertion cannot see a failure that is a runtime cast — which is why the issue went unnoticed. `RuntimeCompilationHelper.compileWith(Processor, …)` lets `hkj-processor` tests run generated code, and `FocusRecognisedContainerTest` compiles a holder with the real processor, loads the generated class, and modifies through every recognised container before reading the component back. **Reverting the table alone fails nine of its cases with the issue's `ClassCastException`.** `GeneratedFocusSourceCompilesTest` gains `Collection` across every shape and setting; the diagnostic's second remedy is a test case rather than a claim.

`gradle clean build` green; `WideningAnalysis` still pinned at 100% line/branch/instruction.

## Review

Five-lens panel (correctness, type safety, immutability, ergonomics, architecture) run after implementation. The `setEach()` randomisation and the mutable-set leak were panel findings I reproduced independently before acting; the release-history and `compiler_errors` entries, the uncompilable skill-table row, and the two "Use" tables still answering `each()` for a `Set` were flagged there too. The panel confirmed the #719 invariant holds — a navigator still composes the static method and cannot disagree with it.

Left as follow-ups, not done here, each verified against the current tree and filed:

- #756 — `@GenerateTraversals` silently skips a `Collection` component; no SPI generator claims it, and nothing is reported
- #757 — `ListGenerator`/`SetGenerator`/`OptionalGenerator` carry unreachable optic-expression overrides that now duplicate the recognised table
- #758 — `FocusProcessor`'s rejection walk still hand-mirrors `WideningAnalysis.collect()` rather than reading a declined step from it
- #759 — `ProcessorUtils.simpleTypeName` prints a type-use annotation into the diagnostic (`@Nullable Set<?>`); pre-existing, and distinct from #750

https://claude.ai/code/session_01KpzttKFH5oXn7zjohgHDg1

## Summary by CodeRabbit

* **New Features**
  * Added traversal support for general `Collection` fields, preserving sets as sets and rebuilding other collections as lists.
  * Improved `Set` traversal to preserve iteration order, carry nulls, remove duplicates, and return unmodifiable results.
  * Generated focus paths now select the appropriate traversal behavior for `List`, `Set`, and `Collection`.

* **Bug Fixes**
  * Prevented incorrect collection widening that could cause runtime `ClassCastException`s.
  * Added clearer diagnostics for raw or wildcard `Set` and `Collection` declarations.

* **Documentation**
  * Expanded guidance and examples for collection traversal, wildcard handling, and generated focus paths.


MagnusSmith:
@coderabbit full review

PR #760, opened 2026-08-28
A generated prism's focus has to be a variant of its source (#755)

### LINKED ISSUE #755: [BUG] A spec-interface prism whose focus is not a subtype of the source generates source that does not compile
Every prism generated from a spec interface builds its `A -> S` side as identity:

```java
public static Prism<Value, String> string() {
  return Prism.of(
    source -> source.isString() ? Optional.of(source.asString()) : Optional.empty(),
    value -> value);
}
```

`Prism.of(Function<S, Optional<A>> getter, Function<A, S> builder)` needs that second function to produce an `S`, and `Prism.build(A)` is the method it stands up. Identity only type-checks when `A` *is* an `S`. Where it is not, the generated file does not compile, and the author meets it as an error in a file they never wrote:

```
error: incompatible types: cannot infer type-variable(s) S,A,T,T
  (argument mismatch; bad return type in lambda expression
    java.lang.String cannot be converted to com.external.Value)
```

There is no line of the spec interface in that message, and no diagnostic from the processor at all.

## The shape is one the documentation recommends

`@MatchWhen`'s own javadoc:

```java
@MatchWhen(predicate = "isText", getter = "asText")
Prism<JsonNode, String> text();     // String is not a JsonNode
```

[Spec Interfaces](https://higher-kinded-j.github.io/optics/optics_spec_interfaces.html), under *`@InstanceOf` versus `@MatchWhen`*:

```java
@MatchWhen(predicate = "isString", getter = "asString")
Prism<Value, String> string();      // String is not a Value
```

and the same page offers `@MatchWhen` "on the generated `hasX`/`getX` oneof accessors" for Protocol Buffers, where a oneof field's type is never the message type. None of those three compile.

## Why it survived

Every `@MatchWhen` and `@InstanceOf` case in the processor's tests uses a focus that *is* a subtype of the source — `Prism<JsonNode, ArrayNode>`, `Prism<Result, Success>`, `Prism<Base, Sub>` — and one fixture sidesteps the question entirely with `Prism<JsonNode, JsonNode>`. The book snippets on that page carry no `<!-- verify -->` marker, so nothing compiles them either. The shape the docs recommend is the one shape nothing exercises.

```mermaid
flowchart TD
    D["Prism&lt;Value, String&gt;<br/>@MatchWhen(isString, asString)"] --> G["getOptional: Value → Optional&lt;String&gt;<br/>reads the getter, fine"]
    D --> B["build: String → Value<br/>generated as value -> value"]
    B --> Q{"is the focus<br/>a subtype of<br/>the source?"}
    Q -->|"ArrayNode &lt;: JsonNode"| OK(["compiles"])
    Q -->|"String is not a Value"| X["error inside the generated file<br/>no processor diagnostic"]

    class D,G,B,OK tier
    class Q decision
    class X error
```

## Both hints, and it is not the same gap as #733

`@MatchWhen` reaches it through any check-and-extract API that yields a value rather than a subtype. `@InstanceOf` reaches it wherever the narrowed type is not itself an `S`: a target reached through a supertype the focus names, `Prism<Base, Marker>` for a `Sub implements Base, Marker`, or `Prism<Node<U>, Twin<?>>` for a `Twin<X> extends Node<Pair<X, X>>`.

[#733](https://github.com/higher-kinded-j/higher-kinded-j/issues/733) (PR [#753](https://github.com/higher-kinded-j/higher-kinded-j/pull/753)) constrained the *narrowing* side — what the test may promise about the value it hands back. This is the other function: what the prism does with that value on the way back. The two checks are independent, and the narrowing one does not imply this one.

## Why it is not simply a missing check

Rejecting the declaration is the floor, and it matches the arm [#718](https://github.com/higher-kinded-j/higher-kinded-j/issues/718), [#723](https://github.com/higher-kinded-j/higher-kinded-j/issues/723) and #733 established: refuse what cannot be written soundly, at the declaration, with what/why/fix.

But `Prism<JsonNode, String>` is a prism people legitimately want, and the docs promise it. A prism onto a value type needs a real build side — `TextNode.valueOf(String)` for Jackson, a `Value.of(String)` factory, a constructor — and the spec API has no way to name one. So a check alone converts a silent failure into "you cannot do this", for a case the library arguably should support.

Naming note: the `A -> S` side is `build` on HKJ's `Prism`, and `builder` on `Prism.of`. `reverseGet` is `Iso`'s name for its own total inverse and should not be borrowed here.

## Options

1. **Reject at the declaration.** One check on both hints: the focus type must be assignable to the source type. Honest, cheap, and turns an error in generated source into one on the offending method. Leaves the value-typed prism unbuildable.
2. **Reject, and add a `builder` attribute** naming the build side, on both hints — `@MatchWhen(predicate = "isText", getter = "asText", builder = "com.fasterxml.jackson.databind.node.TextNode.valueOf")`. A string naming a static factory has precedent in `@TraverseWith` and `@ViaCopyAndSet`; the attribute would be required exactly when the focus is not a subtype, and rejected as redundant when it is. This is what makes the documented examples work.
3. **Reject, and fix the documentation** to only show subtype focuses, dropping the value-typed prism as a supported shape.

Whichever is chosen, the three documented examples need correcting or making true, and the page's snippets want `<!-- verify -->` markers so the next one cannot drift the same way.

Found while fixing #733.

MagnusSmith:
Closes #755, taking the issue's **option 1**: reject at the declaration, and correct the documentation that recommended the shape. The `builder` attribute (option 2) was considered and declined — see *The road not taken* below.

A prism runs both ways. Every prism generated from a spec interface builds back with identity — `value -> value`, the function `Prism.build` stands up — and that is only a source when the focus is one. Nothing checked it, so a focus that was a *value* rather than a variant reached javac inside the generated file:

```
error: incompatible types: cannot infer type-variable(s) S,A,T,T
  (argument mismatch; bad return type in lambda expression
    java.lang.String cannot be converted to com.external.Value)
```

No line of the spec interface in it, and no diagnostic from the processor at all.

## The check

One check, at the call site rather than in either hint, because the requirement belongs to the prism:

```mermaid
flowchart TD
    P["Prism&lt;S, A&gt;"] --> H{"hint parses?"}
    H -->|"no"| M(["needs @InstanceOf or @MatchWhen<br/>the nearer problem"])
    H -->|"yes"| B{"is A an S?"}
    B -->|"TextNode &lt;: JsonNode"| OK(["generated"])
    B -->|"String is not a JsonNode"| E(["rejected at the declaration"])

    class P,OK tier
    class H,B decision
    class M,E error
```

It is asked *after* the hint parses, so a method carrying no hint is still told that first — one problem, one error, and the nearer one.

Both hints are held to it by construction. That includes an `@InstanceOf` whose narrowing is perfectly sound but reaches the focus through a supertype the source does not share — `Prism<Base, Marker>` for a `Sub implements Base, Marker` — and `Prism<Node<U>, Twin<?>>` for a `Twin<X> extends Node<Pair<X, X>>`, which #753's narrowing check passes and this one stops. The two checks are independent: #753 constrains what the test may *promise* about the value; this constrains what the prism does with it on the way back.

## The documentation recommended the broken shape

In three places, none of which compiled:

| Where | Was | Now |
|---|---|---|
| `@MatchWhen` javadoc | `Prism<JsonNode, String> text()` | `Prism<JsonNode, TextNode> text()` |
| Spec Interfaces | `Prism<Value, String> string()` | `Prism<Value, ObjectValue> object()` |
| Spec Interfaces, Protocol Buffers | `@MatchWhen` on any oneof accessor | qualified: message-typed fields, not scalars |

Both the annotation and the page now say why, and name `Prism.of` with a real build side as the hand-written route where the value type is the point. The book's main worked example was already correct — it focuses `StringNode`, `NumericNode` and friends, which is why the fixture that actually compiles never caught this.

## The road not taken

The issue offered a `builder` attribute naming a static factory, which would have made `Prism<JsonNode, String>` expressible rather than merely rejected:

```java
@MatchWhen(predicate = "isText", getter = "asText",
           builder = "tools.jackson.databind.node.TextNode.valueOf")
Prism<JsonNode, String> text();
```

Declined. A spec interface says *which* optic to generate and how to narrow it; naming an arbitrary construction expression is a different job, and one `Prism.of` already does directly and more legibly. The diagnostic points there, so the capability is not lost — only the indirection. That keeps `@InstanceOf` and `@MatchWhen` to what they are: two ways of saying how a value is recognised.

**Still open, and not this PR:** the Spec Interfaces snippets carry no `<!-- verify -->` markers, which is why all three drifted from the compiled fixture beside them. Worth its own pass.

## Verification

- `PrismBuildSideTest`: 7 cases — both hints rejected, both hints accepted, focus-is-source, and the hint-ordering case. Success cases compile the generated source under `-Xlint:unchecked,rawtypes -Werror`.
- The failing shapes were reproduced on merged `main` first, through the processor, so the diagnostics replace a real error rather than a supposed one.
- No existing test changed: every prism in the suite already focused a subtype, which is precisely why the gap survived.
- `gradle build` and `gradle :hkj-processor:check` green, coverage pins included.

https://claude.ai/code/session_01X1fwBkaAvGfoHg1Z1KtedC

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to ensure generated prisms focus on source-compatible variants that can be rebuilt correctly.
  * Improved diagnostics for invalid prism focus types.

* **Documentation**
  * Corrected `@MatchWhen` examples and clarified when to use payload-focused optics.
  * Documented the new validation behavior and migration guidance in the optics reference and release notes.

* **Tests**
  * Added coverage for valid and invalid prism focus types, diagnostic precedence, and generated code behavior.


MagnusSmith:
@coderabbit full review

PR #765, opened 2026-08-28
hkj-book: apply the exemplar treatment to the optics chapter front matter

MagnusSmith:
The three pages above `ch1` — the chapter landing page, the Quickstart and Annotations at a Glance — were the last of the optics chapter left untreated. The landing page had drifted furthest.

## The hierarchy diagram taught subtyping that does not exist

`ch_intro.md`'s ASCII diagram captioned its arrows "can be used as". Against the declarations, only one such arrow is real:

```mermaid
flowchart LR
    A["Getter extends Fold"] --> B["the only subtype relation<br/>between two optic types"]
    C["Lens, Prism, Iso, Affine,<br/>Traversal, Setter, Fold"] --> D["all extend Optic directly:<br/>siblings, not a hierarchy"]
    classDef note fill:#e5c890,stroke:#df8e1d,color:#232634
    class A,C ok
    class B,D note
```

So "a `Lens` can stand in wherever a `Getter` or `Fold` is expected" was false — `lens.asFold()` is an explicit conversion. "An `Iso`, the most specific optic, can stand in as any of the others" was false twice: it has `asLens`, `asTraversal` and `asFold`, and no route to a `Prism` or `Affine` at all. The chapter's own [Conversions](https://github.com/higher-kinded-j/higher-kinded-j/blob/main/hkj-book/src/optics/conversions.md) page already documented these as conversions.

The ASCII art is now a mermaid diagram of what the types declare, with the arrows labelled as `extends`.

The same page also still described **five** sections and listed neither Advanced Optics nor Reference.

## The Quickstart's collection example did not work

A `List` component's generated accessor is already element-level — the processor emits `FocusPath.of(lens, "items").each()` — so the page's own `.each()` was one too many. It compiles, because `each()`'s type parameter is inferred from the assignment through an unchecked cast, and then throws `ClassCastException`, exactly as `TraversalPath.each()`'s javadoc says.

The prism example did not compile at all: `Prism.modify` is `Function<A, A>`, so it rebuilds the variant it narrowed to and cannot turn a `Pending` into a `Shipped`.

Both are now **under the verify gate** with a fixture, so the next person to break them is told. The gate caught a further mistake in my own first fix, which is the point of it.

## Annotations at a Glance

`@TraverseField` was filed among the spec-method hints, under a sentence saying to apply them to abstract methods — it targets a `RECORD_COMPONENT`. It now has its own section. The page also omitted `@GenerateAssembly`, `@GenerateMerge` and `@GenerateErrorEnvelope` while calling itself the complete surface. The "eight spec-method hints" count is correct and now stated as what it is: 2 prism matchers, 4 copy strategies, 2 traversal hints.

## Also

Key Takeaways and See Also throughout; the chapter intro was the only one of the eight without a See Also. Ratchet 313 → 315.

Local gates: `gradle :hkj-examples:bookVerify` green, 34 mermaid diagrams parse clean, every link on the three pages resolves.

## Summary by CodeRabbit

- **Documentation**
  - Expanded optics introductions with clearer capability diagrams, relationships, focus cardinality, and navigation.
  - Updated the quickstart with improved status, traversal, and prism examples, plus additional guidance and tutorials.
  - Restructured the annotations reference with expanded annotation coverage, clearer setup guidance, decision support, and takeaways.

- **Tests**
  - Added verification coverage for the updated optics quickstart examples.
  - Increased the minimum number of documentation snippets validated by the compilation checks.

### DIFF
--- a/hkj-book/src/optics/annotations_at_a_glance.md
+++ b/hkj-book/src/optics/annotations_at_a_glance.md
@@ -5,7 +5,7 @@
 ~~~admonish info title="What You'll Learn"
 - Which annotation to apply for each optic type, and what target each annotation expects (record, sealed interface, enum, method, package-info).
 - How to choose between `@ImportOptics` and an `OpticsSpec` interface for types you don't own.
-- The eight spec-method hints that drive optic generation for external types: prism matchers, lens copy strategies, traversal hints, and `Kind`-field configuration.
+- The eight spec-method hints that drive optic generation for external types: two prism matchers, four lens copy strategies, and two traversal hints.
 - Where in the book to look for the in-depth treatment of each annotation.
 ~~~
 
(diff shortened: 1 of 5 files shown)


MagnusSmith:
@coderabbit full review

PR #772, opened 2026-08-29
@GenerateTraversals: see a Collection component, and say so when one is skipped (#756)

### LINKED ISSUE #756: [BUG] @GenerateTraversals silently skips a Collection component

`@GenerateTraversals` generates nothing for a `Collection<A>` component, and says nothing about it. The build succeeds, the generated class is simply missing a method, and the author finds out at the call site.

## Measured

On `fix/725-set-collection-each`:

```java
@GenerateTraversals
public record Job(Collection<String> tags, List<String> notes) {}
```

```
GenerateTraversals status: SUCCESS   (no error, no warning)
```

```java
public final class JobTraversals {
  private JobTraversals() {
  }

  public static Traversal<Job, String> notes() { ... }
  // no tags()
}
```

`JobTraversals.tags()` does not exist. `notes()` is generated normally, so nothing signals that the component beside it was skipped.

## Why

`Collection` is recognised by four different tables, and only one of them has it.

```mermaid
flowchart TD
    C["Collection&lt;A&gt; component"] --> F["WideningAnalysis.COLLECTION_TYPES<br/>@GenerateFocus"]
    C --> S["hkj-processor-plugins SPI providers<br/>@GenerateTraversals"]
    C --> T["TypeKindAnalyser<br/>@ThroughField auto-detect"]
    C --> R["TraversalCodeGenerator<br/>@ImportOptics / @OpticsSpec"]

    F --> Y["recognised since #725"]
    S --> N1["not listed — component skipped, silently"]
    T --> N2["not listed — rejected as an unrecognised container"]
    R --> N3["no row — no traversal expression"]

    class C,F,S,T,R domain
    class Y ok
    class N1,N2,N3 bad
```

`TraversalProcessor` asks the SPI for a generator that supports the component's type; `hkj-processor-plugins` ships `ListGenerator`, `SetGenerator`, `MapValueGenerator`, `ArrayGenerator` and `OptionalGenerator`, and none of them claims `java.util.Collection`. No generator means no method, and the processor emits no diagnostic for it.

The asymmetry is new. Before #725, `Collection` did not really work under `@GenerateFocus` either — the generated path threw `ClassCastException` on its first traversal. #725 made it a first-class Focus container with `Traversals.forCollection()` behind it, which leaves `@GenerateTraversals` as the annotation that now visibly cannot see a container its sibling handles.

## Two things to fix

**The silence.** A component the processor cannot generate for should say so, whatever the container. That half stands on its own, and is worth doing even if `Collection` support is declined: today the only signal is a missing method.

**The gap.** `Collection` needs a `CollectionGenerator` in `hkj-processor-plugins` (`@ServiceProvider(TraversableGenerator.class)`, a `provides` line in `module-info`, `ZERO_OR_MORE`, `generateOpticExpression()` returning `EachInstances.collectionEach()`), plus a row in `TypeKindAnalyser` so `@ThroughField` auto-detect stops rejecting it, and one in `TraversalCodeGenerator` so `@ImportOptics` and `@OpticsSpec` can name it.

The rebuild policy is already decided and must not be re-derived: `generateModifyF` should call `Traversals.forCollection()`, which rebuilds a set source as a set and everything else as a `List`. #725 collapsed four disagreeing set-rebuild policies into one for exactly this reason — a new generator that writes its own rebuild into generated source would reopen that.

## Notes

- Found by the review panel on #754 (the #725 fix), and deliberately left out of it: adding an SPI provider is a different blast radius from correcting an expression, and `@GenerateTraversals` was not what #725 reported.
- `Collection` carries two limits that `Traversals.forCollection()`'s javadoc already states and any new generator inherits: a `SortedSet` source keeps its elements but not its comparator, and a source that is neither `List` nor `Set` — an `ArrayDeque`, a `PriorityQueue` — comes back a `List`, so the traversal identity law does not hold for it. Those types inherit identity `equals`, and no rebuild could satisfy it.

MagnusSmith:
Closes #756.

`@GenerateTraversals` asked the `TraversableGenerator` SPI for each record component, and no generator claimed `java.util.Collection`, so `record Job(Collection<String> tags, List<String> notes)` generated `notes()` and nothing else — no error, no warning. `@GenerateFocus` has recognised `Collection` since #725, which left the two annotations disagreeing about a container. This closes both halves the issue names: the gap and the silence.

## The gap

```mermaid
flowchart LR
    GT["@GenerateTraversals<br/>CollectionGenerator"] --> TC
    GF["@GenerateFocus<br/>EachInstances.collectionEach()"] --> FC["Traversals.forCollection()"]
    IO["@ImportOptics / @ThroughField<br/>TraversalCodeGenerator COLLECTION row"] --> FC
    FC --> TC["Traversals.traverseCollection<br/>set source → unmodifiable set<br/>anything else → unmodifiable list"]

    classDef policy fill:#a6d189,stroke:#40a02b,color:#232634
    class GT,GF,IO,FC domain
    class TC policy
```

- **`Traversals.traverseCollection`** (hkj-core, `@since 0.4.10`) is the one rebuild policy; `forCollection()` now delegates to it. Generated source calls the helper rather than `forCollection()` for the same reason `ListGenerator` calls `traverseList`: a `Collection<? extends Leaf>` component cannot be typed through `Traversals.<A>forCollection()`, and the wildcard shapes #721 made work for `List` had to work for `Collection` too. The `instanceof Set` decision lives in exactly one place.
- **`CollectionGenerator`** in `hkj-processor-plugins` (`@ServiceProvider`, `provides` line, `ZERO_OR_MORE`, `EachInstances.collectionEach()`), mirroring `SetGenerator` line for line.
- **`ContainerType.Kind.COLLECTION`** rows in `TypeKindAnalyser` and `TraversalCodeGenerator`, so `@ThroughField` auto-detects a `Collection` field and `@ImportOptics` names `Traversals.forCollection()` for it. The `TypeKindAnalyser` match is **exact** (the interface itself, not `Deque`/`Queue`): the rebuilt list could not be handed back to a `Deque`-typed field through the builder setter.

## The silence

`TraversalProcessor` now reports, in the house what/why/fix format and attached to the component:

```
Note: @GenerateTraversals: no traversal was generated for component 'Job.tasks' of type Deque<String>. No TraversableGenerator on the annotation processor path supports Deque. Declare the component as a supported container (List, Set, Collection, Map, Optional or an array), or put a TraversableGenerator for its type on the annotation processor path.
```

The bar for "holds elements" is a **`java.util.Collection` or `java.util.Map` by erasure**. A bare `Iterable` is deliberately not the bar — `java.nio.file.Path` implements it — so a `String`, an `int`, a `Path` or a `Supplier<T>` component is passed over silently as before. The three ways a *claimed* container can still produce no method (raw type, focus index past the type's arguments, non-declared type) draw the same note with their own second sentence; those were silent `return null`s before.

It is a **note**, not a warning, on purpose (`Diagnostics.note` is new alongside `error`/`warning`): the record and its generated class are sound without the method, `@GenerateTraversals` has no per-component opt-out, and a `Messager` warning cannot be suppressed — so a warning would have failed every `-Werror` build with no remedy short of changing the record. No build that passed before fails now.

## Deliberate exclusions

- `@ImportOptics`' implicit `xTraversal()` route (`ExternalLensGenerator`) has the same `generator == null` silence and was **not** given the warning: it imports types the author cannot edit, so "declare the component as a supported container" is not a remedy there, and the explicit `@ThroughField` route already errors.
- `TraversalProcessor`'s two free-text `error(...)` calls were left as they are; their wording is quoted in `compiler_errors.md`.

## Surfaced, not fixed (follow-up candidates)

1. **#773** — `TypeKindAnalyser.detectContainerTypeWithSubtypes` matches `List`/`Set` **subtypes** (`ArrayList`, `TreeSet`), but `forList()`/`forSet()` hand back unmodifiable JDK collections through the raw `(Traversal)` cast in `TraversalCodeGenerator`, so an `ArrayList`-typed field auto-detected as `LIST` gets a `ClassCastException` in the builder setter on first use — `getAll` as well as `modify`, reproduced at runtime. The new `Collection` row is exact precisely for this reason, and now sits beside two rows that are not. Same mechanism as the runtime unsoundness #725 fixed for Focus.
2. **#774** — all three SPI lookup loops (`TraversalProcessor.generatorFor`, `WideningAnalysis.findSpiGenerator`, `ExternalLensGenerator`) are first-match in `ServiceLoader` order and ignore `priority()`; this change extracts the loop verbatim and does not worsen it.

## Verification

- `:hkj-core:check`, `:hkj-processor-plugins:check`, `:hkj-processor:check` (JaCoCo verification + ArchUnit rules) green; `Traversals` at 100% line/branch.
- `assemble testClasses` across all modules under `-Werror` green.
- `:hkj-examples:bookVerify` green with a new `<!-- verify -->` snippet (`record Crew(String name, Collection<Player> members)`) compiled through the real processor with warnings-as-defects; ratchet 315 → 316.
- `book-heading-html-check` clean; `spotlessCheck` clean.
- New tests: `traverseCollection` (Id/Optional applicatives, empty source of both shapes, set/list/deque rebuild); `CollectionGeneratorTest`; SPI method pins; `Collection<Leaf>`, `Collection<? extends Leaf>`, `Collection<?>`, `Collection<? super Leaf>`, raw `Collection`, `Holder<T>(Collection<T>)`, `Holder<F>(Collection<F>)` in the compile matrix (plus the `Set` generic rows that were missing); a behaviour test proving a `Collection<? extends Leaf>` component comes back a `Set` from a set and a `List` from a list; processor tests for the Deque/SortedMap/raw/no-type-parameter/focus-index/type-variable notes (and zero warnings), the silent non-container case (`String`, `int`, `Path`, `Supplier`), and the note's file/line position; `TypeKindAnalyser` exact-match and `@ThroughField` auto-detect for `Collection` with `Deque` rejected.
- Five-lens panel (correctness, type safety, immutability, ergonomics/docs, architecture) run after implementation; every should-fix was applied except the two exclusions above, which are recorded here by decision.
- Post-review: rebased on `main`; the diagnostic downgraded from warning to note (above); the unreachable `getTypeElement("java.util.Collection") != null` arm dropped for patch coverage; CodeRabbit's two comments addressed (`instanceof Set<?>` — the parameterised form was legal since Java 16 as a checked cast, but the wildcard says the same thing without the debate; `llms.txt` line 80 count).


MagnusSmith:
@coderabbit full review

PR #778, opened 2026-08-30
@ThroughField: auto-detect the container interfaces, and explain a narrower one (#773)

### LINKED ISSUE #773: [BUG] @ThroughField auto-detects an ArrayList or TreeSet field and the traversal throws ClassCastException on first use

`@ThroughField` auto-detects a field declared as a `List` or `Set` **subtype** — `ArrayList<String>`, `TreeSet<String>` — and wires it to `Traversals.forList()` / `forSet()`. Those traversals hand back an unmodifiable JDK collection, and the generated builder setter casts it back to the declared subtype. The spec compiles cleanly and the traversal throws `ClassCastException` on first use, read or write.

## Measured

On `main` after #772, through the real `ImportOpticsProcessor` and a runtime load of what it generated:

```java
public record Container(ArrayList<String> items) {
    public Container.Builder toBuilder() { return new Builder().items(items); }
    public static class Builder {
        private ArrayList<String> items;
        public Builder items(ArrayList<String> i) { this.items = i; return this; }
        public Container build() { return new Container(items); }
    }
}

@ImportOptics
public interface ContainerSpec extends OpticsSpec<Container> {
    @ViaBuilder Lens<Container, ArrayList<String>> items();
    @ThroughField(field = "items") Traversal<Container, String> eachItem();
}
```

```
Traversals.getAll(Container.eachItem(), new Container(new ArrayList<>(List.of("a", "b"))))
```

```
java.lang.ClassCastException: class java.util.Collections$UnmodifiableRandomAccessList cannot be cast to class java.util.ArrayList
	at org.higherkindedj.optics.Lens$7.set(Lens.java:368)
	at org.higherkindedj.optics.Lens$6.lambda$modifyF$0(Lens.java:345)
	at org.higherkindedj.hkt.id.Id.map(Id.java:48)
	at org.higherkindedj.optics.Lens$6.modifyF(Lens.java:345)
	at org.higherkindedj.optics.util.Traversals.getAll(Traversals.java:84)
```

The same for a `TreeSet<String>` field: `Collections$UnmodifiableSet cannot be cast to class java.util.TreeSet`. Note it is `getAll` that throws, not only `modify`: `getAll` runs the traversal through the `Id` applicative (`Traversals.java:82-90`), so the lens setter executes on the read path too. No `ArrayList`- or `TreeSet`-typed `@ThroughField` traversal has ever worked.

## Why

```mermaid
flowchart LR
    F["field: ArrayList&lt;String&gt;"] --> D["TypeKindAnalyser<br/>detectContainerTypeWithSubtypes<br/>isSubtype(erased, List) → LIST"]
    D --> G["TraversalCodeGenerator<br/>items().andThen((Traversal) Traversals.forList())"]
    G --> T["forList() rebuilds<br/>Collections.unmodifiableList"]
    T --> S["@ViaBuilder setter<br/>toBuilder().items((ArrayList) newValue)"]
    S --> X["ClassCastException"]

    class F,D,G,T,S domain
    class X bad
```

Three things line up:

1. `TypeKindAnalyser.detectContainerTypeWithSubtypes` matches `List` and `Set` by `isSubtype` (`TypeKindAnalyser.java:411, 423`), and `Map` the same way (`:461`). Its javadoc advertises it: "`ArrayList<String>` is detected as `List<String>`".
2. `TraversalCodeGenerator.generateThroughFieldCode` composes through a **raw** `(Traversal)` cast (`TraversalCodeGenerator.java:118-124`). Its javadoc says the cast is "safe at runtime because subtypes can be traversed using the supertype's traversal" — true of the read half of a traversal in isolation, false of the write half, and the write half runs on every use.
3. `Traversals.forList()` / `forSet()` hand back an unmodifiable list / set by contract (#725 made the set side deliberate), never the source's runtime class.

`Map` is the same shape with one accident: `traverseMapValues` rebuilds into a `new HashMap`, so a `HashMap<K, V>` field survives and a `TreeMap<K, V>` or `LinkedHashMap<K, V>` field throws.

The existing tests do not reach it. `ThroughFieldAutoDetectTest.shouldAutoDetectArrayListAsList` (`:86`), `shouldAutoDetectHashSetAsSet` (`:193`) and `shouldAutoDetectHashMapAsMap` (`:462`) assert only that the generated source *contains* `Traversals.forList()` etc. and compiles — the unsound case is pinned as passing.

#772 added the `Collection` row to the same method as an **exact** match, with the comment "a field declared as a `Deque` or a `Queue` would be rebuilt as a value it cannot hold". That reasoning applies verbatim to the two rows above it; the file now states a rule it breaks two blocks earlier.

## What to fix

Two defensible directions; the first is the smaller change and matches what `Optional` and `Collection` already do in the same method.

**Reject at the declaration.** Match `List`, `Set` and `Map` exactly, as `Optional` and `Collection` are matched. A subtype-typed field then falls into the existing "Cannot auto-detect traversal for field 'x' of type 'ArrayList<String>'" error (`SpecInterfaceAnalyser.java:1199-1216`), which already tells the author to name a `traversal = "..."` explicitly — and an explicit `Traversal<Container, String>` over an `ArrayList` field is theirs to write correctly. The javadoc's "including subtypes" list and the three tests above change with it. Every spec that hits this today throws at runtime, so nothing that worked stops working.

**Rebuild the declared type.** Keep subtype detection and have the generated traversal rebuild into the field's runtime class (`ArrayList::new`, `TreeSet::new` via `EachInstances.fromIterableCollecting`). That reopens the question #725 closed — one rebuild policy per container, held in `Traversals` — and cannot be made total (an `ArrayList` subclass with no copy constructor), so it should not be taken without a reason the first direction lacks.

Whichever is chosen, the `TraversalCodeGenerator` javadoc claim about the raw cast should be corrected, and one runtime test should pin the behaviour rather than the generated text — `RuntimeCompilationHelper.compile` already loads what the processor emits.

## Notes

- Found by the review panel on #772 (the #756 fix), and deliberately left out of it: #756 was about `@GenerateTraversals`, and a change to how `@ThroughField` treats subtypes is a behaviour change to a different annotation.
- Reproduced against `main` after #772 with a throwaway runtime test (not committed); the trace above is from that run.
- The `Collection` row in the same method is exact for exactly this reason and must stay exact whichever direction is taken here.

MagnusSmith:
Closes #773.

`@ThroughField` auto-detection matched a field declared as a `List`/`Set`/`Map` *subtype* and composed the interface traversal into a lens over the concrete type through a raw `(Traversal)` cast. `Traversals.forList()` hands back an unmodifiable `List`, the builder setter casts it to `ArrayList`, and the generated traversal threw `ClassCastException` on first use, on `getAll` as well as `modify` (reproduced at runtime in the issue). The book's own table advertised the shape as supported.

## What the annotation now does

```mermaid
flowchart LR
    F["@ThroughField(field = x)"] --> L{"spec declares<br/>Lens x()?"}
    L -->|no| R1["refused: composes through a lens<br/>the spec does not declare"]
    L -->|yes| D{"lens focus is<br/>List / Set / Collection /<br/>Map / Optional / T[]?"}
    D -->|yes| G["compose the standard traversal"]
    D -->|"narrower: ArrayList, TreeSet,<br/>TreeMap, Deque, int[]"| R2["refused: names the interface<br/>and both remedies"]
    D -->|"anything else"| R3["cannot auto-detect:<br/>name a traversal"]

    class F domain
    class L,D decision
    class G good
    class R1,R2,R3 bad
```

- **Exact match on the interface.** `TypeKindAnalyser.detectContainerTypeWithSubtypes` is gone; `autoDetectTraversalForField` uses the exact `detectContainerType`, as `Optional` and `Collection` already were. A field whose type is narrower than the interface (a concrete container, or another interface such as `Deque`/`SortedSet`) draws a what/why/fix error naming the interface it implements and both remedies: a traversal built with `Traversals.forIterableCollecting` / `forMapValuesCollecting`, or, where the type is yours, the interface. An array of a primitive is refused on the same ground (`forArray()` walks an `Object[]`, and an `int[]` is not one; the reviewer reproduced that one too).
- **Detection reads the lens the composition is typed against.** The generated traversal is `Spec.x().andThen(...)`, the spec's own lens for the field, so the type detected from is that lens's focus (read under the spec's instantiation), not the accessor's return type. An accessor returning `ArrayList` behind a `Lens<S, List<String>>` is sound and is accepted. A spec declaring no lens for the field is refused at the declaration; it could previously only fail inside the generated file, or, when the traversal method shared the field's name, compile into a method that called itself and overflowed the stack on first use (four of our own tests were that shape).
- **The raw cast's javadoc says what it now rests on**: the two types agree at erasure because auto-detection accepts the interface types alone; an explicit `traversal` is the author's undertaking. The stale "safe at runtime because subtypes can be traversed" comments in `TraversalCodeGenerator` and `SpecInterfaceGenerator` are gone.
- **Docs**: the `@ThroughField` annotation javadoc (IDE hover text), `OrderRecordOpticsSpec`'s example javadoc, `copy_strategies.md` (table + paragraph), `traversals.md`, two new `compiler_errors.md` entries, and the 0.4.11-SNAPSHOT release note.

## Tests

- Runtime, through the real processor: a `List<String>` field reads and writes, the modified record is a new one with the source untouched and the field holding the unmodifiable `List` the traversal handed back; the interface traversal named explicitly over an `ArrayList` field throws `ClassCastException` (the premise, pinned); a traversal built with `forIterableCollecting(ArrayList::new)` and named explicitly round-trips (the remedy, exercised).
- Refusals pinned verbatim with `hadErrorCount(1)`: `ArrayList`, `HashSet`, `HashMap`, `Deque` (names `Collection`), `int[]`, a non-generic `class Tags extends ArrayList<String>` (names `List`), a spec with no lens for the field; raw `ArrayList` and a type-variable lens focus fall to the generic "Cannot auto-detect" message.
- `@ThroughField` auto-detects from the declared lens focus (asymmetric accessor/builder shape); `TypeKindAnalyserTest` loses the subtype class (22 compilations) and gains exact-match rows for `ArrayList` and the raw interfaces.

## Verification

`:hkj-processor:check` (100% line/branch/instruction on `SpecInterfaceAnalyser*`, ArchUnit rules), `:hkj-annotations:javadoc`, `:hkj-examples:compileJava`/`javadoc`, full-repo `assemble testClasses` under `-Werror` (none of the 23 in-repo `@ThroughField` uses is affected), `bookVerify`, heading check, spotless. Five-lens panel run after implementation; blockers (the annotation javadoc still promising subtype detection; a dead `isSameType` arm) and every should-fix applied.

## Follow-ups surfaced, not taken here

- **#779** — the composition still goes through a raw cast on the auto-detected path, where the two types are now known to agree. Emitting it **uncast** there (keeping the cast for the explicit-`traversal` string) would let javac check the traversal's *focus* against the container's element type as well (`Traversal<S, String>` over a `List<Integer>` field still compiles and throws today). Needs a flag on `TraversalHintInfo` threaded to the generator and the suppression decision; worth its own issue under the type-safety initiative.
- `@GenerateTraversals` reports an `ArrayList` component with a *note* while `@ThroughField` now errors; justified (no per-component opt-out versus an explicit per-method request) and now stated on the traversals page.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `@ThroughField` traversal auto-detection to use the field’s declared interface type.
  * Prevented unsupported concrete containers, narrower interfaces, primitive arrays, and missing lenses from generating unsafe traversals.
  * Added clearer compiler errors with guidance for specifying an explicit traversal or supported interface.

* **Documentation**
  * Updated guides, examples, and release notes to document supported container types, restrictions, and migration guidance.

* **Tests**
  * Expanded coverage for valid interfaces, rejected declarations, runtime behavior, lens detection, and explicit collecting traversals.


MagnusSmith:
@coderabbit full review

Issue #752, opened 2026-08-27
[FEAT] @GenerateMapping: resolve a generic mix-in's members under the spec's instantiation


MagnusSmith:
Implements the decision recorded on #740: a generic mix-in's members are read under the spec's instantiation rather than refused at the declaration.

`MappingProcessor.checkMixins` currently rejects any ancestor interface that has type parameters and carries vocabulary — a `default` method, a `@MapField`, or an abstract leaf — with *"a generic mix-in's members would need substitution under its instantiation, which is not supported yet."* That is accurate about the code today, and it refuses the shape a mix-in exists for:

```java
interface Emails<T> { default ValidatedPrism<String, T> email() { ... } }
interface UserEmails extends Emails<EmailAddress> {}

@GenerateMapping
interface UserMapping extends MappingSpec<User, UserDto>, UserEmails {}
```

`asMemberOf(UserMapping, email)` answers `ValidatedPrism<String, EmailAddress>`, fully ground. There is correct source to emit, so by the rule #721 and #747 established, this resolves rather than refuses.

```mermaid
flowchart TD
    A["spec member, inherited"] --> B{"does the spec's type<br/>carry an instantiation?"}
    B -->|"no — raw supertype"| R["<b>refuse</b><br/>asMemberOf erases every<br/>member (the #738 regression)"]
    B -->|"yes"| C["asMemberOf(spec.asType(), member)"]
    C --> D{"does a mix-in parameter<br/>the route leaves unbound<br/>survive?"}
    D -->|"yes"| U["<b>refuse</b><br/>nothing pins it (as #742)"]
    D -->|"no"| E["<b>resolve</b><br/>ValidatedPrism&lt;String, EmailAddress&gt;<br/>or the spec's own T, which the Impl declares"]

    classDef bad fill:#e78284,stroke:#d20f39,color:#232634
    classDef ok fill:#a6d189,stroke:#40a02b,color:#232634
    classDef decision fill:#e5c890,stroke:#df8e1d,color:#232634
    class B,D decision
    class R,U bad
    class E ok
```

## The work, in three steps

The risk here is not the substitution, it is **missing a read site**: a member type read as declared and emitted into the Impl puts a free `T` in generated code, which is the #746 bug class. So the routing is separated from the behaviour change, and completeness is checked rather than assumed.

**1. One helper, every read routed through it.** No behaviour change intended.

```java
private TypeMirror specMemberType(TypeElement spec, ExecutableElement member)
```

[...]

PR #780, opened 2026-08-30
@GenerateMapping: hold spec members to what the Impl can carry (#762)

### LINKED ISSUE #762: [BUG] @GenerateMapping emits spec member types the generated Impl cannot hold
`@GenerateMapping` emits a spec member's type into the generated Impl without asking whether the Impl can *hold* it. Three shapes get through, and each fails as a bare javac error inside `<Spec>Impl.java` with no diagnostic from the processor at all. All three reproduce on `main` at `38dc8be05`, and none is generics-specific: each has a non-generic form, which is why they survived the generic-mix-in work in #752.

```mermaid
flowchart LR
    A["spec member"] --> B["read under the spec<br/>(#752)"]
    B --> C{"can the Impl<br/>hold this type?"}
    C -->|"not asked"| D["emitted verbatim"]
    D --> X["javac, inside<br/>the generated file"]
    C -.->|"the check<br/>that is missing"| R["refuse at the<br/>declaration"]

    class C decision
    class X bad
    class R ok
```

## 1. Two mix-ins agreeing on a rename, with covariant return types

`addRenameStubs` dedupes by simple name and emits the first member's return type. Override-equivalent abstracts may legally coexist (JLS 9.4.1), and nothing checks that one stub satisfies both.

```java
interface AV1 { @MapField(to = "label") String name(); }
interface AV2 { @MapField(to = "label") CharSequence name(); }

@GenerateMapping
public interface ASpec extends MappingSpec<Dom, AWire>, AV1, AV2 {}
```
```
error: com.example.ASpecImpl is not abstract and does not override abstract method name() in com.example.AV1
error: name() in com.example.ASpecImpl cannot implement name() in com.example.AV1
  return type java.lang.CharSequence is not compatible with java.lang.String
```

Order-independent, and identical with the two returns supplied by generic mix-ins at different instantiations. The narrowest return that satisfies every declaration is the one to emit; where none exists, that is the refusal.

## 2. A member type the generated package cannot name

```java
// package other
record Secret(String value) {}            // package-private
public interface BVocab { @MapField(to = "label") Secret name(); }

// package com.example
@GenerateMapping
public interface BSpec extends MappingSpec<Dom, BWire>, other.BVocab {}
```
```
error: other.Secret is not public in other; cannot be accessed from outside package   (x3)
```

The spec is legal Java, and the same spec placed in package `other` generates correctly. A `private` nested target gives the sibling message. `ProcessorUtils.firstUnreachableIn` already answers exactly this question and is used by `@GeneratePathBridge` and `@GenerateIsos`; `@GenerateMapping` does not call it.

## 3. A generic *method* leaf writes a variable the Impl never declares

```java
@GenerateMapping
public interface CSpec<T> extends MappingSpec<CPage<T>, CDto<T>> {
    <R> ValidatedPrism<R, R> items();
}
```
→ `cannot find symbol: class R, location: class com.example.CSpecImpl<T>`, seven times.

`abstractLeaves` collects any abstract leaf and `elementMappedSkeleton` emits a field per leaf; a method's *own* type parameters are not substituted by `asMemberOf` and have nowhere to go in a field. This is the #746 family exactly: a generated file naming a variable nothing brings into scope. The `of(...)` factory could declare it, but the field it initialises cannot, so the conforming answer is to refuse at the declaration, as `@GenerateIsos` does for the same reason.

A `@MapField` rename on a generic method (`<R> R name()`) fails the same way through `addRenameStubs`.

## Why they are one issue

All three are the same missing question, asked at the two emission sites (`elementMappedSkeleton`'s leaf fields, `addRenameStubs`' rename stubs): *is the type I am about to write down one this file can hold?* Answering it once covers all three, and the pieces exist — `ProcessorUtils.firstUnreachableIn` for (2), `method.getTypeParameters().isEmpty()` for (3), and a narrowest-return choice for (1).

Each wants a case in `GenericMappingWireAxisTest`, which since #752 compiles generated source under the consuming build's `-Xlint:unchecked,rawtypes -Werror`; a plain `succeeded()` would not have caught (2) or (3), and did not.

## Provenance

Found by the five-lens review panel on #752 and each reproduced here against `main` rather than taken from the report. They were deliberately left out of that PR: none is a regression from it, and each has a non-generic form that predates it.

MagnusSmith:
Closes #762.

`@GenerateMapping` emitted a spec member's type into the generated Impl without asking whether the Impl can hold it. Three shapes got through, each failing as a bare javac error inside `<Spec>Impl.java` with no diagnostic from the processor; none is generics-specific, which is how they survived #752.

## The one question, asked once

The Impl carries a leaf as a constructor-supplied field and a rename as a stub, so every member it re-emits has to answer: *is this a type the generated file can hold?* The refusals live in `validateSpecMethods`, so every tier a spec passes (full mapping, projection, `UpdateSpec`) asks them once, and each fires exactly one diagnostic:

- **A member declaring type parameters of its own** (`<R> ValidatedPrism<R, R> items()`, or a `@MapField` rename `<R> R id()`): a field and a stub have nowhere to declare `<R>`, so the generated file named a variable nothing brings into scope (`cannot find symbol: class R`, seven times in the issue's reproduction). Refused at the declaration; the fix line branches by kind, since a rename's remedy is a concrete return type while a leaf's is element types on the interface's own parameters.
- **A member type the spec's package cannot see**: the Impl is a top-level class in the spec's package that writes the type out in full. Two routes get one past the spec's own compile — a package-private type handed over by a mix-in from another package, and a `private` type nested in the spec's enclosing class, which the flattened Impl cannot see. Both are refused naming the type and the package (`checkMemberTypeReachable`, via `ProcessorUtils.firstUnreachableIn` — the same check `@GenerateIsos` and `@GeneratePathBridge` already make), for renames and abstract leaves, the only member types the Impl writes out.
- **Same-named members with covariantly differing returns**: override-equivalent abstracts may legally coexist (JLS 9.4.1.3), and both `addRenameStubs` and `abstractLeaves` picked whichever member the scan reached first. Both now pick the **subtype-narrowest** of the group (one shared `narrowestMember`), which is the only declared return satisfying the rest — for renames (`String id()` beside `CharSequence id()`) and for wildcard-differing leaves (`ValidatedPrism<TDto, T>` beside `ValidatedPrism<TDto, ? extends T>`) alike. One case the language admits through unchecked conversion has no narrowest — a raw return beside incomparable parameterised ones — and is refused naming every declaration (`checkGroupsHaveNarrowestReturns`), rather than emitting a member javac rejects in the generated file. And a name shared by a rename and a leaf gets the leaf accessor alone: the guard has proven its return satisfies the rename declaration, so a stub as well would declare the method twice (CodeRabbit's catch).

## Tests

Ten cases in `GenericMappingWireAxisTest`, compiled (where anything generates) under the consuming build's `-Xlint:unchecked,rawtypes -Werror` (the #752 discipline; a plain `succeeded()` did not catch two of the original shapes). The covariant case is **order-proof**: `getAllMembers` order is a javac internal, so the pair is mirrored across two specs (whichever member is seen first, one spec needs the replace arm and the other the keep arm) plus a third spec supplying the returns through generic mix-ins at different instantiations, so the chooser reads substituted types. Refusals are pinned verbatim with `hadErrorCount(1)`: the unreachable rename and leaf, the private-nested type, the generic leaf and rename, and the no-narrowest group (its declaration list sorted for a stable message).

## Docs

The limits table in [Diagnostics and limits](https://higher-kinded-j.github.io/latest/mapping/testing.html#diagnostics-and-limits) gains both rows; the shared-vocabulary page carries the reachability rule and the narrowest fold; the element-mapped section carries the generic-method rule; the renames section on the basics page carries the stub-return sentence; the `hkj-mapping` skill is updated (including two claims stale since #752: generic mix-ins are legal unless reached raw). Release note under 0.4.11-SNAPSHOT.

## Verification

`:hkj-processor:check` (`MappingProcessor*` pinned at 100% line/branch/instruction, ArchUnit rules), full-repo `assemble testClasses` under `-Werror`, `bookVerify`, heading check, spotless. Five-lens panel run after implementation; the one blocker (the order-dependent covariant test) and every should-fix applied.

## Surfaced by the panel, not taken here (follow-up candidates)

- **`@GenerateErrorEnvelope`** writes the context record's component types into a companion generated in the *annotated interface's* package and never asks reachability — the same question, a different processor. **`@GenerateAssembly`** has the narrow private-nested variant of the same gap. One follow-up issue would cover both.
- The nested-use-site emission `PageMappingImpl.<Secret>instance()` can name a binding type the *using* spec's package cannot see (component-derived rather than a spec member).
- `@GenerateMerge` was swept and is clean by construction (no mix-ins, no generics, same-package emissions); derived fields and `default` leaves emit no types and need no check.

## Summary by CodeRabbit

- **New Features**
  - Same-named mapping members from mix-ins now resolve to a single generated member using the narrowest compatible return type.
  - Clear diagnostics identify conflicting declarations when no narrowest return type exists.

- **Bug Fixes**
  - Invalid mappings are now reported at their declarations when members declare their own type parameters or use types inaccessible from the generated implementation.
  - Improved handling of abstract leaves, renames, generic mix-ins, and covariant return types.

- **Documentation**
  - Updated mapping guides, diagnostics references, and release history to explain the new behavior and limitations.


MagnusSmith:
@coderabbit full review

PR #781, opened 2026-08-30
Resolve TraversableGenerator selection by priority everywhere (#774)

### LINKED ISSUE #774: [BUG] TraversableGenerator.priority() is never consulted: every SPI lookup takes the first match in ServiceLoader order

`TraversableGenerator.priority()` promises that when several generators support one type, "the highest-priority one wins" (`TraversableGenerator.java:41-45`), and offers `PRIORITY_OVERRIDE` for "explicit overrides of built-in generators" (`:39, :51`). No lookup honours it. All three places a generator is chosen take the **first** match in `ServiceLoader` order; one of them warns when a tie happens to be noticed, the other two say nothing.

## Measured

The three loops on `main`:

`WideningAnalysis.findSpiGenerator` (`WideningAnalysis.java:555-566`), javadoc "Finds the highest-priority SPI generator":

```java
for (TraversableGenerator generator : traversableGenerators) {
  if (generator.supports(type)) {
    if (matched != null && matched.priority() == generator.priority()) {
      warnOfConflict(type, component, matched, generator);
    } else if (matched == null) {
      matched = generator;
    }
  }
}
```

`matched` is assigned once and never replaced, so a later generator with a *higher* priority loses to an earlier one with a lower priority, and no warning is raised for that case — the warning fires only when the priorities are *equal*. A `PRIORITY_OVERRIDE` provider therefore wins or loses by where its `META-INF/services` entry lands relative to the built-in's, which for the shipped plugins is alphabetical by class name.

`TraversalProcessor.generatorFor` (`TraversalProcessor.java:313-321`) and `ExternalLensGenerator.createTraversalMethod` (`ExternalLensGenerator.java:316-326`) are plain first-match loops: `priority()` is never read, and no tie is reported.

`SpiGeneratorConflictTest` pins the equal-priority warning through `DupGeneratorAlpha` / `DupGeneratorBeta`, with `DupGeneratorFallback` (`priority() = -50`) registered *after* both — an order in which first-match and highest-priority agree, so the test cannot tell them apart. There is no test in which an override is registered after the default it overrides.

## Why

```mermaid
flowchart LR
    S["ServiceLoader order<br/>(META-INF/services, alphabetical for shipped plugins)"] --> W["WideningAnalysis.findSpiGenerator<br/>first match; warns on equal priority only"]
    S --> T["TraversalProcessor.generatorFor<br/>first match; no warning"]
    S --> E["ExternalLensGenerator.createTraversalMethod<br/>first match; no warning"]
    P["TraversableGenerator.priority()<br/>'the highest-priority one wins'"] -.->|"read by nothing that decides"| W

    class S,W,T,E domain
    class P bad
```

The contract was added with the cardinality work (release history, 0.4.0: "priority system (`PRIORITY_FALLBACK`, `PRIORITY_DEFAULT`, `PRIORITY_OVERRIDE`)") and the tie warning with it; the resolution step that would make the constants mean something was never written. Every shipped generator matches an exact FQN and no two claim the same type, so the shipped set never exposes it — it is a third-party `PRIORITY_OVERRIDE` for `java.util.List`, or a `PRIORITY_FALLBACK` catch-all, that does. A catch-all registered alphabetically before `ListGenerator` (say `com.acme.AnyIterableGenerator`) silently takes every `List` component in the project.

The three loops also disagree with each other: a type two generators claim can resolve to different generators under `@GenerateFocus` and `@GenerateTraversals` only if their loops differ, and today they differ in what they warn about, not in what they pick — but that is an accident of them all being first-match, not a property anything holds them to. #719 put every *widening* site behind `WideningAnalysis.wideningGenerator` so the tier, the navigator and the composition call cannot disagree; the generator *choice* has no such single home.

## What to fix

**One resolver, honouring the contract.** A single lookup — highest `priority()` wins; among equals, the first in `ServiceLoader` order, with the existing warning — that all three sites read from. `WideningAnalysis` is the natural home given #719, and `ProcessorArchitectureRules.spi_generator_lookups_should_go_through_the_widening_guard` already keys on `findSpiGenerator` by name, so routing the other two loops through it (or through a shared `GeneratorRegistry` both `WideningAnalysis` and the two processors hold) would put the rule in a position to enforce what it is named for. The two silent loops should warn on a tie the same way.

The resolver is a two-line change; the work is the test that pins it: an override registered *after* the default it overrides, and a fallback registered *before* it, asserting that priority wins in both orders and through all three entry points (`@GenerateFocus`, `@GenerateTraversals`, `@ImportOptics`). The test-scope `META-INF/services` file fixes the order, which is exactly what makes such a test possible.

**Or narrow the contract.** If priority is not wanted, delete the constants and the `priority()` default and say the first registered generator wins. Cheaper, but `PRIORITY_OVERRIDE` is the only documented way to replace a built-in generator without forking it, and the SPI has been public since 0.3.8.

## Notes

- Found by the review panel on #772, and deliberately left out of it: #772 extracted `TraversalProcessor`'s loop verbatim into `generatorFor` and did not change what it picks; a resolution change is a behaviour change to all three annotations.
- Not reproduced at runtime: the loops are deterministic and the shipped generators never collide, so a reproduction needs a test-scope override registered after a built-in, which is the test the fix should add.
- Related: #432 (SPI widening enhancements), which introduced the priority constants.

MagnusSmith:
Closes #774.

`TraversableGenerator.priority()` promised that when several generators support one type, the highest-priority one wins. No lookup honoured it: all three places a generator is chosen took the first match in `ServiceLoader` order. The Focus route got the right answer only because `FocusProcessor` pre-sorted its list, a precondition living in a constructor javadoc; `@GenerateTraversals` and `@ImportOptics` had neither the sort nor the tie warning. A `PRIORITY_OVERRIDE` provider therefore won or lost by where its `META-INF/services` entry landed.

## One registry, three routes

```mermaid
flowchart LR
    subgraph Routes
        direction TB
        T["@GenerateTraversals"]
        F["@GenerateFocus widening"]
        I["@ImportOptics"]
    end
    T --> R["GeneratorRegistry.generatorFor(type, component)"]
    F --> R
    I --> R
    R --> W["highest priority() wins,<br/>wherever the services entry lands"]
    R --> E["equal-priority tie: warning<br/>names both, first registered wins"]

    classDef warn fill:#e5c890,stroke:#df8e1d,color:#232634
    class T,F,I,R domain
    class W good
    class E warn
```

`GeneratorRegistry` sorts once at construction, priority descending with a stable sort so equal priorities keep registration order, and answers every choice. It is module-internal, in the `processing` package beside the analyses that consume it, so the exported plugin-author surface stays exactly `TraversableGenerator` and `Cardinality`. `WideningAnalysis.findSpiGenerator` delegates to it, so the #719 widening guard sits on top unchanged; `FocusProcessor`'s manual sort and the javadoc precondition are gone; `TraversalProcessor` and `ExternalLensGenerator` read the same answer, and both now anchor the equal-priority warning to the record component, where before they said nothing. The warning itself now carries the remedy: rank one of the providers with `priority()`, or drop one from the processor path.

Two ArchUnit rules hold the shape, each probe-verified to fail under the attack it forbids: only `GeneratorRegistry` may consult `TraversableGenerator.supports()` or `priority()`, matched over method calls *and* method references (the deleted pre-sort was `TraversableGenerator::priority`) with subtype-typed receivers included; and only the three route lookups may read `generatorFor`, so a site cannot slip past `wideningGenerator`'s turn-away of raw and wildcard-carrying containers (#718) now that `findSpiGenerator` is a delegate.

## Tests

The resolver is a small change; the work is the order the tests fix. The test-scope `META-INF/services` file now registers a `PRIORITY_OVERRIDE` generator *last* and a `PRIORITY_FALLBACK` generator *first*, the two arrangements a first-match loop gets wrong:

- `SpiPriorityResolutionTest` drives `@GenerateTraversals` and `@GenerateFocus` through the real `ServiceLoader`. Which generator won is observable twice over: every marker generator's `modifyF` now throws naming the generator, and the winners and losers declare opposite cardinalities, so the widened tier (`AffinePath` vs `TraversalPath`) names the winner too. The panel reverted the resolution in-tree to confirm what each case pins: the `@GenerateTraversals` and `@ImportOptics` cases fail under first-match, while the Focus cases pin that the route's already-correct behaviour survives the sort moving out of `FocusProcessor.init`. The tie case pins the warning and the first-registered winner on the previously silent `@GenerateTraversals` route.
- `ExternalGeneratorPriorityTest` pins the `@ImportOptics` route through the explicit-list constructor, since `TypeKindAnalyser` only flags standard containers and the shipped generators never collide there; the tie case pins the warning on this route as well.

## Docs

[How Plugin Discovery Works](https://higher-kinded-j.github.io/latest/tooling/generator_plugins.html#how-plugin-discovery-works) now describes priority resolution and the tie warning, the interface snippet shows `priority()`, and a tip covers replacing a built-in generator, which `PRIORITY_OVERRIDE` now actually does. The warning gets its entry in [Compiler Errors](https://higher-kinded-j.github.io/latest/optics/compiler_errors.html). Release note under 0.4.11-SNAPSHOT.

_Behaviour change:_ selection previously followed registration order on some routes, and only the Focus widening reported a tie. Two equal-priority providers claiming one type now draw the warning on every route, which a consuming build running javac with `-Werror` turns into an error; rank one of them, or drop one from the processor path. The shipped generators each claim distinct exact types, so builds using only shipped plugins see no difference.

## Verification

`:hkj-processor:check` (`GeneratorRegistry*` joins the JaCoCo classes pinned at 100% line, branch and instruction; ArchUnit rules), full-repo `assemble testClasses` under `-Werror`, `bookVerify`, heading check, spotless. Five-lens panel run after implementation; no blockers, and every should-fix applied: the two arch-rule escape hatches above, the registry moved out of the exported `spi` package, the remedy appended to the tie warning, and the test javadoc corrected to say which half is regression cover.

## Summary by CodeRabbit

* **New Features**
  * Generator selection now consistently honors priority across traversal generation, Focus widening, and imported optics.
  * Added support for override, default, and fallback generator priorities.
  * Equal-priority conflicts now produce a compile-time warning with deterministic selection.

* **Documentation**
  * Added guidance for resolving generator conflicts and replacing built-in generators.
  * Updated release notes and compiler-error references.

* **Tests**
  * Expanded coverage for priority resolution, conflict warnings, and imported optics.


MagnusSmith:
@coderabbit full review

PR #782, opened 2026-08-31
@ImportOptics: refuse a source type that names a raw type (#771)

### LINKED ISSUE #771: [BUG] @ImportOptics accepts a raw OpticsSpec source type and generates rawtypes-warning optics
`@ImportOptics` refuses a source type that is not a class — a type variable, an array — but not one that *is* a class and simply has no type arguments. `OpticsSpec<Box>` for a `Box<X>` is accepted, and every optic generated from it names `Box` raw. The file then fails the consuming build under its own flags, which is the one place a suppression cannot be written.

Reproduces on `main` at `330d1aa8d`.

## Where it slips past

```mermaid
flowchart TD
    A["OpticsSpec&lt;S&gt;"] --> B{"asElement(S)<br/>is a TypeElement?"}
    B -->|"no — type variable, array"| R["reportUnusableSourceType:<br/>'which is a type variable'"]
    B -->|"yes"| C{"does S supply<br/>its arguments?"}
    C -->|"Box&lt;String&gt;"| G["read under the instantiation"]
    C -->|"Box — raw"| X["accepted, and never asked:<br/>members erase, Box is named raw"]

    classDef gate fill:#e5c890,stroke:#df8e1d,color:#232634
    class B,C gate
    class X bad
    class R,G good
```

The second decision does not exist. A raw `Box` passes the first one — its element *is* a `TypeElement` — and `SpecInterfaceAnalyser` proceeds as though the source type were instantiated.

## Measured

```java
public class Box<X> {
  public Box(X content, String label, List<X> items) { … }
  public X content();
  public String label();
  public List<X> items();
  public Box<X> withLabel(String label);
}
```

```java
@ImportOptics
@SuppressWarnings("rawtypes")
public interface RawBoxOpticsSpec extends OpticsSpec<Box> {
  @Wither("withLabel")
  Lens<Box, String> label();

  @ViaConstructor(parameterOrder = {"content", "label", "items"})
  Lens<Box, List> items();
}
```

Accepted with no diagnostic, generating `RawBoxOptics`:

```java
public static Lens<Box, String> label() {
  return Lens.of(source -> source.label(), (source, newValue) -> source.withLabel(newValue));
}

public static Lens<Box, List> items() {
  return Lens.of(
    source -> source.items(),
    (source, newValue) -> new Box(source.content(), source.label(), newValue));
}
```

Compiled under the flags a consuming build uses:

```
RawBoxOptics.java:25: warning: [rawtypes] found raw type: Box
RawBoxOptics.java:36: warning: [rawtypes] found raw type: Box
RawBoxOptics.java:36: warning: [rawtypes] found raw type: List
RawBoxOptics.java:39: warning: [rawtypes] found raw type: Box
RawBoxOptics.java:39: warning: [unchecked] unchecked call to Box(X,String,List<X>) as a member of the raw type Box
```

Both strategies are affected, so this is the source type rather than any one of them. The last line is the erasure biting rather than merely a raw name: `constructorParameterType` reads the constructor under the raw source type, `asMemberOf` erases it, and the generated call is unchecked.

## Why

`SpecInterfaceAnalyser` gates the source type once:

```java
if (!(typeUtils.asElement(sourceType) instanceof TypeElement sourceTypeElement)) {
  reportUnusableSourceType(specInterface, sourceType);
  return Optional.empty();
}
```

That gate asks whether the type *has* an element, which is the right question for a type variable or an array and no question at all for a raw one. `reportUnusableSourceType`'s own comment is accurate about what reaches it — "a type variable and an array are the only two kinds" — precisely because raw never does.

Downstream, `constructorParameterType` calls `asMemberOf` with no guard, unlike its sibling `memberTypeOf` which asks `ProcessorUtils.carriesInstantiation` first. Under #740 those two were settled as deliberately different, and this issue is why the unguarded one is nonetheless safe to leave alone: the raw source type is what should not have been accepted, and guarding one member reader would bury the fault one level in while the other readers, and the raw `Box` in every generated signature, stayed wrong.

## Fix

A second check after the element gate, refusing a raw source type in its own right. `ProcessorUtils.isRaw` already asks exactly this and documents why raw is not the same as bare:

```java
if (sourceType instanceof DeclaredType declared && ProcessorUtils.isRaw(declared)) { … }
```

Not a third arm of `reportUnusableSourceType`: that one says the type cannot be an optic's source at all, while a raw `Box` names a perfectly good source and is missing its arguments. The remedy differs too — "name the type arguments: `OpticsSpec<Box<String>>`" — and where the spec is itself generic, `OpticsSpec<Box<U>>` is the answer, which the generic-spec path already supports.

`@GeneratePathBridge` refuses the same shape for the same reason, with wording worth matching (`PathProcessor.firstRawIn`, from #748):

> The bridge repeats it verbatim, and a raw type in generated source is a [rawtypes] warning that the suppression on your own declaration does not cover.

## Test

`GenericSpecInterfaceAxisTest` compiles on javac's defaults, where this passes `succeeded()` while failing the build that consumes the file. The row wants the flags `GenericMappingWireAxisTest` and `PathProcessorIntegrationTest` use — `-Xlint:unchecked,rawtypes -Werror` — which is also what would have caught it.

Worth a row per axis cell: source type × (instantiated / raw), against at least `@Wither` and `@ViaConstructor`, since a strategy that rebuilds through a method and one that rebuilds through a constructor fail differently — the second erases, the first only names raw.

## Notes

- The `@ImportOptics({Class...})` auto-detection path is **not** affected: it resolves a generic class properly, generating `static <T> Lens<GenBox<T>, List<T>> items()`. This is specific to a spec interface naming its source raw.
- Not #718, which is a wildcard argument inside an SPI container on `@GenerateFocus`.
- Found while settling the raw-site policies in #740; documented in place there rather than fixed, since the fix is this one.

MagnusSmith:
Closes #771.

`@ImportOptics` refused a source type that is not a class, but not one that is a class and simply has no type arguments. `OpticsSpec<Box>` for a `Box<X>` was accepted, and every optic generated from it named `Box` raw: a `[rawtypes]` warning in a file the author cannot edit, and under `@ViaConstructor` an `[unchecked]` call besides, since a constructor read under a raw type erases its parameters. The file failed the consuming build under its own flags.

## The second question, asked at the gate

```mermaid
flowchart TD
    A["OpticsSpec&lt;S&gt;"] --> B{"asElement(S)<br/>is a TypeElement?"}
    B -->|"no: type variable, array"| R1["refused:<br/>'which is a type variable'"]
    B -->|yes| C{"does S name a raw<br/>type anywhere?"}
    C -->|"Box&lt;String&gt;, Box&lt;U&gt;,<br/>Customer, Registry.Tag"| G["read under the instantiation"]
    C -->|"Box &#183; Outer.Holder &#183; Box&lt;List&gt;"| R2["refused: name the arguments,<br/>or pass the spec's own through"]

    classDef gate fill:#e5c890,stroke:#df8e1d,color:#232634
    class B,C gate
    class G good
    class R1,R2 bad
```

The panel turned the issue's check into a stronger one. The issue asked whether the source type itself is raw; the review reproduced the same `-Werror` failure through two shapes that predicate misses, so the gate asks the whole written type:

- **The source type itself** (`OpticsSpec<Box>`), the issue's case.
- **A member type behind a generic outer written bare** (`OpticsSpec<Outer.Holder>`): raw by JLS 4.8 even though `Holder` declares nothing of its own. A static nested type of a generic outer is not raw and stays accepted, pinned under `-Werror`.
- **A raw type argument** (`OpticsSpec<Box<List>>`): the generated file repeats it verbatim.

The predicate is `ProcessorUtils.firstRawIn`, **moved from `@GeneratePathBridge`** (#748), whose walk already covered arguments, wildcard bounds, arrays and the enclosing chain; `PathProcessor` now delegates, so the two processors that refuse raw shapes share one implementation. The diagnostic names the raw type it found and, where the source type itself is the raw one, shows the generic-spec form (`OpticsSpec<Box<U>>`). The bound hint on the type-variable refusal gained the same predicate: a hint naming a raw bound would steer straight into this refusal.

Per #740's settled policy, `constructorParameterType` stays unguarded: the raw source is what should not have been accepted, and its javadoc now records that the gate is what makes the unguarded read safe.

## What the change displaced

Two tests deliberately drove raw sources through the processor and pinned the old tolerance; both now pin the refusal. The `@InstanceOf` row was that route's only cover for the argument-count mismatch in `InstanceOfNarrowing.match`, so a new row covers the branch through the input that remains legal: a target reaching the source through a raw `extends` clause (`RawLeaf<Y> extends Node`), which pins nothing to `Y` exactly as before.

## Tests

- Refusals pinned verbatim with `hadErrorCount(1)`: the bare source (`@Wither`, and `@Wither` + `@ViaConstructor` proving one error at the declaration rather than one per member), the raw enclosing type, the raw type argument, and the raw-bound hint suppression.
- Acceptance rows under the consuming build's flags (`-Xlint:unchecked,rawtypes -Werror`, applied to the generated file in the same run): concrete `OpticsSpec<Box<String>>` with both strategies, generic `OpticsSpec<Box<U>>`, and static-nested `Registry.Tag`. The panel confirmed these pass pre-fix too; they are the ratchet the issue asked for, not regression cover.
- `SpecInterfaceAnalyser*`, `InstanceOfNarrowing*` and `ProcessorUtils*` hold their 100% line/branch/instruction pins.

## Docs

The [Compiler Errors](https://higher-kinded-j.github.io/latest/optics/compiler_errors.html) entry covers all three shapes and the raw-is-not-bare distinction; the type-variable sibling's fix line no longer suggests a shape this gate would refuse. Release note under 0.4.11-SNAPSHOT with the behaviour-change paragraph.

## Verification

`:hkj-processor:clean check` (coverage pins, ArchUnit rules), full-repo `assemble testClasses` under `-Werror`, `bookVerify`, heading check, spotless. Five-lens panel run after implementation; the raw-enclosing blocker (found independently by three lenses, probe-reproduced end to end) and every should-fix applied: the shared predicate, the stale `constructorParameterType` javadoc rewritten, the bound-hint guard, the nested-type rendering via `simpleTypeName`, and the release note's suppression claim corrected to "a file the spec author cannot edit".

## Summary by CodeRabbit

* **Bug Fixes**
  * Raw source types are now rejected early in optics specifications, including raw enclosing types and type arguments.
  * Compiler diagnostics now clearly identify the raw type and explain how to correct it.
  * Raw types are no longer suggested as fixes for generic type-variable errors.
  * Valid parameterized and static nested types continue to compile correctly under stricter compiler warnings.

* **Documentation**
  * Added guidance on raw generic types and updated release history notes.

### DIFF
--- a/hkj-book/src/optics/compiler_errors.md
+++ b/hkj-book/src/optics/compiler_errors.md
@@ -111,9 +111,15 @@ It is a note rather than a warning on purpose. `@GenerateTraversals` has no per-
 
 **Cause.** The spec interface is generic, and its own type parameter is the source type: `interface BoxOpticsSpec<S extends Box> extends OpticsSpec<S>`. Optics are generated against one named type, read for its members and rebuilt through its constructor, wither or setter, so a type parameter standing for whatever a caller picks has nothing to generate from. An array source type produces the same diagnostic with a different opening, `declares OpticsSpec<String[]>, which is an array type`, and the same remedy.
 
-**Fix.** Name the type the optics are for as the type argument: `OpticsSpec<Box>`. Where the bound names a single type, the message suggests it for you.
+**Fix.** Name the type the optics are for as the type argument, with its own type arguments where it has any: `OpticsSpec<Box>`. Where the bound names a single type that is not raw, the message suggests it for you.
 
 A source type that is itself generic is supported, and the spec names its own type parameters: `interface BoxOpticsSpec<U> extends OpticsSpec<Box<U>>` generates `static <U> Lens<Box<U>, String> label()`. See [Spec Interfaces](optics_spec_interfaces.md#generic-spec-interfaces) for which parameters a generated method declares. It is only a bare type variable, standing for the whole source type, that has no source to read.
+
+### "'XOpticsSpec' declares `OpticsSpec<Box>`, which names the raw type 'Box'"
+
+**Cause.** The source type names a generic type without its arguments. Every generated optic repeats the source type verbatim, so the generated file, which you cannot edit, would carry a `[rawtypes]` warning that the `@SuppressWarnings` on your own spec does not cover, and a `@ViaConstructor` rebuild read under a raw type erases its parameters into an `[unchecked]` call besides. Three shapes draw the error: the source type itself written bare (`OpticsSpec<Box>` for a `Box<X>`), a member type behind a generic outer written bare (`OpticsSpec<Outer.Holder>`, raw by JLS 4.8 even though `Holder` declares nothing of its own), and a raw type argument (`OpticsSpec<Box<List>>`). Raw is not the same as bare: a non-generic source type, or a static nested type of a generic outer, has no arguments to supply and is accepted as written.
+
+**Fix.** Name the raw type's arguments in the `OpticsSpec` clause: `OpticsSpec<Box<String>>`, `OpticsSpec<Outer<String>.Holder>`, `OpticsSpec<Box<List<String>>>`. A spec whose optics should stay generic declares its own type parameters and passes them on, `interface BoxOpticsSpec<U> extends OpticsSpec<Box<U>>`, as above. See [Spec Interfaces](optics_spec_interfaces.md#generic-spec-interfaces).
 
 ### "@ThroughField: '...' reaches field 'items', which is declared as `ArrayList<String>` rather than as the List interface"
 
--- a/hkj-book/src/release-history.md
+++ b/hkj-book/src/release-history.md
@@ -28,6 +28,10 @@ _In development. Release notes will be added here as changes land on `main`._
 
   _Behaviour change:_ selection previously followed registration order on some routes, so an override could win or lose by where its services entry landed, and only the Focus DSL's widening reported a tie. Two equal-priority providers claiming one type now draw the warning on every route, which a consuming build running javac with `-Werror` turns into an error; rank one of them, or drop one from the processor path.
 
+- **`@ImportOptics` asks a raw source type for its arguments** ([#771](https://github.com/higher-kinded-j/higher-kinded-j/issues/771)): a spec interface whose source type names a raw type anywhere, the type itself (`OpticsSpec<Box>` for a `Box<X>`), a member type behind a generic outer written bare (`OpticsSpec<Outer.Holder>`), or a raw type argument (`OpticsSpec<Box<List>>`), is refused at the declaration, pointing at the type to complete and at the generic-spec form (`OpticsSpec<Box<U>>`) for optics that should stay generic. See the entry in [Compiler Errors](optics/compiler_errors.md#xopticsspec-declares-opticsspecbox-which-names-the-raw-type-box).
+
+  _Behaviour change, moving a javac failure to the declaration:_ the shape previously generated a file naming the type raw throughout, which failed the consuming build under `-Xlint:rawtypes -Werror` in a file the spec author cannot edit; the spec now hears about it at its own declaration.
+
 ### [v0.4.10](https://github.com/higher-kinded-j/higher-kinded-j/releases/tag/v0.4.10) (30 August 2026)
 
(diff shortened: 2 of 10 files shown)


MagnusSmith:
@coderabbit full review

PR #783, opened 2026-09-01
Render diagnostic type names structurally, keeping type-use annotations out (#759)

### LINKED ISSUE #759: [BUG] A type-use annotation leaks into the type name a diagnostic prints
`ProcessorUtils.simpleTypeName` renders a `TypeMirror` for a diagnostic by stripping package prefixes off `type.toString()`. A type-use annotation is part of that string and survives, so it reads as though it were part of the type's name.

## Measured

```java
@GenerateFocus
public record Holder(@Nullable Set<?> f) {}
```

```
@GenerateFocus: record component 'Holder.f' has a wildcard type argument in @Nullable Set<?>.
The optic instance that widens that container is inferred from the field type, and a wildcard
has no ground instantiation to infer it from. Declare the component with concrete type
arguments, such as Set<Object>, or drop @GenerateFocus from the record: @GenerateLenses and
@GenerateTraversals compose no optic instance and take the component as written.
```

The message is internally inconsistent: it names the offending type as `@Nullable Set<?>` and then suggests `Set<Object>`, which silently drops the annotation the first half just printed. A reader following the suggestion literally would remove a nullness contract they never meant to touch.

## Why

```java
public static String simpleTypeName(TypeMirror type) {
  return type.toString().replaceAll("\\b(?:[a-z][\\p{Alnum}_]*\\.)+", "").replace(",", ", ");
}
```

`toString()` on an annotated `TypeMirror` yields `@org.jspecify.annotations.Nullable java.util.Set<?>`. The regex removes lowercase-leading dotted segments, which takes `org.jspecify.annotations.` and `java.util.` and leaves the `@Nullable` marker in place. `concreteAlternative`, which builds the suggestion, goes through `TypeElement.getSimpleName()` and the resolved type arguments instead, so it never sees the annotation — hence the mismatch between the two halves.

The formatter is shared, so this reaches every diagnostic that names a type through it, not only `@GenerateFocus`'s.

## Fix

Either drop leading type-use annotations in `simpleTypeName`, or build the name from the element and its resolved type arguments the way `concreteAlternative` already does, so the two halves of a message are produced the same way. The second is more work and removes the class of mismatch rather than this instance.

Whichever is chosen, the annotation should be dropped consistently: the diagnostic is about the *type argument*, and the annotation is not what makes the widening unwritable.

## Notes

- Pre-existing; it became easy to hit when #725 brought plain `java.util.Set` and `java.util.Collection` under the raw/wildcard rejection rule, which had previously only reached SPI containers.
- Related to #750 but a different mechanism and the opposite symptom. #750 is `TypeName.get(TypeMirror)` **dropping** type-use annotations on the way into generated *code*; this is `type.toString()` **keeping** one on the way into a *diagnostic message*. Fixing either does not fix the other.
- Found by the review panel on #754.

MagnusSmith:
Closes #759.

A diagnostic that named a type rendered it from the mirror's string form, so a type-use annotation survived into the message: a component declared `@Nullable Set<?>` was reported as having "a wildcard type argument in `@Nullable Set<?>`" beside the suggestion "such as `Set<Object>`", which reads as advice to drop a nullness contract the diagnostic was never about.

## One rendering for both halves

`ProcessorUtils.simpleTypeName` now builds the name the way the suggestion beside it always was built: structurally, so an annotation never enters.

- A declared type renders its enclosing chain (recursively where the enclosing type is carried, `Outer<String>.Holder`, and from the enclosing elements where the nesting is static, `Registry.Tag`), its simple name, and its resolved arguments — the same text the old package-stripping produced for every annotation-free type, which is why all forty call sites' pinned messages pass unchanged.
- Wildcards render as written (`?`, `? extends Number`, `? super Number`), arrays by component, type variables by name, and primitives by their kind, since `@Min(1) int` carries the annotation in its string form too.
- An unresolvable type and an intersection keep the string fallback, with annotations stripped, argument lists and all, plus the leading dot javac's annotated-type printing leaves behind. The fallback is where an error type renders best: its element walk would lose the qualifier on a nested `Missing.Inner` that the string keeps.

The class of mismatch goes rather than the instance, in both directions:

- `concreteAlternative`'s "such as" suggestions now carry the enclosing chain through the shared `declaredHead`, so a nested container's suggestion names a type that compiles; and its renderings of bounds and resolved wildcards pass through the same method, so `Set<? extends @Nully CharSequence>` suggests a clean `CharSequence`.
- The equal-priority tie warning from #774 concatenated the `TypeMirror` directly, the one message that stripped neither annotations nor packages; it now renders through the same method (the sweep this issue's review deferred to here).
- `MappingProcessor` carried a private renderer *shadowing* the shared name, whose non-declared arm had the same leak its own javadoc denied; it is renamed `parameterTypeName` and that arm delegates.

Generated source remains the opposite, deliberate decision: `typeNameOf` keeps type-use annotations (#750), and `simpleTypeName`'s javadoc now says so, so the two are not "aligned" by a future hand.

## Tests

- A unit contract table in `ProcessorUtilsTest` (the file's capturing-processor pattern): annotated top-level, argument, primitive, argument-carrying annotation and array component; both wildcard bounds; static and instance nesting; all eight primitive kinds; an intersection bound; a raw site; a type variable. The annotated rows pin #759; the rest are marked as characterising the rendering the old string form already produced.
- Unresolvable types in their own row, with the ERROR kind and the failed compilation pinned as premises: `Missing`, nested `Missing.Inner` (the qualifier kept), `Missing<String, Integer>` (the argument spacing exercised for real), and `@Ranged(from = 0) Missing` stripped clean.
- The issue's exact reproduction pinned in `FocusRecognisedContainerTest`: `@Nully Set<?>` reported as "a wildcard type argument in `Set<?>`" with the suggestion "such as `Set<Object>`" — both halves the same rendering.
- `ProcessorUtils*`, `GeneratorRegistry*` and `MappingProcessor*` hold their 100% line/branch/instruction pins; every verbatim diagnostic pin in the module passes unchanged.

## Docs

Release note under 0.4.11-SNAPSHOT. `simpleTypeName`'s javadoc records what renders structurally, what falls back, and the deliberate contrast with `typeNameOf`.

## Verification

`:hkj-processor:clean check` (coverage pins, ArchUnit rules), full-repo `assemble testClasses` under `-Werror`, `bookVerify`, heading check, spotless. Five-lens panel run after implementation; no blockers, and every should-fix applied: primitives moved out of the string fallback (probed: `@Size(max=5) int` had garbled to `(max=5) int` under the first cut's regex), the error-type routing corrected to keep nested qualifiers, the tie-warning and `MappingProcessor` sweeps, the `declaredHead` extraction, and the fallback's no-op remnants now genuinely exercised by the multi-argument unresolved row.

## Follow-up surfaced, not taken here

Around fifteen further sites concatenate a `TypeMirror` or qualified name directly into a message (`SpecInterfaceAnalyser`, `MappingProcessor`, `ErrorEnvelopeProcessor`); several deliberately print pasteable fully-qualified names, so each wants a per-site decision rather than a blanket sweep.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compile-time diagnostics so reported types consistently omit type-use annotations.
  * Type names now accurately display resolved element and argument types, including wildcards, primitives, arrays, and nested types.
  * Conflict warnings and suggested alternatives now use clearer, more consistent type names.

* **Tests**
  * Added coverage for annotated, wildcard, primitive, intersection, nested, and unresolved types in diagnostic output.

### DIFF
--- a/hkj-book/src/release-history.md
+++ b/hkj-book/src/release-history.md
@@ -32,6 +32,8 @@ _In development. Release notes will be added here as changes land on `main`._
 
   _Behaviour change, moving a javac failure to the declaration:_ the shape previously generated a file naming the type raw throughout, which failed the consuming build under `-Xlint:rawtypes -Werror` in a file the spec author cannot edit; the spec now hears about it at its own declaration.
 
+- **Diagnostics render a type without its type-use annotations** ([#759](https://github.com/higher-kinded-j/higher-kinded-j/issues/759)): a type named in a compile-time message is now built from its element and resolved arguments rather than the mirror's string form, so a component declared `@Nullable Set<?>` is reported as `Set<?>`, the same rendering as the corrected declaration suggested beside it. A message no longer reads as advice to drop a nullness contract it never meant to touch.
+
 ### [v0.4.10](https://github.com/higher-kinded-j/higher-kinded-j/releases/tag/v0.4.10) (30 August 2026)
 
 This release gives the mapper a stock vocabulary and takes every generating annotation through a generics pass: each one now emits source the consuming build compiles under `-Werror`, or explains at the declaration what it needs instead. The book gains a **Mapping at the Boundary** chapter with a law-checked capstone, and the house style it introduced now runs through the Optics and Resilience chapters.
--- a/hkj-processor/src/main/java/org/higherkindedj/optics/processing/FocusProcessor.java
+++ b/hkj-processor/src/main/java/org/higherkindedj/optics/processing/FocusProcessor.java
@@ -475,6 +475,9 @@ public class FocusProcessor extends AbstractProcessor {
                 .map(
                     resolved ->
                         resolved == null ? "Object" : ProcessorUtils.simpleTypeName(resolved));
-    return element.getSimpleName() + "<" + arguments.collect(Collectors.joining(", ")) + ">";
+    return ProcessorUtils.declaredHead(declaredType)
+        + "<"
+        + arguments.collect(Collectors.joining(", "))
+        + ">";
   }
 }
--- a/hkj-processor/src/main/java/org/higherkindedj/optics/processing/GeneratorRegistry.java
+++ b/hkj-processor/src/main/java/org/higherkindedj/optics/processing/GeneratorRegistry.java
@@ -11,6 +11,7 @@ import javax.lang.model.element.Element;
 import javax.lang.model.type.TypeMirror;
 import javax.tools.Diagnostic;
 import org.higherkindedj.optics.processing.spi.TraversableGenerator;
+import org.higherkindedj.optics.processing.util.ProcessorUtils;
 
 /**
  * Resolves which {@link TraversableGenerator} handles a container type.
@@ -106,7 +107,7 @@ public final class GeneratorRegistry {
         "Multiple TraversableGenerator SPI providers with equal priority ("
             + other.priority()
             + ") support type "
-            + type
+            + ProcessorUtils.simpleTypeName(type)
             + ": "
             + matched.getClass().getName()
             + " and "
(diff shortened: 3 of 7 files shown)


MagnusSmith:
@coderabbit full review

MagnusSmith:
@coderabbit  full review

MagnusSmith:
@coderabbit full review

Issue #761, opened 2026-08-28
The book gate does not reach the pages that document what the processor rejects


MagnusSmith:
[#755](https://github.com/higher-kinded-j/higher-kinded-j/issues/755) happened because three documented `@MatchWhen` examples recommended a shape that has never compiled. The processor checks now reject that shape, but nothing stops the *prose* drifting again: the page it drifted on is not gated.

## Measured

`bookVerify` compiles a snippet only when it carries a `<!-- verify -->` marker, or is an anchored `{{#include}}` of real source. Counting java fences across both directories the gate reads:

| | java fences | `<!-- verify -->` | `{{#include}}` | gated |
|---|---:|---:|---:|---:|
| `hkj-book/src` | 2542 | 95 | 110 | **8%** |
| `.claude/skills` | 301 | 214 | 0 | **71%** |

(95 + 214 = 309, which is `MINIMUM_VERIFIED_SNIPPETS`; the include floor is 110. Both ratchets are in place and working — this is not a case of a guard that has been eroded, it is one that was never extended.)

The shipped skills are the better-gated artefact, and they are *derived* from the book. The pages the spec-interface annotations are documented on:

| page | java fences | gated | unchecked |
|---|---:|---:|---:|
| `optics/optics_spec_interfaces.md` | 17 | 7 | **10** |
| `optics/importing_optics.md` | 9 | 2 | 7 |
| `optics/copy_strategies.md` | 14 | 8 | 6 |
| `optics/compiler_errors.md` | 4 | 0 | 4 |
| `optics/annotations_at_a_glance.md` | 1 | 0 | 1 |

`optics_spec_interfaces.md` is the page that carried the #755 examples, and its seven `{{#include}}`s are the reason the *main* worked example was correct all along: `JsonNodeOpticsSpec` is a compiled fixture focusing `StringNode` and `NumericNode`. The drift was entirely in the hand-written fences beside it.

## Two parts, and the second is the interesting one

**1. The positive snippets can be gated today.** Ten fences on `optics_spec_interfaces.md`, three of which are ordinary compilable code; the rest need a per-page fixture supplying the illustrative types (`Value`, `ObjectValue`, `Shape`, `Circle`). That is exactly the pattern `copy_strategies.md` already uses, so it is a known job.

[...]

PR #784, opened 2026-09-01
Maybe bridges to the JDK: toOptional() and stream()

MagnusSmith:
`Maybe` gains its JDK boundary bridges: `toOptional()` and `stream()`.

## What and why

- **`toOptional()`** completes the round-trip that `fromOptional` started: `Just(t)` becomes `Optional.of(t)`, `Nothing` becomes `Optional.empty()`. The conversion is total in both directions, because a `Just` never holds null.
- **`stream()`** renders a `Maybe` as a zero-or-one-element `Stream` with exactly `Optional.stream()`'s contract. Its value is the one idiom `java.util.stream` users reach for: a pipeline of lookups keeps its hits with `.flatMap(Maybe::stream)`, where today the same line reads `.flatMap(m -> m.map(Stream::of).orElseGet(Stream::empty))`.

Both are two-line defaults over the sealed hierarchy's pattern switch, in the style of the existing `toEither` pair.

**A deliberate line, stated in the javadoc, the book and the release note:** `stream()` stays on `Maybe` alone among the optional-shaped types. An empty stream renders absence faithfully; the same conversion on `Either` or `Try` would *silently* discard the error the value carries. The word is load-bearing (the panel's catch): the repo does ship discarding conversions, and they stay named for what they drop, the way `EitherPath.toOptionalPath()` is.

**The book had already promised this method.** Four pages write `maybePath.run().toOptional()` today (`quickstart`, `cheatsheet`, the effect-path overview, `patterns`) — snippets that could not have compiled before this change, none under the `bookVerify` gate (#761 territory). This lands the method they assumed, and settles the name.

## Swept beneficiaries

`Affines.just()`, `Prisms.just()` and `MaybePath.toOptionalPath()` each hand-rolled `maybe.isJust() ? Optional.of(maybe.get()) : Optional.empty()`; all three now read through `toOptional()`, and the panel's grep confirms they were the last in main. One test-side tidy in `NaturalTransformationPropertyTest`, and the file's `Nothing` pattern arms now use `_` per house style.

## Tests and docs

- `MaybeTest` gains a JDK-bridges section in the house assertion vocabulary (`assertThatMaybe`): totality both ways, the round-trip in both directions including the empty leg, presence-as-length with content, and the flatMap idiom end to end. `hkj-core` holds its 100% line/branch/instruction pin.
- The [Maybe Monad](https://higher-kinded-j.github.io/latest/monads/maybe_monad.html#interacting-with-maybe-values) page lists both methods beside the `toEither` bullets, documents `fromOptional` (cited on the page but never shown), and carries a "Bridging to the JDK" example with the Maybe-only rationale. The Optional page's conversion row loses its null-unsound `Optional.ofNullable(maybe.orElse(null))` advice in favour of the bridge; the cheatsheet gains the `Stream` escape-hatch row. Release note under 0.4.11-SNAPSHOT.

## Verification

`:hkj-core:clean check` (JaCoCo pin), full-repo `assemble testClasses` under `-Werror`, `bookVerify`, heading check, spotless. Panel review run after implementation; no blockers, and every should-fix applied: the `MaybePath` sweep, the "silently" precision, the javadoc's closing pointer corrected to routes that exist (`Either.fold`, `Try.foldFailureFirst` — the old sentence linked `Maybe`'s own method and a name deprecated for 0.5.0), the Optional-page row, and the missing round-trip leg.

## Summary by CodeRabbit

* **New Features**
  * Added conversions from `Maybe` to JDK `Optional` and `Stream`.
  * `Maybe.stream()` produces either one value or an empty stream.
  * Improved interoperability guidance and examples for `Maybe`, `Optional`, and streams.

* **Documentation**
  * Updated cheat sheets, monad guides, and release notes with the new conversion options and usage recommendations.


MagnusSmith:
@coderabbit  full review

PR #785, opened 2026-09-01
@ThroughField: compose the auto-detected traversal without the raw cast (#779)

### LINKED ISSUE #779: [FEAT] @ThroughField: compose the auto-detected traversal without the raw cast, so javac checks the focus too

A `@ThroughField` traversal is generated as the spec's lens for the field composed with the container traversal, and the composition goes through a raw cast:

```java
return SpecOptics.items().andThen((Traversal) Traversals.forList());
```

with `@SuppressWarnings("unchecked")` on the generated method. After #773 (PR #778) the cast has nothing left to hide on the auto-detected path: detection reads the focus of the spec's own lens and accepts only the container interfaces, so the lens focus and the traversal's source type agree at erasure by construction. What the cast still hides is the *other* half of the composition: the traversal's focus against the container's element type.

## Measured

Traced against `main` after #778, not run:

```java
public record Bag(List<String> items) { ... }

@ImportOptics
public interface BagSpec extends OpticsSpec<Bag> {
  @ViaBuilder Lens<Bag, List<String>> items();

  Traversal<Bag, Integer> eachItem();      // Integer over a List<String>
}
```

`autoDetectTraversalForField` computes a `ContainerType` for `List<String>` and consumes only its `kind()`; `elementType()` is never compared with the method's declared focus. The generated method compiles under the suppression and throws `ClassCastException` on the caller's first `getAll` or `modify`. Written without the cast, javac rejects it where it is declared:

```java
public static Traversal<Bag, Integer> eachItem() {
  return BagOptics.items().andThen(Traversals.forList());
  // incompatible types: Traversal<Bag,String> cannot be converted to Traversal<Bag,Integer>
}
```

## Why the cast can go on the auto-detected path

```mermaid
flowchart LR
    A["auto-detected:<br/>lens focus is List / Set /<br/>Collection / Map / Optional / T[]"] --> U["compose uncast<br/>javac checks source AND focus"]
    E["explicit traversal = &quot;...&quot;:<br/>a string the processor<br/>cannot type-check"] --> C["keep the raw cast<br/>and the suppression"]

    class A,E domain
    class U good
    class C decision
```

`Traversals` gives exactly the types the composition needs: `forList() → Traversal<List<A>, A>`, `forSet() → Traversal<Set<A>, A>`, `forCollection() → Traversal<Collection<A>, A>`, `forOptional() → Traversal<Optional<A>, A>`, `forArray() → Traversal<A[], A>`, `forMapValues() → Traversal<Map<K, V>, V>`, and `Lens.andThen(Traversal<A, B>) → Traversal<S, B>`. With the lens focus exactly one of those interfaces, `BagOptics.items().andThen(Traversals.forList())` infers unaided: `A` from the lens focus, `B` checked against the method's declared focus. Generic specs (`Lens<Box<U>, List<U>>`), a `Map<K, V>` focus and a `String[]` focus all infer the same way; the one shape that still needs the cast is a lens focus carrying a wildcard (`List<? extends Leaf>`), which `ProcessorUtils.hasUndenotableTypeArguments` already recognises, and a raw focus is refused before this point.

The explicit-`traversal` route is different in kind: the annotation carries a string the processor splices into the generated source verbatim, so there is nothing to type-check at generation time, and the cast is the author's undertaking that theirs rebuilds the declared type. It keeps the cast and the suppression.

## What to do

- `TraversalHintInfo.forThroughField` carries only the traversal string; it needs to say whether the traversal was auto-detected (a boolean, or two factory methods), threaded to the two consumers: `TraversalCodeGenerator.generateThroughFieldCode`, which emits the cast, and `SpecInterfaceGenerator` (around the `THROUGH_FIELD` check), which adds the `@SuppressWarnings("unchecked")`.
- On the auto-detected path with a denotable lens focus, emit `Spec.x().andThen(Traversals.forList())` with no cast and no suppression. A focus-type mismatch then reaches the author as javac's own `incompatible types` inside the generated file, which is the #746 family; better still, compare `containerType.elementType()` (boxed, wildcards resolved through `ProcessorUtils.resolveWildcard`) with the method's declared focus in `autoDetectTraversalForField` and refuse at the declaration in the what/why/fix shape, so the uncast composition is a second line of defence rather than the diagnostic.
- Golden files and `GeneratedFocusSourceCompilesTest`-style axes should pin both shapes: an auto-detected method compiles clean under `-Xlint:unchecked,rawtypes -Werror` with no suppression; an explicit-traversal method keeps the cast and still compiles.

## Notes

- Surfaced by the review panel on #778, where two lenses proposed it independently; left out of that PR because it changes the generated shape for every auto-detected `@ThroughField` method and needs the `TraversalHintInfo` change, which is a type-safety-initiative item (issues #560–#565) rather than a fix.
- This is the last unchecked cast on the `@ImportOptics` traversal path; removing it also removes a `@SuppressWarnings` from generated source, which is what the checker gate in #695 targets.

MagnusSmith:
Closes #779.

A `@ThroughField` traversal composed the spec's lens with the container traversal through a raw cast, under a generated `@SuppressWarnings("unchecked")`. After #773 the cast had nothing left to hide on the auto-detected path except the one thing it still hid: the traversal's focus against the container's element. A `Traversal<Bag, Integer>` over a `List<String>` field compiled and threw `ClassCastException` on the caller's first use.

## What changes

```mermaid
flowchart LR
    A["auto-detected,<br/>wildcard-free lens focus"] --> U["focus held to what the traversal<br/>hands back, then composed uncast:<br/>no cast, no suppression"]
    W["auto-detected,<br/>wildcard-carrying focus"] --> C["cast and suppression stay;<br/>the focus still held: an extends-<br/>wildcard's bound, or Object<br/>under a super-wildcard"]
    E["explicit traversal = &quot;...&quot;"] --> X["cast and suppression stay;<br/>the focus is the author's"]

    class A,W,E domain
    class U good
    class C,X decision
```

- **The refusal is the diagnostic; the uncast composition is the second line of defence.** `autoDetectTraversalForField` holds the method's focus to what the standard traversal hands back — the element, a `Map`'s values, or `Object` where the element sits behind a super- or unbounded wildcard (the panel's catch: skipping that case left the original disease alive there). The comparison is **containment, not sameness**: `Types.isSameType` is specified false for wildcards, so the first cut falsely refused a sound `Traversal<Bag, ? extends CharSequence>` that had always compiled — `Types.contains` accepts it and still refuses both a narrowing and a widening focus, each named in the message with its own failure mode (a widening focus pollutes rather than throws).
- **The checked body is a target-typed local, and the local is load-bearing.** The generated lens is a generic method; called bare as a receiver it infers against nothing (found the hard way on the generic-spec axis). `Lens<Box<U>, List<U>> lens = SubjectOptics.items(); return lens.andThen(Traversals.forList());` target-types the lens call and hands javac both halves.
- **The analyser validates; the generator decides the shape** (the panel's restructure). The analyser marks an auto-detected traversal over a denotable focus as a candidate and carries the lens focus; `SpecInterfaceGenerator` makes the final call against the very `methodTypeParameters` list it declares on the method — a spec variable only the lens names has nowhere to come from in the generated file and keeps the cast — and one value, the checked local's `TypeName` built by the same `buildOpticReturnType` as the sibling lens's signature, drives both the body shape and the suppression, so the two cannot disagree.
- **The explicit `traversal` string keeps the cast and the suppression** and is the deliberate escape hatch: the same mismatched focus the auto path refuses is accepted there, pinned by test. The remaining generated suppressions elsewhere (`PathSourceProcessor`, `EffectAlgebraProcessor`, `MappingProcessor`, the plugin traversal generators) sit behind casts the processor genuinely cannot type-check; only this composition had a checkable half.

## Tests

Sixteen rows in a new `CheckedComposition` section, panel-audited against pre-change HEAD (six fail before, the rest are labelled ratchets): the issue's reproduction and the widening, wildcard-mismatch, `Map`-value, `Optional`-element and super-wildcard-element refusals verbatim with `hadErrorCount(1)`; the wildcard-focus-over-element regression pin; uncast shape under `-Xlint:unchecked,rawtypes -Werror` with neither cast nor suppression; the explicit-path escape hatch; both generator-guard arms through raw-accessor fixtures (comments say why raw is what lets such a lens generate); a unit row pinning the checked body's emission. `SpecInterfaceAnalyser*` holds its 100% pin; the full suite, including #778's runtime rows, passes over the new generated shape.

## Docs

The `@ThroughField` javadoc, the [auto-detection section](https://higher-kinded-j.github.io/latest/optics/copy_strategies.html#throughfield-auto-detection) (its own paragraph, clear of the #773 remedy), a [Compiler Errors](https://higher-kinded-j.github.io/latest/optics/compiler_errors.html) entry, and the 0.4.11-SNAPSHOT release note — each stating the wildcard-free qualifier rather than claiming the uncast shape for every auto-detected method.

_Behaviour change:_ a focus that did not contain the container's element compiled in 0.4.10 and misbehaved at runtime (throwing where it narrowed, polluting where it widened); it is now refused at the declaration. The generated shape changes for auto-detected methods over a wildcard-free lens focus, with the same behaviour.

## Verification

`:hkj-processor:clean check` (coverage pins, ArchUnit rules), `:hkj-annotations:javadoc`, full-repo `assemble testClasses` under `-Werror`, `bookVerify`, heading check, spotless. Five-lens panel run after implementation; no blockers, and every should-fix applied: the two wildcard regressions above, the generator-side restructure, the docs qualifiers, the escape-hatch and widening rows, and the message rework.

## Summary by CodeRabbit

- **Bug Fixes**
  - `@ThroughField` now validates that the declared focus matches the element type produced by auto-detected traversals.
  - Invalid focus declarations are rejected at compile time instead of causing `ClassCastException` during use.
  - Compatible wildcard and generic focus types continue to be supported.
  - Auto-detected compositions now receive stronger type checking and avoid unnecessary casts where possible.
  - Explicit traversals retain their existing cast behavior.

- **Documentation**
  - Added guidance and compiler-error documentation covering focus mismatches, wildcard handling, and recommended fixes.


MagnusSmith:
@coderabbit full review

PR #786, opened 2026-09-03
@GenerateFocus: reject a declined container from the analysis that declined it (#758)

### LINKED ISSUE #758: FocusProcessor's rejection walk hand-mirrors WideningAnalysis.collect()
#719 established that `@GenerateFocus` has **one** widening analysis: `WideningAnalysis` decides, `FocusProcessor` emits what it decided, and a navigator composes the static method rather than re-deriving anything. `FocusProcessor.reportUndenotableWidening` is the one place that still asks the same question a second way, by walking the component's type itself and reproducing the analysis's stopping rules by hand.

```mermaid
flowchart TD
    D["record component"] --> A["WideningAnalysis.collect()"]
    D --> W["FocusProcessor.reportUndenotableWidening()"]

    A --> A1["peels a layer"]
    A1 --> A2{"widens here?"}
    A2 -->|no| A3["stop, no step"]
    A2 -->|yes| A4["append step, descend"]

    W --> W1["peels a layer"]
    W1 --> W2{"widensHere(...)?"}
    W2 -->|no| W3["stop, no diagnostic"]
    W2 -->|yes| W4["check undenotable, descend"]

    A3 -.->|"must agree"| W3
    A4 -.->|"must agree"| W4

    classDef one fill:#a6d189,stroke:#40a02b,color:#232634
    classDef two fill:#e5c890,stroke:#df8e1d,color:#232634
    class A,A1,A2,A3,A4 one
    class W,W1,W2,W3,W4 two
```

Both walks must reach the same containers and stop in the same places, and nothing enforces it. They agree today — the #754 review traced eleven nested shapes and found no drift — but the agreement is maintained by hand, which is precisely the shape #719 removed everywhere else.

## What made it more visible

#725 extended the mirrored walk to a second family of containers. It also left the processor choosing *which* undenotability question to ask:

```java
boolean undenotable =
    recognised
        ? analysis.recognisedWidensUndenotably(declaredType)
        : WideningAnalysis.widensUndenotably(generator, declaredType);
```

The analysis knows which of those applies; the processor should not have to.

## Two options

**Cheap.** One method on the analysis — `widensUndenotably(TypeMirror, TraversableGenerator)` — that answers for a recognised container and an SPI one alike, so `FocusProcessor` asks one question and the branch above disappears. Leaves the mirrored walk in place.

**Better.** Have `collect()` record a *declined* step when it turns a container away, carrying the type that caused it. `reportUndenotableWidening` then reads the analysis's own result and reports it, instead of re-walking to rediscover what the analysis already worked out. That removes the second walk entirely and makes the two answers the same object, which is what #719 did for the static-method / navigator pair.

The second option needs care with the `@Nullable` interaction: `analyse()` appends a `NULLABLE` step only when `steps` is empty, and a declined container currently leaves it empty. A declined step would change that, so the tier for `@Nullable Set<?>` needs pinning either way (it fails the build regardless, since the declaration is rejected in the same round).

## Guard to add alongside

`GeneratedFocusSourceCompilesTest` covers the over-generation half — no error may come from generated output. Nothing covers the other half: a container the analysis declines silently and the walk never reports would emit a plain `FocusPath` with no diagnostic, and no test would notice. A corpus asserting "declined by the analysis ⇒ reported at the declaration" would close it, and would keep the two walks honest for as long as they both exist.

## Notes

Found by the review panel on #754. Not a defect: no drift was found. It is the last hand-maintained copy of an answer #719 made single-sourced everywhere else.

MagnusSmith:
Closes #758.

`@GenerateFocus` had one analysis of what a component widens to and a second, hand-written walk that re-derived where that analysis stops, so that a raw or wildcard-carrying container could be rejected at its declaration. The two agreed by care alone. This takes the issue's second option: the analysis records the container it turns away, and the processor reports from the same result the generated method is built from.

```mermaid
flowchart LR
    subgraph before["before"]
        direction TB
        D1["record component"] --> A1["WideningAnalysis.collect()"]
        D1 --> W1["FocusProcessor.reportUndenotableWidening()"]
        A1 --> E1["static method"]
        W1 --> R1["diagnostic"]
        A1 -.->|"must agree"| W1
    end
    subgraph after["after"]
        direction TB
        D2["record component"] --> A2["WideningAnalysis.collect()"]
        A2 --> WD["Widening(tier, focusType, steps, declined)"]
        WD --> E2["static method"]
        WD --> R2["diagnostic"]
    end

    classDef one fill:#a6d189,stroke:#40a02b,color:#232634
    classDef two fill:#e5c890,stroke:#df8e1d,color:#232634
    class A1,E1,A2,WD,E2,R2 one
    class W1,R1 two
```

## What changed

- **`WideningAnalysis.Widening` carries `declined`**, the `DeclaredType` the walk turned away, or null. `collect()` returns it; the SPI layer moves into `collectSpi()`, which tells "no generator supports this" from "the generator cannot write this widening". The cardinality check comes first, so a `ZERO_OR_MORE` container nothing steps into is left alone rather than declined, and nothing beneath it is looked at. A declined walk stops where it met the container, so the steps still describe a method that compiles: the widening as far as it goes.
- **`FocusProcessor` reports from that result.** The mirrored walk, `widensHere`, and the `recognised ? … : …` choice of undenotability question are gone. The static-method branch analyses once, reports `declined` if set, and builds the method from the same `Widening`. A field a navigator took reaches its element through a container the navigator's guard already admitted, so it has nothing to report.
- **The navigator's turned-away signal honours its own contract.** `widensUndenotableSpiContainer` documents itself as "true when only the type arguments stand between this component and a navigator", but did not ask whether the element was generic, which `navigatorTarget` does. Measured before the change: `Map<String, ? extends Box<String>>` under `generateNavigators = true` was rejected while `Map<String, Box<String>>` compiled as a plain `FocusPath`. The wildcard twin now compiles the same way. This is the one behaviour change, and it is listed in the release history. One asymmetry remains and is deliberate: the concrete twin gets the existing note explaining that a generic target gets no navigator, while the wildcard twin is silent, because that note's advice (`widenCollections = true`) would lead it straight to a rejection.
- **The SPI lookup is classified once.** `WideningAnalysis.spiLookup` answers `None`, `Refused(generator)` or `Admitted(generator)`, so no site ever holds a bare generator for a container whose widening cannot be written. The walk switches on it, the navigator steps into `Admitted` only, and the navigator's turned-away check matches `Refused`. `findSpiGenerator`, `wideningGenerator` and the public `widensUndenotably` are gone, which removes the second application of the guard the review panel spotted.
- **The ArchUnit rule now guards what matters.** "SPI lookups go through the guard" (an allow-list of callers) becomes "a refused generator widens nothing": only the walk and the navigator's turned-away check may read `Refused.generator()`, and both read it to ask what the generator would have done, never to widen. A companion test pins that the rule sees exactly those two readers, so a rename cannot make it pass vacuously. The registry rule's reader list follows the lookup's new name.
- `recognisedWidensUndenotably` is removed and `typeArgumentAt` made private.

## The `@Nullable` interaction

A declined container is a separate component of `Widening`, not a step, so `analyse()`'s rule that `NULLABLE` is appended only when no step was taken is untouched: `@Nullable Set<?>` still widens through `.nullable()` to an `AffinePath<Holder, Set<?>>` that compiles, and the declaration is rejected all the same. Pinned in `WideningAnalysisDeclinedTest` and in the rejected corpus.

## Tests

- `GeneratedFocusSourceCompilesTest` gains the guard the issue asks for: a rejected-shape corpus beside the still-compiles one, each entry asserting the container the analysis turns away is named at the declaration and that nothing comes back from generated output.
- `WideningAnalysisDeclinedTest` reads `Widening.declined()` straight off the analysis through a probe processor: which layer is declined under which setting, that the steps stop where it was met, and the `@Nullable` tier.
- `FocusProcessorSpiEnhancementsTest` pins the generic-element corner for both the concrete and the wildcard container.

`gradle :hkj-processor:check` is green: tests, ArchUnit rules, spotless and the JaCoCo gates, with `WideningAnalysis`, `NavigatorClassGenerator` and `FocusProcessor` at 100% line, branch and instruction coverage.

https://claude.ai/code/session_01PCF71VE6S9uGidw5jtjyog

## Summary by CodeRabbit

- **Bug Fixes**
  - Generic navigable containers now compile successfully when navigation does not descend into their generic elements.
  - Container handling now consistently preserves plain `FocusPath` behavior for equivalent concrete and wildcard forms.
  - Containers that cannot be safely widened continue to be rejected with clear diagnostics.
  - Nullable components retain their expected behavior when widening is declined.

- **Documentation**
  - Updated the release history with details of the improved container and navigator handling.

### DIFF
--- a/hkj-book/src/release-history.md
+++ b/hkj-book/src/release-history.md
@@ -40,6 +40,10 @@ _In development. Release notes will be added here as changes land on `main`._
 
   _Behaviour change, moving a runtime failure to the declaration:_ a focus that did not contain the container's element compiled in 0.4.10 and misbehaved at runtime, throwing where it narrowed and polluting the container where it widened; it is now refused. The generated shape changes for auto-detected methods over such a lens focus (a target-typed local in place of the cast), with the same behaviour.
 
+- **`@GenerateFocus` leaves a wildcard container alone when nothing would widen it** ([#758](https://github.com/higher-kinded-j/higher-kinded-j/issues/758)): a `ZERO_OR_MORE` container whose element is a generic navigable record, such as `Map<String, ? extends Box<String>>` under `generateNavigators = true`, gets no navigator whatever its arguments, so its wildcard is no longer reported and the component stays a plain `FocusPath`, as its concrete twin `Map<String, Box<String>>` already did. The check that refuses a raw or wildcard-carrying container now reads the same result the generated method is built from, so the two cannot disagree. See [Custom Containers](optics/focus_containers.md#supported-container-types).
+
+  _Behaviour change, in that one corner:_ the shape was refused in 0.4.10 although nothing widened it; it now compiles as written. Every container that was rejected because something did widen it is rejected still, with the same message.
+
 ### [v0.4.10](https://github.com/higher-kinded-j/higher-kinded-j/releases/tag/v0.4.10) (30 August 2026)
 
 This release gives the mapper a stock vocabulary and takes every generating annotation through a generics pass: each one now emits source the consuming build compiles under `-Werror`, or explains at the declaration what it needs instead. The book gains a **Mapping at the Boundary** chapter with a law-checked capstone, and the house style it introduced now runs through the Optics and Resilience chapters.
--- a/hkj-processor/src/main/java/org/higherkindedj/optics/processing/FocusProcessor.java
+++ b/hkj-processor/src/main/java/org/higherkindedj/optics/processing/FocusProcessor.java
@@ -21,11 +21,8 @@ import javax.lang.model.element.RecordComponentElement;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.element.TypeParameterElement;
 import javax.lang.model.type.DeclaredType;
-import javax.lang.model.type.TypeKind;
-import javax.lang.model.type.TypeMirror;
 import org.higherkindedj.optics.Lens;
 import org.higherkindedj.optics.annotations.GenerateFocus;
-import org.higherkindedj.optics.processing.spi.Cardinality;
 import org.higherkindedj.optics.processing.spi.TraversableGenerator;
 import org.higherkindedj.optics.processing.util.Diagnostics;
 import org.higherkindedj.optics.processing.util.ExcludeFromJacocoGeneratedReport;
@@ -216,7 +213,6 @@ public class FocusProcessor extends AbstractProcessor {
 
     // Generate FocusPath methods for each component
     for (RecordComponentElement component : components) {
-      reportUndenotableWidening(component, widenCollections, navigatorGenerator, analysis);
       MethodSpec method = null;
 
       // Try to create a navigator method if navigators are enabled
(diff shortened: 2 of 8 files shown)


MagnusSmith:
@coderabbit full review

PR #793, opened 2026-09-05
Gate the pages that document what the processor refuses (#761)

### LINKED ISSUE #761: The book gate does not reach the pages that document what the processor rejects

MagnusSmith:
Closes #761.

The book gate could only ask one question — *does this compile?* — so the pages documenting what the processor **refuses** were the least verifiable in the book while being exactly the ones a processor change invalidates. #755 is what that costs: three `@MatchWhen` examples recommended a shape that has never compiled, on a page whose seven `{{#include}}`s were correct throughout. Every diagnostic reworded in #723, #730, #733, #748 and #755 was reworded with no test that the book still quoted it correctly.

This PR does two things: it teaches the gate to assert a refusal, and then it rolls the gate across the book.

```mermaid
flowchart TD
    S["a book snippet"] --> Q{"what is it showing?"}
    Q -->|"real source"| I(["{{#include}}"])
    Q -->|"working code"| V(["&lt;!-- verify --&gt;"])
    Q -->|"code the processor rejects"| R(["&lt;!-- verify:rejects &quot;…&quot; --&gt;<br/>compiles, and must fail<br/>quoting the message"])
    Q -->|"code it merely warns about"| N(["&lt;!-- verify:reports &quot;…&quot; --&gt;<br/>compiles, and a note or<br/>warning must quote it"])

    classDef added fill:#8caaee,stroke:#1e66f5,color:#232634
    class S,I,V tier
    class Q decision
    class R,N added
```

## The mechanism

| Marker | What it asserts |
|---|---|
| `<!-- verify -->` | the snippet compiles, with no error and no warning |
| `<!-- verify:rejects "…" -->` | the snippet does **not** compile, and one of the errors quotes the fragment |
| `<!-- verify:reports "…" -->` | the snippet compiles, and a note or a warning quotes the fragment |

The quoted fragment is the half that matters: a bare "still refused" says nothing about the wording, and the wording is what rots. `verify:reports` covers the diagnostics that do not stop a build — `@GenerateTraversals`' note for a container no generator claims, `@GeneratePathBridge`'s warning when no `@PathVia` method survives — which a compile check cannot see at all.

Guards on the marker itself:

- a fragment must be at least ten characters, so it cannot be whittled down until it matches any message;
- a misspelt tail (`<!-- verify:rejcts … -->`) is an error naming the page and line, rather than being read as the bare form and quietly turning a rejection check into a compile check.

Three deliberate mutations proved it, each caught with an actionable message (reverted afterwards): a reworded error, a `@GeneratePrisms abstract class` turned into an `enum`, and a reworded note.

## The rollout

`SnippetExtractor` gained a run of structural fixes as the sweep hit them, each with a page that could not otherwise be gated:

- a loose statement now keeps every line until its braces close, so an anonymous class written inside one (an interpreter, a `Natural`) stays where the page put it;
- a method signature may wrap before its parameter list, which a generic return type routinely does;
- a snippet that is **only** body-less signatures becomes an `interface`, where a quoted signature type-checks — which is what brings the `@InstanceOf`/`@MatchWhen` comparison, and the `StreamOps` reference tables, under the gate;
- a fixture may be generic, and its bounded parameters are carried onto the `extends Fixture<…>` clause with their bounds intact;
- the look-ahead past a run of annotations steps over a *whole* wrapped annotation, so a `@JsonSubTypes({…})` listing a type per line no longer hides the declaration beneath it;
- `package` lines are dropped, imports may carry a trailing note, and a nested type may be `private` or `protected`;
- `typeNameOf` reads the declaration with its comments stripped: a note ending in "…for each case of the sealed interface" was making the extractor read the next line's `public` as the type's name, so the snippet's type never matched the fixture's and both were emitted;
- a type declared on the **same line** as its annotations is recognised as a declaration, which is how the book writes a one-line model record — before this the whole line read as an annotation and the type was left inside the statement body, where it could not carry a companion;
- a field ends at its semicolon rather than where its braces first balance, so an initialiser passing a lambda with a body no longer has its remaining arguments cut loose.

| Chapter | Snippets now gated |
|---|---:|
| `optics/` | 721 |
| `effect/` | 504 |
| `monads/` | 285 |
| `functional/` | 143 |
| `transformers/` | 142 |
| `hkts/` | 102 |
| `glossary/` | 81 |
| `spring/` | 79 |
| `tutorials/` | 55 |
| `resilience/` | 45 |
| `examples/` | 43 |
| `tooling/` | 17 |

Every chapter has been swept. Measured like for like:

| | before | after |
|---|---:|---:|
| carry a marker | 102 / 2556 (**4%**) | 2221 / 2607 (**85%**) |
| marker **or** `{{#include}}` | 200 / 2556 (**8%**) | 2320 / 2607 (**89%**) |

What stays unmarked is catalogued in `BOOK-SNIPPETS.md`, each entry naming the page that
prompted it: shape quotations of the library's own types, laws written as equations,
reference tables of bare calls, free type variables, dependencies the gate does not carry
(Vavr, JMH, jOOQ, Lombok, Immutables), a `package-info` annotation, a complete runnable file
repeated at the end of a page, and the classes a page continues across several blocks. That
catalogue is explicitly **not** a closed set: around eighty fences match none of its entries,
mostly deliberate pseudo-code and equivalence pairs, and the document now says so rather than
implying every gap was a decision.

## What compiling the book found

The defects fall into families, and they repeat:

**Phantom API.** `Path.of`, `Path.ioOf`, `IOPath.delay` (22 sites), `StatePath` (an entire type that is `WithStatePath`), `Instances.eitherT` misused, `Maybe.fold`, `ID.unwrap`, `Const.of`, `Reader.asks`, `ReaderT.lift`, `TrampolineUtils.map2StackSafe`, `StreamOps.empty`, `FreeAp.analyze` (it is `analyse`), `TaskMonad.INSTANCE`, `OrderErrors.outOfStock`, and two wrong import paths (`optics.annotation`, `optics.each.Each`).

**Kind helpers are enum singletons.** `reader_monad.md` was written end to end against a static API that does not exist; `StateTKindHelper.runStateT` and `EitherFKindHelper.widen` were called on the class.

**`For` past its first binding hands over the accumulated tuple**, not the last value — wrong on eight pages, in three chapters.

**Unconstrained witnesses.** `var eitherTMonad = Instances.eitherT(futureMonad)` leaves `L` to infer to `Object`; the same for `readerT`'s environment and `stateT`'s state. The chapter's own Common Compiler Errors §4 documents this, and the chapter's own pages tripped over it.

**Paths that carry no outer monad.** `Path.maybe`, `Path.optional` and `Path.either` were each fed an asynchronous lookup on the page that introduces the matching transformer — which is precisely why the transformer section follows.

**`Scope.join()` takes no arguments**, on seven pages across two chapters.

**Duplicate declarations in one block** — around twenty, including several before/after pairs that declared the same variable twice.

The optics chapter, swept last and by far the largest, added families of its own:

**A composition declared as the wrong optic.** `composition_rules.md` states that a prism composed
with a lens settles at `Affine` and then declares one of those compositions as a `Traversal`;
`affine.md` composed an affine to a timeout and declared it `String`; `folds.md` asked
`ProductLenses.quantity()` of a record with no quantity.

**A companion nobody asked for.** Eleven records across the cookbook, the composition rules and the
recipes named an `XLenses` or an `XPrisms` without carrying the annotation that generates it, and
several composed off the *interface* rather than off the variant a prism had just narrowed to.

**One name, two meanings.** `mapAt` was both the score map's `At` and the settings map's;
`config` meant three different records on one page; `Order`, `Customer` and `ValidationError` each
meant two things; `loLens` meant a `Range` lens in one section and a `Triple` lens in another.

**Constants reading locals.** A `static final` map of per-environment auditors, a `ValidationOptics`
class and a registration validator each read names that were locals in the block above them.

**Element types nothing pinned**, so they inferred `Object`: three `ListTraversals` slices, a
`Prisms.some()`, a `Prisms.just()` and a `Fold.of`.

**Methods that are not there.** `IxedInstances.listIxed().ix(4).getOptional(…)` (it is
`IxedInstances.get(listIx(), 4, …)`), `Lenses.get`, `Traversals.exists`, `IntSumMonoid.INSTANCE`,
`StringMonoid.INSTANCE`, `CFMonad.INSTANCE`, and a `MockOpticInterpreter<Person>` built as a
generic taking one canned value where the class the page had just written takes the caller's stubs.

Plus: `Natural`'s method is generic, so it can never be a lambda; `Saga.of`'s two overloads make an implicitly typed compensation lambda ambiguous; `StateT.mapT` is the one `mapT` that takes the target monad; and a page cannot declare a type called `Record`.

## What the pre-merge review changed

A five-lens review over the finished branch found defects in the sweep itself, and they are
fixed here rather than left for a follow-up:

- **Three pages were gated against stand-ins for the types they document.** `capabilities.md`
  quoted the six capability interfaces against copies declared *without* `sealed` - so the gate
  certified a hierarchy a reader cannot implement, and could only pass because the fake was
  there. `context_security.md` and `context_request.md` were gated against clones of
  `SecurityContext` and `RequestContext`, which is precisely what stops the gate noticing those
  classes moving. All three now follow the convention the book already uses for its own types:
  the quotation is unmarked, the fixture imports the real class, and every worked use is held
  to it. That **lowers** `MINIMUM_VERIFIED_SNIPPETS` by fourteen, deliberately: those snippets
  were passing against stand-ins, so the count was inflated by fourteen rather than reduced.
- **`Error` was a placeholder that bound to `java.lang.Error`.** No page declared it, so in a
  reader's project `Path.left(new Error("failed"))` still compiles - silently putting a
  `Throwable` in the left channel - and the fixtures declared a local `Error` so the gate could
  not see the swap. Renaming it to `AppError` across both roots exposed three book pages and one
  skill that were compiling against the JDK type.
- **A fixture richer than its page turns a book error into a passing test.** The optics cookbook
  and the focus-navigation page each called a generated accessor for a component their own model
  omitted; a fuller fixture had been hiding it.
- **`verify:rejects "cannot find symbol"`** asserted nothing in particular - javac's universal
  error - and contradicted `BOOK-SNIPPETS.md`'s own rule. It now quotes the typo it is about.
- **`MINIMUM_SKILLS_SNIPPETS` is new**, for the reason the diagnostic floor exists: one combined
  total cannot protect a root, so gating a book snippet would offset deleting a skill's marker.
- Plus three defects in the harness and the sweep: `endOfField` ended a member only at a `;`, so
  a one-line `static` method swallowed the lines after it; twelve fence openers were written at
  column zero under an indented marker, which breaks list continuation; and three "phantom API"
  corrections were wrong - `Monoids.maximum()`, `OpticOps.exists` and `MaybePath.getOrElse` all
  exist, and the pages that worked around them now use them.

## Ratchets

| | before | after |
|---|---:|---:|
| `MINIMUM_VERIFIED_SNIPPETS` | 316 | 2435 |
| `MINIMUM_SKILLS_SNIPPETS` | - | 214 |
| `MINIMUM_DIAGNOSTIC_SNIPPETS` | — | 50 |

## Follow-up

The HKJ checker is a javac *plugin*, so the gate does not run it, and `compile_checks.md`'s own examples are therefore two of the three snippets on that page that cannot be marked. Turning it on works — it fires correctly on both, and finds around a dozen real advisories elsewhere in the book — but it changes what every gated snippet must satisfy, so it belongs in its own change. `BOOK-SNIPPETS.md` records it.

https://claude.ai/code/session_01DcyFMv6hfbU5yCgirypngj


MagnusSmith:
@coderabbit full review

PR #816, opened 2026-09-08
@Generated reaches every nested type the processors write (#798)

### LINKED ISSUE #798: [BUG] `@Generated` stops at the outer class, so nested generated types count against coverage
## What

`@EffectAlgebra` writes `@Generated` on the top-level types it emits and on none of the nested ones. JaCoCo's generated-code filter reads the annotation on each class file, and a nested class is its own class file, so `XOps.Bound`, `XKind.Witness` and `XKindHelper.XHolder` are measured as though someone had written them.

## Measured

Generated by 0.4.10 for an algebra `XOp`:

```java
// XOpOps.java
@Generated
public final class XOpOps {                                 // line 27-28
  …
  public static final class Bound<G extends WitnessArity<TypeArity.Unary>> {   // line 127: no annotation
```

```java
// XOpKind.java
@Generated
public interface XOpKind<A> extends Kind<XOpKind.Witness, A> {              // line 17-18
  final class Witness implements WitnessArity<TypeArity.Unary> {}           // line 22: no annotation
```

What JaCoCo then reports, from one adopter's `jacocoTestReport.csv` across two modules, each with one algebra (instructions missed / covered, branches missed / covered):

| module | class | instr. missed | instr. covered | br. missed | br. covered |
|---|---|---|---|---|---|
| A | `XOpOps.Bound` | 81 | 0 | 12 | 0 |
| A | `XOpKind.Witness` | 7 | 0 | 2 | 0 |
| A | `XOpKindHelper.XOpHolder` | 0 | 6 | 0 | 1 |
| B | `YOpKindHelper.YOpHolder` | 0 | 6 | 0 | 1 |

Module B hides the first two by name, with a comment that explains what they are:

```kotlin
// HKJ @EffectAlgebra codegen the module does not call. YOpOps.Bound is the
// composed-effects instance API (for programs spanning several algebras via
// @ComposeEffects); this slice has one algebra and uses YOpOps' static
// constructors, which ARE covered. YOpKind.Witness is a phantom type marker with
// no callable body. …
"com/example/effects/YOpOps\$Bound*",
"com/example/effects/YOpKind\$Witness*"
```

Module A has no such exclude, so its 95.5% line floor quietly absorbs 81 uncovered instructions it cannot act on. Both modules gate on branch coverage as well as line, and `Bound`'s twelve missed branches are the kind of number that decides whether a floor holds.

## Why it matters beyond coverage

Error Prone's generated-source skip and NullAway's `TreatGeneratedAsUnannotated` read the same annotation, so the nested types are also held to a nullability contract their template never made. Nothing fails today only because nobody calls `Bound` from checked code.

## Fix

Emit `@Generated` on every nested type the processors write. With JavaPoet that is `addAnnotation(Generated.class)` on the nested `TypeSpec.Builder`s as well as the outer. `Generated` already has `CLASS` retention, which is what JaCoCo needs; nothing else changes. The golden files pin it.

Worth a sweep of the other generators for the same shape: `@ComposeEffects`' `*Support` (its `BoundSet` and the `Inject` instances), the `*KindHelper` holders, and any nested companion the optics processors emit. One test that walks every generated `TypeElement` in the golden set and asserts the annotation on each nested type closes the class.

## After

A consuming build needs no exclude list.

MagnusSmith:
Closes #798.

## What

Every nested type a processor writes now carries `@Generated` itself. Seven sites in five generators:

| generator | nested type |
|---|---|
| `@EffectAlgebra` | `XKind.Witness`, `XKindHelper.XHolder`, `XOps.Bound` |
| `@ComposeEffects` | `XSupport.BoundSet` |
| `@GenerateAssembly` | `XAssembly.StageN` |
| `@GenerateErrorEnvelope` | `XErrors.ContextBuilder` |
| `@GenerateFocus` | every `XFocus.*Navigator` |

## Why

A nested type compiles to a class file of its own, and JaCoCo's generated-code filter reads the annotation from the class file it is measuring, never from an enclosing one. The outer marker alone left the nested types measured as though someone had written them, which is what the issue's adopter saw as 81 missed instructions on an `Ops.Bound` nobody calls, and what this repo's own `hkj-examples` output shows when JaCoCo's `Analyzer` is run over it:

| class | instructions counted before | after |
|---|---|---|
| `FraudCheckOpOps` (outer) | 0 | 0 |
| `FraudCheckOpOps$Bound` | 31 | 0 |
| `FraudCheckOpKind$Witness` | 7 | 0 |
| `FraudCheckOpKindHelper$FraudCheckOpHolder` | 6 | 0 |
| `PaymentEffectsSupport$BoundSet` | 15 | 0 |

```mermaid
flowchart LR
  S["XOps.java"] --> O["XOps.class<br/>@Generated"]
  S --> B["XOps$Bound.class<br/>@Generated (new)"]
  O --> F["JaCoCo filter: skipped"]
  B --> F
```

`Generated` is `CLASS`-retained already, so nothing else changes: the marker is not `@Documented`, the HKJ checker's gate already skipped nested types through the enclosing declaration, and the generated source differs only by the annotation line the three assembly goldens now pin.

## The test

`GeneratedMarkerReachesNestedTypesTest` compiles one fixture per generator and reads what the coverage tool reads: the `RuntimeInvisibleAnnotations` of every generated class file, parsed with `java.lang.classfile`, asserting the marker on each. Each row names the nested classes its fixture must produce, so a fixture that stops exercising the shape fails rather than passing vacuously. The existing `contains("@Generated")` assertions could not see this bug, since the outer annotation satisfies them; the new test fails on all five rows without the fix.

## Out of scope, deliberately

`@GenerateTraversals`, `@ImportOptics` and `@GenerateFolds` write their implementations as anonymous classes, which cannot carry a type annotation, so `XTraversals$1` and friends are still measured (`hkj-examples` builds 32 of them, 22 to 24 instructions each). Annotating the anonymous class's method is not enough, as the constructor javac writes for it stays measured (6 instructions on a fixture); only a named nested type is filtered whole. That is a different-shaped change to three generators, filed as #817.

## Verification

- `gradle :hkj-processor:clean :hkj-processor:check` is green on a clean build: 2366 tests, spotless, ArchUnit and the JaCoCo gates. The five touched generators are at 100% line and instruction coverage; the one missed branch in `ComposeEffectsProcessor` (line 547) predates this change.
- `:hkj-examples:compileJava` and `:hkj-examples:compileTestJava` are green, so every generated shape in the examples still compiles under `-Werror` with the checker on.
- The new test fails on all five rows with the processor changes stashed, and passes with them.
- JaCoCo 0.8.14's `Analyzer` over the regenerated `hkj-examples` output reports 0 instructions for every nested type in the table above, and for `SignupUserAssembly$Stage1` and `CompanyFocus$HeadquartersNavigator`.

https://claude.ai/code/session_01Pa775nMrUgnwWUsgm58XDY

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved coverage reporting for generated nested types by ensuring they carry the appropriate generated-code marker.
  - Generated stage, navigator, context, effect, and witness types are now recognized consistently by coverage tools.

- **Tests**
  - Added comprehensive validation to confirm generated nested class files contain the expected marker.
  - Updated generated-output verification fixtures to reflect the corrected annotations.


MagnusSmith:
@coderabbit full review

PR #820, opened 2026-09-08
MaybePath, OptionalPath and AffinePath take a supplier for the error (#794)

### LINKED ISSUE #794: `MaybePath.toEitherPath` and `OptionalPath.toEitherPath` lost the lazy overload their underlying types have
## What

`Maybe.toEither` offers both an eager and a deferred form:

```java
default <L> Either<L, T> toEither(L leftValue)
default <L> Either<L, T> toEither(Supplier<? extends L> leftSupplier)
```

The Path layer over it offers only the eager one:

```java
// MaybePath
public <E> EitherPath<E, A> toEitherPath(E error)

// OptionalPath
public <E> EitherPath<E, A> toEitherPath(E errorIfEmpty)
```

So the error value is always constructed, including on the `Just` / present path where it is thrown away. For a cheap record that is only waste; for an error that formats a message, reads a `MessageSource`, or captures a stack trace, it is waste on the hot path.

`TryPath.toEitherPath(Function<? super Throwable, ? extends E>)` is already deferred, because it has a `Throwable` to map. The `Maybe`/`Optional` side has nothing to map and so needs a `Supplier` to get the same property.

## How it was found

Gating the book under #761. Five pages document the deferred form, which has never compiled:

| Page | What it writes |
|---|---|
| `effect/conversions.md` | a whole "With lazy error" example, and a service example beside it |
| `home.md` | the front page's headline railway |
| `quickstart.md` | "Chain It Together" |
| `effect/effect_path_overview.md` | the chapter's order-processing rewrite |
| `effect/patterns.md` | the service section, twice |

`conversions.md` is the one that names it as a feature ("With lazy error"), and the other four read like it was copied from there. That is the shape of a documented API that was planned and not landed, rather than five independent typos.

## Proposal

Add the deferred overload to both Path types, mirroring `Maybe.toEither`:

```java
public <E> EitherPath<E, A> toEitherPath(Supplier<? extends E> errorSupplier)
```

Both types already delegate to a `Maybe`/`Optional` that has the deferred form, so the implementation is a delegation and the laws are the underlying type's.

**Overload ambiguity is worth checking before committing to the name.** `toEitherPath(E)` with `E` inferred from the argument and `toEitherPath(Supplier<? extends E>)` are distinguishable at a call site that passes a lambda, but `toEitherPath(someSupplierVariable)` where `E` could itself be a `Supplier` is not. `Maybe.toEither` already carries that risk and has lived with it; a `toEitherPathLazy` / `toEitherPathGet` name would avoid it at the cost of symmetry.

## Meanwhile

The #761 sweep rewrites all five pages to the eager form, because that is what the library has today. If this lands, the `conversions.md` "With lazy error" section should come back - it documents a real distinction, and it was the only page teaching it.

MagnusSmith:
Closes #794.

`Maybe.toEither` carries both an eager and a deferred form. The Path layer over it offered only the eager one, so the error value was built on the branch that throws it away. For a plain record that is waste; for an error that formats a message, reads a `MessageSource`, or captures a stack trace, it is waste on the path that succeeds.

`MaybePath`, `OptionalPath` and `AffinePath` now carry `toEitherPath(Supplier<? extends E>)` beside the eager overload. Each delegates to the same branch it already had, so the laws are the underlying type's.

## The ambiguity the issue asked be checked

The issue flagged that `toEitherPath(E)` and `toEitherPath(Supplier<? extends E>)` might not be distinguishable, and offered `toEitherPathLazy` as the fallback name. They are distinguishable, and the symmetric name holds:

```mermaid
flowchart TD
    A["toEitherPath(arg)"] --> B{"What is arg?"}
    B -->|"a lambda () -> e"| C["deferred overload"]
    B -->|"a method reference Err::new"| C
    B -->|"a variable whose type is a Supplier"| C
    B -->|"anything else, a functional-interface error included"| D["eager overload"]
    C --> E["error built only on the absent branch"]
    D --> F["error built either way"]
```

An error type that is *itself* a functional interface is unaffected, because it is not a `Supplier`. The one shape to watch is such an error written **as a lambda**, which reads as a supplier and fails to infer; naming the error type picks the eager overload back: `path.<MyError>toEitherPath(() -> "boom")`. That is the risk `Maybe.toEither` already carries.

## What does not get the overload, and why

`toValidationPath(E, Semigroup<E>)` has exactly the same eager cost, and cannot share the name. Its `Semigroup<E>` parameter pins the same `E` in both candidate signatures, so neither is more specific, and **the bare lambda — the idiomatic way to write a deferred error — is ambiguous**. Existing eager calls still resolve, and a type witness or a hoisted `Supplier` variable rescues the deferred call, so this is narrower than "everything breaks"; it is still disqualifying, because the one form a reader would naturally write does not compile:

```
reference to toValidationPath is ambiguous
  both method <E#1>toValidationPath(E#1,Semigroup<E#1>) in MaybePath
  and  method <E#2>toValidationPath(Supplier<? extends E#2>,Semigroup<E#2>) in MaybePath
```

The blocker is a second parameter that mentions `E`, not arity: `AffinePath.toEitherPath(S, E)` has two parameters and takes the deferred form cleanly, because the overloaded position is the only one that mentions `E`. A deferred `toValidationPath` needs a name of its own; happy to file that separately.

## Behaviour changes

Two call shapes change without a compiler warning:

- `toEitherPath(null)` used to yield `Left(null)` and now selects the supplier overload and throws. Cast the null to the error type (`path.toEitherPath((MyError) null)`) if a null left was meant.
- A `Supplier` passed *as the error itself* now supplies the error rather than becoming it.

Everything else is untouched: the eager overload wins for every ordinary argument, so no existing call site moves.

`OptionalPath.toEitherPath(E)` rejects a null error, so the deferred overload beside it holds a supplied error to the same contract — a supplier returning null fails at the conversion rather than producing a `Left(null)` that NPEs downstream. `MaybePath` permits a null error on both overloads, as its eager form always has.

## Documentation

The issue's "Meanwhile" note: five pages documented the deferred form before it existed and were rewritten to the eager form under #761. [Type Conversions](https://github.com/higher-kinded-j/higher-kinded-j/blob/fix/794-lazy-toeitherpath/hkj-book/src/effect/conversions.md) gets its "With lazy error" example back, since it was the page teaching the distinction, and the `OptionalPath` section beside it gains the same line. The other four keep the eager form, which reads better for the cheap records they use.

Also swept: the cheatsheet conversion table, a 0.4.11 release-history entry, `CrossPathConversionsExample`, and three skill pages that told an assistant `toEitherPath` takes a value "not a supplier" — `hkj-guide/SKILL.md`, `hkj-guide/reference/patterns.md` (with one new gated snippet) and `hkj-bridge/SKILL.md`.

## Verification

- `:hkj-core:clean check` and `:hkj-examples:bookVerify` — **BUILD SUCCESSFUL**, exit 0.
- JaCoCo: 0 missed instructions, branches and lines on all three new methods, their generated lambdas, and the new `suppliedError` guard.
- 9 tests in `CrossPathConversionTest.DeferredErrorConversions` and 3 in `AffinePathBridgeTest.ToEitherPathTests` covering supplier-not-called on success, called-once on absence, overload selection by argument shape, and the null guards.
- Snippet floors raised 2435 → 2436 and 214 → 215 for the added gated snippet.

https://claude.ai/code/session_01B9gn887ZTkPfZmRhr6dUCF

## Summary by CodeRabbit

- **New Features**
  - Added deferred error construction for `MaybePath`, `OptionalPath`, and `AffinePath` conversions to `EitherPath`.
  - Error suppliers are invoked only when the path cannot produce a value, avoiding unnecessary work on successful conversions.
  - Added validation for null suppliers and null errors produced by suppliers.

- **Documentation**
  - Updated guides, examples, cheat sheets, and release history to explain the new overloads and usage patterns.

- **Tests**
  - Added coverage for lazy invocation, overload selection, and null handling.


MagnusSmith:
@coderabbit full review

MagnusSmith:
Thanks — all three actioned in e66f8440. Two are good catches; the second needed measuring, and the result corrects both of us.

**`toValidationPath` rationale (release-history.md) — you are right that my claim was wrong, though not for the reason given.**

I said the overload would be "ambiguous at every call site". It is not. But `() -> "missing"` does not select the supplier overload either — it is the one shape that *is* ambiguous. Compiled against the real `Semigroup`, with the two overloads declared:

| call shape | result |
|---|---|
| `toValidationPath("missing", sg)` | eager, resolves cleanly |
| `toValidationPath(() -> "missing", sg)` | **ambiguous** |
| `p.<String>toValidationPath(() -> "missing", sg)` | deferred — the witness rescues it |
| `toValidationPath(supplierVariable, sg)` | deferred |

So existing eager calls are fine, and the deferred call is reachable with a type witness or a hoisted local. Narrower than I claimed, still disqualifying: the one form a reader would naturally write does not compile, and requiring a witness for the common case is a worse method than a differently named one. The release note, this PR and #821 now say that instead.

**Bare `null` (MaybePath javadoc) — correct, and it was a real gap.**

The release note documented it; the javadocs and Type Conversions did not, which is precisely where a reader meeting the eager form’s permissive contract would be surprised. Added to `MaybePath`, `OptionalPath` and `AffinePath`, and to the book. One correction to the suggested wording: the recovery is a **cast**, `path.toEitherPath((MyError) null)` — a type witness still selects the deferred overload (`path.<MyError>toEitherPath(null)` throws), so the docs name the cast specifically.

**hkj-guide patterns reference — fair.**

That page is read by an assistant generating code, so an incomplete selection rule there produces calls that do not compile. It now names both shapes that need the error type spelled out: a functional-interface error written as a lambda, and a bare `null`.

`:hkj-core:check` and `:hkj-examples:bookVerify` green, exit 0.

PR #828, opened 2026-09-11
@GenerateMapping spreads a nested domain record across a flat wire (#674)

### LINKED ISSUE #674: [FEAT] @GenerateMapping: flatten/unflatten a nested domain component against a flat wire
## The gap

MapStruct maps `@Mapping(source = "address.city", target = "city")` in both directions, auto-creating intermediate objects when unflattening. HKJ has no answer: `@MapField(to = ...)` is a flat wire name, a derived `Getter` can flatten outbound but `parse` ignores derived fields by design, and a domain component with no wire counterpart is a hard error. Bidirectional flattening is therefore impossible today, and the implicit fix ("restructure the wire to nest") fails exactly when the wire format is externally fixed. This blocks a large class of real mappers.

A nested domain component declared as fed by a group of flat wire components, both directions, with the doctrine intact:

```java
public record Customer(String name, Address address) {}
public record Address(String street, String city, String postcode) {}
// externally fixed flat wire:
public record CustomerDto(String name, String street, String city, String postcode) {}

@GenerateMapping
public interface CustomerMapping extends MappingSpec<Customer, CustomerDto> {
  @FlattenedInto({"street", "city", "postcode"})   // strawman syntax, see below
  Address address();
}
```

- `build` flattens: `dto.street = domain.address().street()` and so on.
- `parse` unflattens: the flat fields assemble the `Address` through `Validated.fields()`, so failures locate as `address.street: must not be null` — the flat wire gains the domain's nested paths, consistent with the domain-named-paths contract.
- Nesting composes: the assembled component may itself route through leaves for its fields.

- **Declaration syntax.** Candidates: (a) an annotation on an abstract marker method naming the wire fields (strawman above); (b) `@MapField(to = "street", of = "address.street")` per flat field; (c) a prefix convention (`address_street`) with an opt-in. Weigh explicitness vs ceremony; the collision sweep and rename machinery must compose.
- **Name mapping inside the group.** Same-named fields map positionally-by-name (street↔street); renamed inner fields need a story.
- **Arity interaction.** The assembled component consumes its own `fields()` ladder; interaction with the parse arity ceiling (see the parse-arity draft) needs stating.
- **Location grammar.** Errors locate under the domain path (`address.city`), which for a flat wire means paths that name fields the client never sent grouped under a prefix it never sent; extend the domain-named-paths documentation honestly.
- **Depth.** One level first; recursive flattening (`address.geo.lat`) only if the design falls out naturally.
- **Projection/patch/update tiers.** Decide per tier whether flattened groups participate or are rejected initially.

## Dependencies / scope

- Depends on: nothing hard; benefits from the parse-arity work (#670) landing first.
- In: design document with the syntax decision, then implementation for the full record-record tier (build + parse), diagnostics (unknown wire field in group, overlapping groups, group vs rename collisions), laws, goldens, book section.
- Out (first cut): bean wires, generic specs, sparse tier, recursive depth beyond one level.

MagnusSmith:
Closes #674.

A nested domain record can now map to a flat wire, both directions. `Customer` holds an `Address`; the wire, fixed by someone else, carries `street`, `city` and `postcode` as plain fields. No single wire component carries the address, so a leaf could not map it and the pair was a hard error. A `@Flatten` marker named after the domain component declares the spread:

```java
public record Address(String street, String city, String postcode) {}
public record Customer(String name, Address address) {}
public record CustomerDto(String name, String street, String city, String postcode) {}

@GenerateMapping
public interface CustomerMapping extends MappingSpec<Customer, CustomerDto> {
  @Flatten
  Address address();
}
```

```java
// generated
public CustomerDto build(Customer domain) {
  return new CustomerDto(domain.name(), domain.address().street(), domain.address().city(), domain.address().postcode());
}
public Validated<NonEmptyList<FieldError>, Customer> parse(CustomerDto wire) {
  return Validated.fields()
      .field("name", hkj$ifPresent(wire.name(), Validated::validNel))
      .field("address", Validated.fields()
        .field("street", hkj$ifPresent(wire.street(), Validated::validNel))
        .field("city", hkj$ifPresent(wire.city(), Validated::validNel))
        .field("postcode", hkj$ifPresent(wire.postcode(), Validated::validNel))
        .apply(Address::new))
      .apply(Customer::new);
}
// parse(new CustomerDto("Ada", null, "Leeds", null))
// Invalid(NonEmptyList[address.street: must not be null, address.postcode: must not be null])
```

## Design

The issue left the syntax open between an annotation naming the wire fields, per-field `@MapField(to, of)` paths, and a prefix convention. This takes the marker shape the spec vocabulary already has (`@MapField`, `@OptionalBridge`): an abstract method named after the domain component, stubbed by the Impl, spread **by name**. The record's components (the group) then take the whole existing vocabulary by their own names, with no new attribute grammar:

- `@MapField(to = "addressLine1") String street();` renames a group member;
- `default ValidatedPrism<String, Postcode> postcode()` converts one, and makes the mapping fallible exactly as a top-level leaf would;
- `@OptionalBridge Optional<String> fax();` bridges a member that is `Optional`, declared locally or inherited;
- a member that is itself a record nests through its own spec; containers lift.

```mermaid
flowchart LR
    D["Customer<br/>name, address: Address(street, city, postcode)"] -->|"build: domain.address().street() ..."| W["CustomerDto<br/>name, street, city, postcode"]
    W -->|"parse: nested fields() ladder"| D
    W -.->|"failure locates as<br/>address.street"| E["located FieldError"]
```

Inside the processor there is no new correspondence kind. `Correspondence` gained a nullable `Group` (name + record type); `classify` classifies each group member against the wire exactly as a top-level component, through the extracted `classifyComponent`, tagged with its group, and the members stay contiguous. `parseLegs` folds a group's run into one leg carrying its own `fields()` ladder, which is what gives the domain-named paths for free; `reverseArgs` rebuilds the record for `asIso()`; `domainRead` reads through the group for `build`. `collectFlattened` runs before renames, leaves and bridges are collected, since a group's member names join what those may bind to; the projection and component-count checks count wire slots rather than domain components.

The marker restates the record it spreads, and its return type must be the component's own type, so a spec that drifts from its domain fails to compile rather than spreading whatever the component has become (the bridge marker's rule). The annotation is retained in the class file: the classpath index decides whether a dependency's spec is parse-capable by counting the wire components its domain calls for, and a flattened component calls for one per member. Without the retention a flattening spec in a dependency would register as a projection and nothing could nest it.

Rules, each a compile error naming both sides: a group member may not share a name with a domain component or with another group's member, so two components of one record type cannot both be spread (the wire component would need two sources); a derived field may not be named after a member; a rename may not be named after the flattened component itself, whether on the same method or inherited beside the marker; a wire component named after the flattened component must be fed by a rename from another component; a marker cannot combine with `@MapField` or `@OptionalBridge`, cannot be leaf-shaped, needs no parameters and no body; a local marker naming no component is refused with a did-you-mean, an inherited one stays inert (the bridge precedent). Diagnostics inside a group say where the member came from: a missing counterpart reads `domain field 'Address.postcode' (spread from 'address') has no wire counterpart named 'postcode'`, a collision names the domain paths (`'name' and 'address.street'`), and the too-many-components error lists the groups it counted and offers `@Flatten` for a record component with no counterpart. An all-identity group keeps the mapping lossless, so `asIso()` survives and reassembles the record.

Not supported yet, each refused with a diagnostic: bean-shaped wires, generic specs, projections (a wire with fewer components), sparse `UpdateSpec`s, a record with no components, a raw or wildcard-carrying record type (the group is reassembled as `new Address(...)`, which such a type cannot be written into), a group wider than one `fields()` ladder (the chunked emitter writes statements, not an expression, so nested chunking is a follow-up), and a second level of spreading (a marker naming a group member is refused; a record inside the group nests through its own spec).

## Tests

`MappingProcessorFlattenTest` (44 cases): emission for identity, leaf, rename, nested spec, inner bridge (local and inherited), containers inside the group, a flattened first component, two groups, a rename feeding the wire component named after the group, a generic domain instantiated concretely, and a flattening spec nested by another in the same compilation; agreeing mix-in markers; `asIso()` reassembly; a runtime round trip with located failures; and every refusal above. `MappingProcessorClasspathTest` nests a flattening spec from a dependency and runs the composed parse. `GeneratedMappingLawsTest` law-checks an all-identity group's `asIso()` and a leaf-carrying group's `asValidatedPrism()` through `MappingLaws`. A `FlattenedCustomerMappingImpl` golden pins the generated text (two groups, a rename and a leaf inside). `MappingProcessor*` stays at 100% line, branch and instruction coverage.

## Docs

`mapping/structure.md` gains "Flattening a nested component onto a flat wire" directly after the nesting section, including a compiled and run example from `RecordMappingBook` with its output; the limits table in `testing.md` has the row; `basics.md`, `codecs.md`, the annotations-at-a-glance table and the glossary mention the marker where they list the vocabulary; the hkj-mapping skill has a section with a verified snippet and a table row; `@GenerateMapping`, `@MapField` and the new `@Flatten` javadocs describe it; the release history has an entry. The chapter introduction now credits MapStruct with deep path flattening rather than nested-path flattening, since one level is covered here. In passing, two annotation javadocs lost their issue-number provenance.

## Summary by CodeRabbit

- **New Features**
  - Added `@Flatten` support for spreading nested domain records across flat wire fields.
  - Supports bidirectional mapping, field renames, conversions, optional bridges, nested mappings, and containers.
  - Validation errors from flattened fields are accumulated and reported using nested domain paths.

- **Documentation**
  - Added guidance, examples, glossary entries, release history, supported limitations, and diagnostics for flattened mappings.

- **Tests**
  - Added coverage for round trips, validation, nested mappings, naming conflicts, unsupported configurations, and generated output.


MagnusSmith:
@coderabbit full review

PR #833, opened 2026-09-12
@GenerateMapping lifts Set, array and Map-key components (#675)

### LINKED ISSUE #675: [FEAT] @GenerateMapping: widen container lifting — Set and array elements, Map keys
## The gap

Element lifting covers `List`, `Optional` and `Map` values only. MapStruct additionally maps `Set`, arrays, `Iterable`, `Stream` elements and `Map` keys. In HKJ a `Set<EmailAddress>` component maps only as whole-value identity or through a hand-written whole-container leaf, and differing map key types are a hard error ("keys are identity"). The lifting HKJ does have is the best-in-class version (indexed and keyed error location); this issue widens the type coverage without weakening that.

- `Set<E>` and `E[]` components lift through an element leaf or nested spec exactly as `List<E>` does, both directions.
- `Map<K, V>` components may declare a key leaf alongside the value leaf, so `Map<String, User>` ↔ `Map<UUID, UserDto>` maps with both sides converted.
- Failures stay located and accumulating.

- **Set location grammar.** A set has no index. Candidates: locate by the failing source element's rendered value (`emails.{nope}: not an email address`), by encounter order of the iteration (documented as unstable), or unlocated-within-the-component (`emails: ...`, accumulating). This is the design decision of the issue; the map-key grammar precedent (locate by key) suggests locating by source value, with the dotted-path ambiguity caveat extended to it.
- **Set semantics.** Element mapping can collapse cardinality (two sources mapping to one target); decide silently-collapse vs located error.
- **Arrays.** Component arrays (`String[]`) lift with index location like lists; primitive arrays (`int[]`) have no element mapping to do and stay identity (elements cannot be null either).
- **Map keys.** Unlocks the current keys-identity rule. A failing key locates as what? (The value grammar locates under the key; a failing key has no stable name — likely locate by the source key's rendering.) The identity-keys rule was a deliberate simplification, not a doctrine; revisiting it needs the location story first.
- **Explicitly out.** `Stream` components (a stream in a record is a smell; document instead), `Iterable`/`Collection` supertypes (require a construction policy; `List`/`Set` cover the real cases), `@AdderStrategy`-style accumulation (bean-wire adders are a separate bean-shape concern).
- **Identity scans.** The identity-container null scan (`hkj$allPresent`/`hkj$valuesPresent`) extends to `Set`/array identity components in the same change, keeping the doctrine uniform.

## Dependencies / scope

- Depends on: nothing hard; the ValidatedPrism bulk forms likely grow `parseAll(Set)`/array overloads or an internal shared shape.
- In: Set + array element lifting, map-key leaves, location-grammar decision documented in the book's doctrine paragraph, laws, goldens, diagnostics for the newly expressible mistakes.
- Out: Stream/Iterable, adders, deep/nested container combinations beyond the existing one-level lifting.

MagnusSmith:
Closes #675.

Element lifting covered `List`, `Optional` and a `Map`'s **values**. A `Set<EmailAddress>` component could only be copied whole or mapped through a hand-written whole-container leaf, an array could not lift at all, and a `Map` whose two sides keyed differently was a hard error. All three lift now, in both directions and on every tier — dense, projection, bean wire and sparse `UpdateSpec` — and a same-typed `Set` or reference array carries the identity null scan the other containers already did.

The lifting HKJ already had is the better version, because every failure is located and accumulating. That is the part this had to widen without weakening.

## What each container locates by

```mermaid
flowchart LR
    L["List&lt;E&gt;, E[]"] --> LI["by INDEX<br/><code>emails.1</code>"]
    S["Set&lt;E&gt;"] --> SE["by the ELEMENT's rendering<br/><code>emails.nope</code>"]
    M["Map&lt;K, V&gt;"] --> MK["by the SOURCE key<br/><code>attributes.en.email</code>"]

    classDef c fill:#a6d189,stroke:#40a02b,color:#232634
    classDef l fill:#8caaee,stroke:#1e66f5,color:#232634
    class L,S,M c
    class LI,SE,MK l
```

A set has no index, and its iteration order is not part of its contract — numbering would name a *different* element on the next run. The value is what identifies the element, so that is what the path says, extending the caveat the map-key grammar already carried. A `null` set element has no rendering at all, and a set holds at most one, so it reports unlocated as `emails: must not contain a null element`, kept distinct from `must not be null`, which says the set itself is absent.

`Kind.LIST` became `Kind.ELEMENTS` and now covers `Set` as well, with no new emission arm: `ValidatedPrism`'s bulk forms are overloaded on the container, so the component's own type picks `parseAll(Set)` over `parseAll(List)`. Arrays needed their own kind, because only they pass an array constructor — a generic array cannot be created otherwise.

## Map keys

```java
@GenerateMapping
public interface ProfileMapping extends MappingSpec<Profile, ProfileDto> {

  // values, named after the component, as before
  default ValidatedPrism<UserDto, User> attributes() { return UserCodecs.user(); }

  // keys: the annotation names the component, because the value leaf already owns that name
  @MapKey("attributes")
  default ValidatedPrism<String, Locale> attributeLocale() { return Codecs.locale(); }
}
```

Either side may convert alone. A failing key locates by the **source** key, so the path names what the caller sent rather than what it parsed to, and an entry wrong on both sides reports both reasons there. Without a key leaf the key types must still match, and that refusal now offers a pasteable `@MapKey` declaration instead of telling you to restructure.

`ValidatedPrism` gained the matching bulk forms: `parseAll`/`buildAll` over a `Set` and over arrays, plus `parseKeys`/`parseEntries` and their total twins.

## Two decisions worth naming

**Collapse.** A collapse needs a non-injective leaf, which already breaks the `ValidatedPrism` section law — so neither case below comes from a lawful leaf, and neither reaches `asIso()`. Where one happens, the containers differ because the *loss* differs: collapsed set elements are equal to each other, so the set still holds everything it was given (silent); two map keys parsing to one key discard a whole entry whose value need not match (a located `duplicates an earlier key`).

**Same container both sides.** A `List` against a `Set`, or an array against a `List`, stays a plain type mismatch rather than silently changing what the collection promises.

## Limits, each a diagnostic

- A lifted array's element must be able to name its own constructor. Java forbids generic array creation, so a type variable or a parameterised element (`T[]`, `List<Tag>[]`) is refused — otherwise the generated Impl would carry `T[]::new` and fail to compile in the user's build, where they could not fix it. Raw and unbounded-wildcard elements are reifiable and still lift.
- A primitive array is copied whole: a `ValidatedPrism` cannot focus a primitive, and a primitive element cannot be null.
- A record carrying an array component has identity `equals`, so `MappingLaws` cannot law-check it; the suite asserts that round trip elementwise instead.

## Review

The standing five-lens panel ran before this PR and its findings are folded into the last commit — including a silently-dropped leaf on a same-typed unnameable array, a `@MapKey` that converted the wrong types and was ignored rather than reported, a `buildKeys` that could render a wire its own `parseKeys` rejects, and `@GenerateMerge` not scanning `Set`/array fills, which would have left the located-null doctrine differing between the two processors.

## Verification

`:hkj-processor:check` green at its 100% line/branch/instruction pins, `:hkj-core:check` green, `:hkj-examples:bookVerify` green, full `build` green. A new golden pins every new emission in one Impl, and the law suite covers the `Set` and converting-key tiers.

One source-compatibility note, from `parseAll`/`buildAll` becoming overloaded: a bare `null` literal now needs the type it stands for, and `prism::parseAll` is now *inexact*, so where nothing else fixes the container type it needs a target type or a lambda. Both are called out in the release entry.

## Summary by CodeRabbit

- **New Features**
  - Extended mapping support to lift elements in `Set` and arrays.
  - Added conversion of map keys through the new `@MapKey` annotation.
  - Added bulk validation and construction for sets, arrays, map keys, and map entries.
  - Improved error locations for collection indexes, set elements, and map keys.
  - Added duplicate-key detection when converted map keys collide.
  - Added handling and diagnostics for null collection elements and container mismatches.

- **Documentation**
  - Expanded mapping guides, examples, API references, and release history for the new container and map-key capabilities.


MagnusSmith:
@coderabbit full review

Issue #673, opened 2026-07-28
[FEAT] @GenerateMapping: the Optional bridge for record wires


MagnusSmith:
## The gap

A domain `Optional<String> nickname` against a bean wire's nullable `String nickname` bridges
automatically (null ↔ empty, the documented Optional bridge). The same shape against a **record**
wire does not: the bridge is bean-only, so with a leaf in place `nickname = null` parses to a
located `nickname: must not be null` while `build` happily emits null — an asymmetric round trip
the comparator hit immediately, because real-world record DTOs use nullable reference fields for
optional data (Jackson's default), not `Optional` components. The current answer (declare the
wire component as `Optional<String>`) changes the wire contract.

## What "good" looks like (user perspective)

The bean bridge's semantics, available on record wires by opt-in (never silently, since on a
record wire null-is-an-error is the doctrine's default):

```java
@GenerateMapping
public interface CustomerMapping extends MappingSpec<Customer, CustomerDto> {
  @OptionalBridge   // strawman: opt-in per component, name/shape to be decided
  Optional<String> nickname();
}
// parse: null -> Valid(Optional.empty()); value -> Valid(Optional.of(v)) or through the leaf
// build: empty -> null; present -> value
```

## Things to explore / constraints

- **Opt-in shape.** A marker on an abstract component method (strawman above), a `@MapField`
  attribute, or recognising the pattern implicitly. Implicit conflicts with the null doctrine
  (null on a record wire is an error by default and that must stay the default); the review's
  position is explicit opt-in only.
- **Doctrine wording.** The book's "a record cannot express absence, it can only be wrong"
  sentence needs the carve-out: an opted-in bridged component expresses absence deliberately.
- **Symmetry law.** `build(domain with empty) == wire with null` and the round trip; extends the
  existing bean-bridge behaviour, so laws and goldens mirror it.
- **Interaction with leaves.** A bridged component with an element leaf validates the present
  value through the leaf (matching the bean bridge's leaf handling).
- **Sparse tier.** Unaffected (bean wires only), but the rules list should cross-reference.

## Dependencies / scope

- Depends on: nothing.
- In: opt-in marker + record-wire bridge emission (parse and build), diagnostics (bridge on a
  non-Optional domain component; bridge on a bean wire is redundant and should say so), laws,
  goldens, book updates.
- Out: tri-state semantics; `Optional` wire components (already work today).

PR #834, opened 2026-09-12
An inherited derived field is inert on a sparse UpdateSpec (#826)

### LINKED ISSUE #826: [BUG] A derived-field mix-in cannot serve a MappingSpec and its UpdateSpec sibling
A mix-in carrying a derived field cannot be shared between a full `MappingSpec` and its PATCH sibling: the `UpdateSpec` refuses it, and reports the refusal at the mix-in, where the fix it prescribes would break the full spec.

This is the defect [#673](https://github.com/higher-kinded-j/higher-kinded-j/issues/673) fixed for `@OptionalBridge` (`checkNoBridges` now skips inherited members), left behind in the sibling check `checkNoDerivedFields`, which still scans every member including inherited ones.

## Measured

On `main` (`7d3c852fb`):

```java
interface UserVocabulary {
  default Getter<User, String> display() {
    return Getter.of(u -> "user " + u.name());
  }
}

@GenerateMapping
interface UserMapping extends UserVocabulary, MappingSpec<User, UserDto> {}

@GenerateMapping
interface UserPatchMapping extends UserVocabulary, UpdateSpec<User, UserPatchDto> {}
```

```
error: @GenerateMapping: the derived-field method 'display' has no meaning on a sparse
UpdateSpec. A derived field feeds build(); a sparse update only writes present wire properties
into the domain, so there is nothing to derive. Remove the method, or use a full MappingSpec if
you need a total build().
```

Both prescribed fixes are wrong for the shape:

- *Remove the method* breaks `UserMapping`, whose wire has a `display` component with no domain source.
- *Use a full MappingSpec* discards the sparse semantics the PATCH spec exists for.

The only escape is duplicating the vocabulary into two interfaces.

## Why the promise is explicit

[Standard Codecs and Shared Vocabulary](https://higher-kinded-j.github.io/mapping/codecs.html#shared-vocabulary-mix-in-interfaces) states it:

> `UpdateSpec` mappings inherit vocabulary the same way, element leaves included, so the leaf a full spec lifts over a `List` serves its PATCH sibling unchanged.

and its Key Takeaway: *one plain interface serves every spec, PATCH siblings included*.

The same page now enumerates the inheritable vocabulary as "renames, leaves, derived fields **and** `@OptionalBridge` markers". Three of those four are inert when they cannot apply. Derived fields are the exception.

## Why

Every other rule that refuses vocabulary it cannot use splits local from inherited. `checkNoDerivedFields` does not:

| check | locally declared | inherited |
|---|---|---|
| `checkLocalLeavesBind` | error | **inert** |
| `checkNoSealedVocabulary` | error | **inert** |
| `checkBridgesApply` (unmatched) | error | **inert** |
| `checkNoBridges` (sparse) | error | **inert** (fixed in #673) |
| `checkNoDerivedFields` (sparse) | error | **error** |

```mermaid
flowchart TD
    V["UserVocabulary:<br/>default Getter&lt;User,String&gt; display()"] --> M["UserMapping<br/>(MappingSpec)"]
    V --> P["UserPatchMapping<br/>(UpdateSpec)"]
    M --> OK(["build fills 'display'"])
    P --> X(["refused at the mix-in,<br/>with no fix that keeps both"])

    class V,M,P wire
    class OK tier
    class X error
```

## What "good" looks like

`checkNoDerivedFields` refuses only a **locally declared** derived field, matching its four siblings. An inherited one stays inert: the sparse tier never consults it, so nothing is generated differently and nothing needs saying.

The local refusal keeps its current message, which is right for a spec that wrote the method itself.

## Scope

- In: the local/inherited split in `checkNoDerivedFields`, a regression test that one vocabulary serves a `MappingSpec` and an `UpdateSpec`, and a line in the book confirming the promise now holds for all four vocabulary kinds.
- Out: any change to what a derived field *means* on either tier. It stays a `build`-only concept.

## Note

Found by the pre-PR review panel on [#824](https://github.com/higher-kinded-j/higher-kinded-j/pull/824) while fixing the `@OptionalBridge` half, and deliberately left alone there as out of scope for #673.

MagnusSmith:
Closes #826.

A mix-in carrying a derived field could not be shared between a full `MappingSpec` and its PATCH sibling. `checkNoDerivedFields` scanned every member, inherited ones included, so the `UpdateSpec` refused the method at the mix-in that declared it, and both fixes the diagnostic prescribed were wrong for the shape: removing the method breaks the full spec whose wire has the derived component, and switching to a full `MappingSpec` discards the sparse semantics the PATCH spec exists for. The only escape was duplicating the vocabulary into two interfaces.

## The rule family

A spec's vocabulary is matched against one of two things, and that decides whether an unmatched inherited member can be ignored:

```mermaid
flowchart TD
    M["inherited vocabulary member"] --> Q{"matched against what?"}
    Q -->|the domain| D["leaf, @OptionalBridge, @Flatten<br/>shared by a spec and its PATCH sibling"]
    Q -->|the wire| W["@MapField rename<br/>NOT shared: wires differ"]
    D --> I(["inert when it binds to nothing"])
    W --> R(["must name a component of every<br/>extending spec's wire"])

    class M,D,W wire
    class I tier
    class R error
```

`checkNoDerivedFields` now sits on the left with its siblings: only a **locally declared** derived field is refused, which is right for a spec that wrote the method itself. An inherited one stays inert, since the sparse tier emits no `build()` to consult it. This is the split #673 gave `checkNoBridges`, left behind in the sibling check.

| check | locally declared | inherited |
|---|---|---|
| `checkLocalLeavesBind` | error | inert |
| `checkNoSealedVocabulary` | error | inert |
| `checkBridgesApply` (unmatched) | error | inert |
| `checkNoBridges` (sparse) | error | inert |
| `checkNoDerivedFields` (sparse) | error | **inert** (was: error) |

## Also in this change

Both local refusals in the family now name the escape hatch this opens: moving the member to a mix-in a full `MappingSpec` also extends, which is the one remedy that keeps both specs working. `checkNoBridges` drops an `inheritedNote(...)` call that has been dead since #673 gave that check its local guard, and which misleadingly suggested the message could still name a mix-in. `checkNoDerivedFields` takes the sibling rules' guard form, testing the cheap owner comparison before resolving the return type.

**A documentation correction the review turned up.** The book's shared-vocabulary promise needed a caveat, and the obvious one was wrong. `@Flatten` is not the only inherited member still refused: a `@MapField` rename is refused too, and on the full tier as well as the sparse one, because it alone is matched against the *wire* rather than the domain. A vocabulary shared with a projection or a PATCH bean therefore renames only what they all carry. `codecs.md`, `beans_patch.md` and the mapping skill now say so. `testing.md`'s limits row and the skill's PATCH section also drop the blanket claim that `@OptionalBridge` is meaningless on a sparse spec, stale since #673.

## Testing

`MappingProcessorTest.SpecInheritance.oneVocabularyServesBothTiers` drives the existing `AccountVocabulary` fixture, a rename plus a leaf plus a `display()` derived field, through both an `AccountMapping extends MappingSpec` and a new `AccountPatchMapping extends UpdateSpec`. The full spec still emits `display().get(domain)`; the PATCH sibling folds the rename and the leaf and never mentions it. The local half stays pinned by the existing `MappingProcessorUpdateTest.derivedFieldRejected`, so both arms of the new condition are exercised and `MappingProcessor*` holds its 100% line/branch/instruction pin.

`gradle build` green.

## Follow-ups

Three adjacent defects the pre-PR review panel surfaced, each measured, each filed separately rather than widened into this change: the unguarded rename above, `@Flatten`'s own mix-in sharing refusal (where inertness would trade a pointed diagnostic for a vaguer one, so it needs a design call), and a spec extending both `MappingSpec` and `UpdateSpec`, which registers as nestable but emits only `updateFrom`.

## Summary by CodeRabbit

- **New Features**
  - PATCH mappings can now share vocabulary with full mappings while safely ignoring inherited derived fields and optional bridges.
  - Inherited renames are validated against the properties available on each mapping, including PATCH and projection types.

- **Bug Fixes**
  - Improved validation distinguishes inherited configuration from declarations made directly on sparse PATCH specifications.
  - Invalid local derived fields, optional bridges, flatten markers, and unsupported renames continue to be rejected.

- **Documentation**
  - Updated mapping, mix-in, PATCH, and diagnostics guidance to reflect these rules.


MagnusSmith:
@coderabbit full review

Issue #830, opened 2026-09-11
[BUG] @GenerateMapping: a raw List/Map identity component generates uncompilable code, and a JAXB list bridged from Optional turns absent into empty


MagnusSmith:
Two defects in `@GenerateMapping`'s dense tiers (full mappings and projections, `build`/`parse`/`patch`). Both were found while reviewing #829 and both reproduce on `main` for full mappings. #829 makes them newly reachable on bean projections, which used to be refused outright.

## 1. A raw `List` or `Map` identity component generates code that does not compile

A component declared with a raw `List` or `Map` on both sides copies by identity. The generated Impl fails to compile, with the error pointing into the generated file rather than at the spec.

```java
@SuppressWarnings("rawtypes")
public record Raw(String id, List items) {}
@SuppressWarnings("rawtypes")
public record RawDto(String id, List items) {}

@GenerateMapping
public interface RawMapping extends MappingSpec<Raw, RawDto> {}
```

```text
RawMappingImpl.java:35: error: incompatible types: invalid constructor reference
```

The leg the processor emits is the null-scanning one, `.field("items", hkj$allPresent(wire.items()))`. The helper is generic, `<E> hkj$allPresent(List<E>)`, so a raw argument makes the call unchecked and its result erased, and the ladder's `.apply(Raw::new)` no longer types. With `-Xlint:unchecked` there are also four warnings in the generated file.

Every dense shape fails the same way:

| Shape | Result |
|---|---|
| full mapping, record wire, raw `List` | `invalid constructor reference` |
| full mapping, record wire, raw `Map` | `invalid constructor reference` |
| full mapping, bean wire, raw `List` | `invalid constructor reference` |
| projection, bean wire, raw `List` | `Object cannot be converted to List` |
| sparse `UpdateSpec`, raw `List` | compiles cleanly |

**Cause.** `identityKind` classifies any `java.util.List` or `java.util.Map` as `IDENTITY_LIST`/`IDENTITY_MAP`, raw included. The sparse tier already guards this: `sparseIdentityKind` falls back to plain `IDENTITY` when the container is raw or carries a wildcard argument, because the helper cannot type either. The dense tiers never got the same check. A wildcard argument happens to compile on the dense tiers, so only raw containers fail there.

```mermaid
flowchart LR
    C["identity List/Map component"] --> T{"type arguments<br/>proper types?"}
[...]

PR #840, opened 2026-09-12
Name the element type where inference would capture it (#838, #839)

### LINKED ISSUE #838: [BUG] @GenerateMapping: a wildcard-carrying domain element under an Optional bridge generates uncompilable code
A domain `Optional<T>` whose **element type carries a wildcard argument** generates an Impl that does not compile when the component bridges. The error points into generated source, and on a bean wire the shape needs no annotation at all, because the bridge is automatic there.

Found while fixing #830; this is a third defect in the same family — a generated leg whose type the emitted code cannot express — reproducing on `main` and untouched by #832.

## Repro

A bean wire, so the bridge is taken automatically:

```java
public record Bookmarks(String owner, Optional<List<? extends CharSequence>> urls) {}

public class BookmarksDto {
  private String owner;
  private List<? extends CharSequence> urls;
  public String getOwner() { return owner; }
  public void setOwner(String owner) { this.owner = owner; }
  public List<? extends CharSequence> getUrls() { return urls; }
  public void setUrls(List<? extends CharSequence> urls) { this.urls = urls; }
}

@GenerateMapping
public interface BookmarksMapping extends MappingSpec<Bookmarks, BookmarksDto> {}
```

```text
BookmarksMappingImpl.java:36: error: incompatible types: invalid constructor reference
        .apply(Bookmarks::new);
               ^
    incompatible types: Optional<List<CAP#1>> cannot be converted to Optional<List<? extends CharSequence>>
  where CAP#1 is a fresh type-variable:
    CAP#1 extends CharSequence from capture of ? extends CharSequence
```

A record wire opting in with `@OptionalBridge` fails identically.

**The domain element is what triggers it, not the wire type, and not the container.** Two probes narrow it:

- A non-container element fails the same way: `Optional<Holder<? extends Number>>` against a `Holder<? extends Number>` property gives `Optional<Holder<CAP#1>> cannot be converted to Optional<Holder<? extends Number>>`. So this is not the identity-container rule.
- A **plain `String` wire property** mapped through a leaf `ValidatedPrism<String, List<? extends CharSequence>>` into a domain `Optional<List<? extends CharSequence>>` fails too, with the same message. No wildcard appears on the wire side at all.
- The mirror case compiles: a wildcard-carrying *wire* property read through a leaf into a **non-wildcard** domain element (`Optional<Tag>`) is fine, because nothing puts a capture inside the `Optional`.

So: the bridge breaks exactly when the domain `Optional`'s element type carries a wildcard, on both the identity branch and the leaf branch.

## Cause

The bridged leg has to produce the domain component's `Optional<E>`, and it builds that `Optional` with a generic call whose type argument is *inferred* rather than named:

```java
// identity branch
.field("urls", Validated.validNel(Optional.ofNullable(wire.getUrls())))

// leaf branch
.field("urls", Optional.ofNullable(wire.getUrls())
    .map(v -> urls().parse(v).map(Optional::of))
    .orElseGet(() -> Validated.validNel(Optional.empty())))
```

Inference captures the wildcard, so the leg's type becomes `Optional<List<CAP#1>>`. `Optional<T>` is invariant, so that is not `Optional<List<? extends CharSequence>>`, and the ladder's `.apply(Bookmarks::new)` no longer types.

Nothing absorbs the capture the way a *component* absorbs it on an unbridged identity leg — there the field type is `List<CAP#1>`, and `List<CAP#1>` **is** a `List<? extends CharSequence>`, which is why an unbridged wildcard container compiles today. Wrapping it in an invariant `Optional` is what closes that door.

```mermaid
flowchart LR
    R["read for a wildcard-carrying element"] --> B{"bridged?"}
    B -->|no| U["field type List&lt;CAP#1&gt;"]
    U --> V{"D::new wants<br/>List&lt;? extends CharSequence&gt;"}
    V -->|"the wildcard absorbs the capture"| K["compiles today"]
    B -->|yes| O["field type Optional&lt;List&lt;CAP#1&gt;&gt;"]
    O --> A{"D::new wants<br/>Optional&lt;List&lt;? extends CharSequence&gt;&gt;"}
    A -->|"Optional is invariant"| X["does not compile"]
    class K ok
    class X bad
    class A,B,V decision
```

Only `parse` is affected. `build` writes `domain.urls().ifPresent(v -> wire.setUrls(v))` on a bean and `domain.urls().orElse(null)` on a record, neither of which infers anything.

The third branch, the null-scanning one, is already withheld for a wildcard container, so it needs nothing.

## Suggested fix

Name the type instead of letting it be inferred, on both branches, with the bridged element type the correspondence already resolves:

```java
// identity branch
Validated.validNel(Optional.<List<? extends CharSequence>>ofNullable(wire.getUrls()))

// leaf branch
.map(v -> urls().parse(v).map(Optional::<List<? extends CharSequence>>of))
```

javac probes confirm both compile warning-free under `-Xlint:unchecked,rawtypes -Werror` on the shapes above. The witness is available at the emission site, so this is local to the two `CodeBlock`s that build the leg.

Worth deciding during implementation: emit the witness always, or only where the element type carries a wildcard. Always is one code path and always correct; conditionally keeps today's output byte-identical for every existing golden.

## Tests the fix should add

- A bridged wildcard element on both wire shapes (bean automatic, record `@OptionalBridge`), on the identity branch and through a leaf, compiled under `-Xlint:unchecked,rawtypes -Werror`, plus a runtime check that absent, present-empty and present values stay distinct across `parse(build(d))`.
- A non-container wildcard-carrying element, so the case is pinned as a property of the bridged leg rather than of identity containers.
- The mirror case that already works — a wildcard *wire* type read through a leaf into a non-wildcard domain element — so the fix is not over-applied.

## Workaround

Declare the domain element without a wildcard (`Optional<List<CharSequence>>`, `Optional<List<String>>`).

### LINKED ISSUE #839: [BUG] @GenerateMapping: a raw or wildcard getter-only List cannot be filled by the JAXB addAll strategy
A bean wire whose getter-only `List` property is **raw** or carries a **wildcard argument** takes the JAXB collection strategy, and the `getX().addAll(...)` the strategy emits cannot type either. The raw case is an unchecked warning in generated source, which fails any user build running `-Werror`; the wildcard case is a hard error.

Found while fixing #830; this is the build-side sibling of that issue's parse-side defect, reproducing on `main` and untouched by #832.

## Repro

A JAXB-style bean: the getter creates the list on first call, there is no setter.

```java
@SuppressWarnings("rawtypes")
public record Doc(String title, List tags) {}

@SuppressWarnings("rawtypes")
public class DocDto {
  private String title;
  private List tags;
  public String getTitle() { return title; }
  public void setTitle(String title) { this.title = title; }
  public List getTags() {
    if (tags == null) { tags = new ArrayList<>(); }
    return tags;
  }
}

@GenerateMapping
public interface DocMapping extends MappingSpec<Doc, DocDto> {}
```

```text
DocMappingImpl.java:26: warning: [unchecked] unchecked call to addAll(Collection<? extends E>) as a member of the raw type List
    wire.getTags().addAll(domain.tags());
                         ^
```

The same bean with `List<? extends CharSequence>` on both sides is a hard error:

```text
DocMappingImpl.java:28: error: incompatible types: List<CAP#1> cannot be converted to Collection<? extends CAP#2>
    wire.getTags().addAll(domain.tags());
                                       ^
  where CAP#1,CAP#2 are fresh type-variables:
    CAP#1 extends CharSequence from capture of ? extends CharSequence
    CAP#2 extends CharSequence from capture of ? extends CharSequence
```

| Getter-only `List` property | Result |
|---|---|
| `List<String>` | compiles cleanly (today's supported shape) |
| raw `List` | unchecked warning in generated source; fails a `-Werror` build |
| `List<? extends CharSequence>` | hard error, `cannot be converted to Collection<? extends CAP>` |

A property with a *setter* is unaffected on both: `wire.setTags(domain.tags())` types for raw and wildcard alike, which is why #830's matrix (setter-based beans) stays green.

## Cause

`BeanPropertyAnalyser` selects `WriteSite.CollectionAdd` for any getter-only property whose type is exactly `java.util.List`, with no look at its type arguments, and `CollectionAdd.write` emits `receiver.getX().addAll(value)`. `addAll(Collection<? extends E>)` cannot be called on a raw receiver without going unchecked, and on a wildcard receiver the capture on the receiver and the capture on the argument are two different fresh variables, so no argument satisfies it.

```mermaid
flowchart LR
    P["getter-only List property"] --> S{"type arguments?"}
    S -->|"List&lt;String&gt;"| A["getTags().addAll(v)<br/>types cleanly"]
    S -->|"raw List"| W["unchecked call<br/>(fails -Werror)"]
    S -->|"List&lt;? extends X&gt;"| E["two captures never meet<br/>hard error"]
    class A ok
    class W,E bad
    class S decision
```

Contrast the setter path, which is why the strategies diverge: a setter takes the property's declared type verbatim, so it never asks the compiler to relate two captures.

## Suggested fix

Refuse the shape, with a what/why/fix diagnostic from `BeanPropertyAnalyser`: the JAXB collection strategy fills a property through `addAll`, which needs the element type the declaration withholds. That matches the two refusals already next to it — the raw `Map` that cannot lift, and (after #832) the getter-only `List` that cannot carry an `Optional` bridge — and it keeps generated source free of suppressions, which the checker gates guard.

The fix line has two real remedies: declare the type arguments, or give the property a setter (which types for raw and wildcard alike).

The alternative, an unchecked cast plus a suppression in the generated Impl, silences the raw case but cannot rescue the wildcard one, since no cast makes two captures equal. Not recommended.

Note the property must still be *reported*, not silently dropped: dropping it would make a full mapping look like a projection and quietly lose the component.

## Tests the fix should add

- The refusal on a raw and on a wildcard-carrying getter-only `List`, on a full mapping and on a projection.
- The unchanged `List<T>` getter-only shape, compiled under `-Xlint:unchecked,rawtypes -Werror` (the existing JAXB test does not carry those options).
- The same raw and wildcard properties *with* a setter, still mapping, so the refusal is pinned to the write site rather than to the type.

## Workaround

Declare the type arguments on the getter-only `List`, or give the property a setter.

MagnusSmith:
Closes #838. Closes #839.

Two shapes where a container's type arguments defeated generated code, both reachable on `main`, both found while reviewing #830.

## 1. A wildcard-carrying domain element under an `Optional` bridge did not compile

The bridged leg rebuilds the domain's `Optional` from the wire read, and it did so with a generic call whose type argument was *inferred*, so a wildcard captured. `Optional` is invariant, so `Optional<List<CAP>>` is not the declared `Optional<List<? extends CharSequence>>` and the ladder's `.apply(D::new)` no longer typed.

```mermaid
flowchart LR
    R["read for a wildcard-carrying element"] --> B{"bridged?"}
    B -->|no| U["field type List&lt;CAP&gt;"]
    U --> V{"D::new wants<br/>List&lt;? extends CharSequence&gt;"}
    V -->|"the wildcard absorbs the capture"| K["compiled already"]
    B -->|yes| O["field type Optional&lt;List&lt;CAP&gt;&gt;"]
    O --> A{"D::new wants<br/>Optional&lt;List&lt;? extends CharSequence&gt;&gt;"}
    A -->|"Optional is invariant"| X["did not compile"]
    class K ok
    class X bad
    class A,B,V decision
```

The leg now names the element instead of inferring it, on both of its shapes — `Optional.<E>ofNullable(read)` where the read copies, `.map(Optional::<E>of)` where it parses through a leaf.

It names it **only where inference would capture**, so no existing golden moves:

| element | witness | why |
|---|---|---|
| `List<? extends CharSequence>` | yes | its own argument is a wildcard |
| `Outer<? extends Number>.Inner<String>` | yes | capture reaches the enclosing link |
| `Holder<List<? extends CharSequence>>` | no | a wildcard below the element's own arguments never captures |
| `Outer<String>.Inner` | no | nothing on the chain is a wildcard |
| wildcard on the **wire** side only | no | the domain element is what the `Optional` must hold |

**The scan comes back with it.** The bridged leg withheld the element null scan for a wildcard container only because it could not name the `Optional`'s argument — which is precisely what it now does. So a present list is scanned for null elements whatever its arguments, and the bridge stops being a hole in the located-null doctrine. Only a raw container still copies plain, as on every other leg.

## 2. A raw or wildcard getter-only `List` cannot be filled by the JAXB `addAll` strategy

`addAll(Collection<? extends E>)` needs the element type: over a raw receiver the call is unchecked, which fails any build running `-Werror`; over a wildcard one the receiver and the argument capture separately, so no argument can satisfy it. The generated Impl is what breaks, and the author cannot edit it, so the refusal now lands on the declaration that can be changed.

**Asked only where a `build` is emitted.** The bean model is shared with the sparse tier, which reads such a property and never writes it — so an `UpdateSpec` over a JAXB PATCH body, the shape most likely to have one, maps exactly as before.

**The two causes are reported apart**, because their remedies differ:

```text
… is a getter-only List, which a build cannot fill (not supported yet). … a raw List names
none, so the call is unchecked. Declare the type arguments on 'getTags()', for example
List<T>, or give 'tags' a setTags setter, …

… is a getter-only List<? extends CharSequence>, which a build cannot fill (not supported
yet). … a wildcard argument is a fresh type at each mention, so the receiver and the
argument never meet. Replace the wildcard on 'getTags()' with the element type it stands
for, or give 'tags' a setTags setter, …
```

A wildcard author already has an argument, so is never told to add one.

## Shared rule

The typability rule the refusal needs is the one the sparse tier's pinned null scan already had, so both read one predicate. The ladder is now:

- `scanTypable` — not raw. What the null scan needs wherever the ladder infers the leg's result.
- `hasProperArguments` — also no wildcard argument. What a call written *over* the element needs: the sparse tier's `setIfPresent` method reference, and the `addAll` write.

## Tests

The tier matrix gains a bridged wildcard case across all four dense shapes. The bean tests pin both leg shapes, the enclosing-type wildcard, the non-container element, the three shapes that must keep the shorter leg, a located null element inside a bridged wildcard list, a three-way absent / present-empty / present round trip, and the sparse tier keeping the property the dense tier refuses. The JAXB happy path now compiles under `-Xlint:unchecked,rawtypes -Werror`, which is what would have caught the raw case in the first place.

`./gradlew build` green; `:hkj-processor:check` green including the 100% line/branch/instruction gate.

## Summary by CodeRabbit

- **New Features**
  - Improved mapping support for `Optional` values containing wildcard-typed collections, including reliable round-trip handling.
  - Wildcard-typed bridged collections now retain null-element validation.

- **Bug Fixes**
  - Bean mappings now provide clearer diagnostics when getter-only lists lack a concrete element type.
  - Sparse update mappings continue to support reading raw or wildcard getter-only lists.

- **Documentation**
  - Updated mapping guidance to explain getter-only list requirements, build behavior, and bridged-container null handling.


MagnusSmith:
@coderabbit full review

PR #851, opened 2026-09-12
An Optional PATCH property is the remedy offered (#680)

MagnusSmith:
Related to #680, which stays open for the `JsonNullable<T>` carrier. This PR does not close it.

A sparse `UpdateSpec` pairing a plain PATCH property with a domain `Optional` component is refused, and the refusal's first remedy was to change the model:

```
Model the field as a nested record or a sentinel value instead of Optional, or declare the PATCH
property as Optional<java.lang.String> (a present empty Optional then encodes 'set to empty').
```

Both of those change the wire contract, and neither is needed. The refusal now offers the one shape that changes nothing on the wire:

```
Declare 'nickname' as Optional<java.lang.String> with the field defaulting to null, not
Optional.empty(), which would read every omitted property as a clear: null then leaves the
component unchanged, and a present empty Optional sets it empty.
```

## Why that shape

An `Optional<T>` PATCH property whose field defaults to `null` already carries JSON Merge Patch's three states:

| Request body | Bean property | `updateFrom` |
|---|---|---|
| `{}` | `null` | leaves the component unchanged |
| `{"nickname": null}` | `Optional.empty()` | clears it |
| `{"nickname": "ada"}` | `Optional.of("ada")` | sets it, or a located error if a leaf rejects the value |

Three details in the message are load-bearing:

- **It names the wire property's own type, not the domain element type.** With a `String` wire, a domain `Optional<EmailAddress>` and an element leaf `ValidatedPrism<String, EmailAddress>`, the old text suggested `Optional<EmailAddress>`. That compiles to an identity write, so the leaf never runs (`not-an-email` patches in as valid), and Jackson then refuses `{"email":"a@b"}` outright. `Optional<String>` keeps the leaf lifting and the JSON unchanged. Where no leaf converts a differing type, the author gets the no-update-source diagnostic naming the element leaf, which beats a remedy that silently changes the JSON shape.
- **It names the `Optional.empty()` default as the trap.** That default reads an omitted property as a clear, so every request leaving the field out wipes the stored value.
- **Its "why" no longer calls a null-clears rule JSON Merge Patch.** A plain property cannot tell an omitted field from an explicit `null`, so such a rule would clear on omission, which Merge Patch does not; it also read as contradicting the remedy. The same parenthetical is corrected in the method's javadoc and in `beans_patch.md`.

The sibling refusal of an `@OptionalBridge` on an `UpdateSpec` offered the same `Optional<T>` remedy without the null default, and now carries it too.

## A test for the half nobody pinned

What `updateFrom` does with each bean state was already covered in `MappingProcessorUpdateTest`. Which bean state a request body produces is Jackson's decision, and nothing covered it, so the caveats `beans_patch.md` documents rested on untested behaviour. `OptionalPatchPropertyJsonTest` (in `hkj-spring`, against a mapper carrying the module the starter registers) pins all four facts: an omitted property reads `null`, an explicit `null` reads `Optional.empty()`, a value reads `Optional.of`, and an `Optional.empty()` field default reads an omitted property as a clear.

Processor tests:

- `optionalBridgeRejected` asserts the new remedy text.
- `optionalBridgeRemedyKeepsTheWireType` pins that a `String` wire with an `EmailAddress` element leaf is told `Optional<java.lang.String>`.
- The `@OptionalBridge` sparse refusal test asserts the null-default clause.

## Docs

- `beans_patch.md`: the inaccurate Merge Patch parenthetical.
- Glossary `UpdateSpec` entry: "leafless" made precise (a whole-component leaf does patch from a plain property) and the null default added.
- `UpdateSpec` javadoc: the `Optional` rule, which is the refusal users meet most and the javadoc never mentioned.
- A 0.4.11 release-history bullet.

## Review

Five-lens pre-PR panel (correctness, type safety, immutability, ergonomics and docs, architecture) over the real diff, findings verified before acting. The domain-element-type flaw was found by the ergonomics lens from the code and reproduced end to end by the correctness lens (leaf bypassed, validation skipped, JSON contract broken, identically through a shared mix-in). That flaw predates this PR, but the PR had made it the only remedy.

Deliberately not changed:

- **Type rendering** still goes through `TypeMirror.toString`, so a type-use annotation appears in the message. Every sibling sparse-tier diagnostic renders the same way, and switching this one alone would make it the odd one out; a sweep of the family is the right shape for that.
- **The Jackson test uses a hand-built mapper**, not Spring Boot's customised one. The starter configures no `Optional` null handling today, verified by two lenses, and a Spring-context variant would add cost to guard a setting nothing touches.
- **Restructuring `beans_patch.md`** to give the `Optional` property its own bullet with a verified snippet is a book rewrite beyond this change.

## Verification

`gradle :hkj-processor:clean :hkj-processor:check` (including `jacocoTestCoverageVerification`), `:hkj-annotations:check`, `:hkj-spring:autoconfigure:check` and `:hkj-examples:bookVerify`, BUILD SUCCESSFUL on `main` at #854. Rebased onto #850 (five shared files), #853 and #854 (the release history only), and #856, which lets a dense-tier `Optional` bridge nest through a spec and touched four of these files without reaching the sparse tier. The merges kept every change above; none added a refusal or book sentence offering the same `Optional` remedy, and the rule the new advice rests on, that a nested spec does not lift through a sparse container, still stands. The new and changed tests each ran and passed: `optionalBridgeRejected`, `optionalBridgeRemedyKeepsTheWireType`, the `@OptionalBridge` sparse refusal, and all four `OptionalPatchPropertyJsonTest` cases.


MagnusSmith:
@coderabbit full review

PR #854, opened 2026-09-12
A client keeps the @OnStatus overrides it inherits from a jar (#849)

### LINKED ISSUE #849: [BUG] A client extending a base interface from a jar loses its @OnStatus overrides
A `@HkjHttpClient` interface that extends a base interface from another module loses every `@OnStatus` override on the inherited methods. The generated client falls back to the default decoder for the declared error type, and nothing is reported.

This is the defect #844 fixed for the mapping annotations, in the HTTP client: `@OnStatus` and `@OnStatuses` are `RetentionPolicy.SOURCE`, so a method read from a class file carries no trace of them.

## Measured

The base interface declares the overrides:

```java
public interface BaseApi {
  @GetExchange("/{id}")
  @OnStatus(value = 404, error = NotFound.class)
  @OnStatus(value = 409, error = Conflict.class)
  EitherPath<ApiErr, UserDto> getUser(@PathVariable String id);
}
```

and a client extends it:

```java
@HttpExchange("/users")
@HkjHttpClient
public interface ChildApi extends BaseApi {}
```

The two compilations below differ only in where `BaseApi` comes from. Both succeed, with no warnings, and generate different decoders:

| `BaseApi` is | generated `ChildApiClient` decoder |
|---|---|
| a source file in the same compilation | `ResponseErrorDecoders.<ApiErr>forDefault(decoderFactory, ApiErr.class).on(404, NotFound.class).on(409, Conflict.class).build()` |
| a class file on the classpath | `decoderFactory.create(ApiErr.class)` |

So a 404 decodes to `NotFound` when the base interface is a sibling source file, and to the generic `ApiErr` default when the same interface arrives as a jar. The client compiles and runs either way, so the difference only shows up as an error response of the wrong type at runtime.

## Why

```mermaid
flowchart TD
    C["@HkjHttpClient ChildApi<br/>extends BaseApi"] --> M["abstractMethods: getAllMembers(iface)<br/>includes inherited methods"]
    M --> R["readStatusOverrides:<br/>method.getAnnotationsByType(OnStatus.class)"]
    R --> Q{"where was BaseApi compiled?"}
    Q -->|"this compilation"| OK(["overrides read,<br/>status dispatch generated"])
    Q -->|"a dependency"| X(["SOURCE retention: none survive,<br/>default decoder, silently"])

    class C,M,R,Q wire
    class OK tier
    class X error
```

`HkjHttpClientProcessor.abstractMethods` deliberately collects inherited members. Its javadoc says so: "including those inherited from super-interfaces, so a client interface that `extends` a shared base still generates a complete native interface and facade". `readStatusOverrides` then reads `getAnnotationsByType(OnStatus.class)` off each of them. The MaybePath warning, which says `@OnStatus` has no effect on a `MaybePath` method, reads them the same way.

For a method whose declaring interface came from a class file, the annotations were discarded when that interface was compiled, so the processor finds none and generates the default decoder.

The shape is documented as supported. [Declarative HTTP clients](https://higher-kinded-j.github.io/spring/declarative_http_clients.html) lists inheriting methods from a precompiled base in a dependency jar among its caveats, requiring that jar to be compiled with `-parameters`, and says nothing of `@OnStatus`. The existing inheritance tests in `HkjHttpClientProcessorTest` cover a base interface compiled alongside the client only.

## What "good" looks like

`@OnStatus` and `@OnStatuses` become `RetentionPolicy.CLASS`, as `@MapField`, `@OptionalBridge`, `@MapKey`, `@GenerateFocus` and `@TraverseField` now are. The processor reads them through `javax.lang.model` at compile time and nothing reflects on them, so `CLASS` is enough.

Retaining `@OnStatuses` matters as much as `@OnStatus`: two overrides on one method are stored as the container annotation, and `getAnnotationsByType` unwraps it only if it survived.

## Scope

- In: the retention change on both annotations, and a cross-module regression test with the base interface compiled to class files, asserting the generated client matches the one-compilation case, including the repeated-annotation container.
- Out: `@HkjHttpClient` itself, which is read only off the interface being generated in this compilation.

## Note

Found by the pre-PR review panel on #847, which swept the repository's annotations for the retention defect #844 fixed. Reproduced independently with the base interface compiled to class files. Pre-existing on `main`.

MagnusSmith:
Closes #849

## Problem

`@OnStatus` and `@OnStatuses` were `SOURCE`-retained. `HkjHttpClientProcessor` collects inherited abstract methods, so a method a client interface inherits from a base interface in a jar arrived with no trace of its overrides. Both builds below succeed with no warnings and generate different decoders:

| `BaseApi` is | generated `ChildApiClient` decoder |
|---|---|
| a source file in the same compilation | `ResponseErrorDecoders.<ApiErr>forDefault(decoderFactory, ApiErr.class).on(404, NotFound.class).on(409, Conflict.class).build()` |
| a class file on the classpath | `decoderFactory.create(ApiErr.class)` |

A 404 then decoded without the subtype its override named, and code matching on that subtype took the wrong branch at runtime.

## Change

Both annotations are retained in the class file. `@OnStatuses` matters as much as `@OnStatus`: two overrides on one method are stored as the container, and `getAnnotationsByType` unwraps it only if it survived. `CLASS` rather than `RUNTIME`, since the processor reads them through `javax.lang.model` and nothing reflects on them.

Once overrides cross the boundary their checks fire on methods read from class files, and javac gives such a diagnostic no file and no line: the author learned an override was not assignable without learning which method. Every per-method diagnostic now decides where it lands the way `PathProcessor` already does for `@PathVia`:

```mermaid
flowchart TD
    M["a problem with one of the client's methods"] --> Q{"declared by the<br/>@HkjHttpClient interface?"}
    Q -->|yes| L(["reported at the method's own declaration"])
    Q -->|"no: inherited, from a jar<br/>or compiled alongside"| C(["reported on the client interface:<br/>'getUser', inherited from 'BaseApi': ..."])

    class M,Q decision
    class L,C tier
```

A method read from a class file can also name a type its own module could see and the client's cannot, and javac reports nothing about it. An override type or a declared error type missing that way is now reported as not on this compilation's classpath, rather than as "not assignable" or as a failure inside generated code. The check applies to class-file methods only: in source javac reports an unresolvable name itself, and a second message with dependency advice would misdirect a typo.

## Behaviour change

- A base interface jar compiled against an earlier release has already lost its overrides; rebuilding it restores them.
- An inherited override is checked where it was silently ignored, so one naming a type that is not assignable to the method's declared error type, or one missing from the client's classpath, fails a build that compiled before. The duplicate-status and `MaybePath` warnings reach an inherited override too, which fails a `-Werror` build.
- A diagnostic about an inherited method, including one from a base compiled alongside, is reported on the client interface rather than at the base's declaration.
- The annotations are in the class files of the interfaces that declare them. A module compiling against those class files without `hkj-spring-boot-client` warns under `-Xlint:all` (`[classfile] Cannot find annotation method`), so the book now advises a module declaring `@OnStatus` on a shared interface to expose the client as `api`.

Recorded in the release history.

## Tests

`HkjHttpClientProcessorClasspathTest` compiles the base module on its own (no processor, `-parameters`, as the book requires of a base in a jar) and lays it out as class files:

- the generated client and native interface equal a one-compilation build byte for byte, with both override shapes (the `@OnStatuses` container and a bare `@OnStatus`) named first;
- a non-assignable, a duplicate and a `MaybePath` override inherited from the jar are each reported on the client, naming the method;
- an override type, and separately a declared error type, missing from the client's classpath each say so, once;
- an inherited unsupported return type is reported on the client.

With the annotations back at `SOURCE`, the three override tests fail; the other two pin behaviour that never depended on retention. In `HkjHttpClientProcessorTest`, a method the client declares itself is still reported on its own line, and a misspelt override type in the client's own source is left to javac rather than called a missing dependency.

Docs: the per-method section and the precompiled-base caveat in Declarative HTTP clients, the hkj-spring skill, and the release history.

## Review

A five-lens panel (correctness, type safety, immutability, ergonomics, architecture) reviewed the first commit; the second applies its findings. The missing-type check was first written for every method, which four lenses independently caught. Declined, with reasons:

- Naming parameter types or an enclosing type when an inherited method is overloaded or nested: the prefix follows `PathProcessor.names`, which has the same limit, and the two should change together.
- Reporting a base compiled alongside at its own declaration rather than on the client: one rule for every inherited method, as `PathProcessor.rejectMethod` has.
- Deferring a client whose override names a type another processor generates later in the same round: that shape failed before this change too (javapoet rejects the unresolved name), and a deferral is a separate change.
- Pre-existing and unrelated to retention: an inherited method from a generic base (`Child extends Base<ApiErr>`) is analysed with the base's type variables rather than the client's instantiation, and `ReturnInfo.error()` is `@Nullable` where two callers take it as non-null. Worth an issue of their own.

## Summary by CodeRabbit

- **Bug Fixes**
  - Inherited `@OnStatus` overrides now work when the base client interface is compiled in another module or dependency.
  - Validation errors for inherited methods identify the client interface, method, and source interface.
  - Missing error types and unsupported inherited configurations now produce clearer diagnostics.

- **Documentation**
  - Added migration guidance for rebuilding older dependencies and exposing required client dependencies.
  - Documented inherited status overrides, diagnostics, and related build warnings.


MagnusSmith:
@coderabbit full review

PR #856, opened 2026-09-13
An optional nested object maps through its own spec (#825)

### LINKED ISSUE #825: [FEAT] @GenerateMapping: the Optional bridge through a nested spec
A domain `Optional<T>` bridges to a nullable wire component `T` when the element copies by identity, or through a leaf named after the component. It does not bridge when the element pair is mapped by a **sibling spec**, which is how every other correspondence in the mapper reaches a nested record. So the shape a JSON boundary reaches for constantly, an optional nested object, has no mapping.

`resolveCorrespondence` says so in a code comment, and has since the bean bridge landed:

```java
// Optional bridge: a domain Optional<DE> maps to a nullable wire component PE. Empty <->
// null/absent; the element is copied (identity) or mapped through a leaf, exactly as an
// Optional element would be. (An Optional bridge through a nested spec is a follow-up.)
```

Split out of #673, which added the record-wire opt-in; the limitation predates it and applies to bean wires just the same.

## Measured

On `main` (`7d3c852fb`), with `AddressMapping` registered in the same compilation:

```java
record Address(String city) {}
record AddressDto(String city) {}

@GenerateMapping
interface AddressMapping extends MappingSpec<Address, AddressDto> {}

record Customer(String id, Optional<Address> address) {}
record CustomerDto(String id, AddressDto address) {}

@GenerateMapping
interface CustomerMapping extends MappingSpec<Customer, CustomerDto> {
  @OptionalBridge
  Optional<Address> address();
}
```

```
error: @GenerateMapping: domain field 'Customer.address' is Optional<com.example.Address>,
bridged to the nullable record component 'address' of type com.example.AddressDto, but the
element types differ and no leaf converts them. ... Declare '@OptionalBridge default
ValidatedPrism<com.example.AddressDto, com.example.Address> address()' as the component's only
spec method, replacing any marker for it, or align the element types.
```

The refusal is accurate and the fix works, but it asks the author to hand-write a leaf delegating to a mapping the processor has already generated and can already see:

```java
@OptionalBridge
default ValidatedPrism<AddressDto, Address> address() {
  return AddressMappingImpl.INSTANCE.asValidatedPrism();
}
```

That is exactly what `resolveNestedSpec` writes for every other container.

## Why

Every other container consults the registry before giving up; the bridge branch does not.

```mermaid
flowchart TD
    P["Optional&lt;Address&gt; domain,<br/>AddressDto wire"] --> B{"bridge branch"}
    B --> L{"leaf over the<br/>element types?"}
    L -->|yes| OK1(["OPTIONAL_BRIDGE<br/>through the leaf"])
    L -->|no| I{"element types<br/>the same?"}
    I -->|yes| OK2(["OPTIONAL_BRIDGE<br/>identity copy"])
    I -->|no| E(["refused"])
    E -.->|"the step List and Optional<br/>containers already take"| N["resolveNestedSpec:<br/>a registered parse-capable<br/>spec for the element pair"]

    class P,N wire
    class OK1,OK2 tier
    class B,L,I decision
    class E error
```

`List<Address>` → `List<AddressDto>` and `Optional<Address>` → `Optional<AddressDto>` both lift through the sibling spec. Only the *bridged* pair, `Optional<Address>` → nullable `AddressDto`, does not.

## What "good" looks like

The bridge consults `resolveNestedSpec` for the element pair before refusing, so the spec above needs no leaf at all:

```java
@GenerateMapping
interface CustomerMapping extends MappingSpec<Customer, CustomerDto> {
  @OptionalBridge
  Optional<Address> address();   // nests automatically, like every other container
}
```

`build` emits `domain.address().map(AddressMappingImpl.INSTANCE::build).orElse(null)`; `parse` reads `Optional.ofNullable` and delegates the present value to the nested `parse`, so failures locate as `address.city` exactly as they do through a `List` or an `Optional`.

- **Ambiguity** is already handled: `resolveNestedSpec` reports two candidate specs for a pair, and the bridge should inherit that rather than re-implement it.
- **Recursion.** A self-recursive optional (`record Node(String id, Optional<Node> parent)` against a nullable `NodeDto`) must terminate the way the existing self-recursive nesting test does.
- **Tier arithmetic is unchanged.** A bridged component is already a non-identity correspondence, so no `asIso()` and a bridged projection still takes `patch`.
- **Both wire shapes.** The bean bridge should gain it in the same change; nothing here is record-specific.
- **The refusal stays** for a pair that no registered spec covers, and should keep naming the leaf to declare. It could also name a registered-but-not-parse-capable spec, the way `projectionSpecHint` does for the whole-component case.

## Dependencies / scope

- Depends on: nothing. #673 has landed.
- In: nested-spec resolution inside the bridge branch on both wire shapes, the diagnostic when nothing covers the pair, laws, goldens, and the book's [Optional fields](https://higher-kinded-j.github.io/mapping/basics.html#optional-bridge) and [Nesting](https://higher-kinded-j.github.io/mapping/structure.html) pages.
- Out: tri-state semantics; anything about `Optional`-typed wire components, which map today.

MagnusSmith:
Closes #825.

## What changes for users

A domain `Optional<T>` against a nullable wire component `T` (the Optional bridge) used to take its present element through an identity copy or an element leaf only. An element pair mapped by a sibling `@GenerateMapping` spec was refused, and the refusal asked for a hand-written leaf delegating to the Impl the processor had already generated. Every other container (`List`, `Set`, arrays, `Optional`, `Map`) already nested through that spec.

```java
@GenerateMapping
interface ReferralMapping extends MappingSpec<Referral, ReferralDto> {
  @OptionalBridge
  Optional<Customer> referrer();   // nests through CustomerMapping, no leaf
}
```

```mermaid
flowchart TD
    P["Optional&lt;Customer&gt; domain,<br/>nullable CustomerDto wire"] --> L{"bridged element leaf?"}
    L -->|yes| OK1(["through the leaf"])
    L -->|no| I{"element types<br/>the same?"}
    I -->|yes| OK2(["identity copy"])
    I -->|no| N{"specs serving the site<br/>for the element pair"}
    N -->|one| OK3(["nests through the spec"])
    N -->|two| A(["ambiguous: names the<br/>bridged leaf that chooses"])
    N -->|none| E(["refused: names the leaf,<br/>a spec for a record pair,<br/>and any spec that cannot nest"])

    class P wire
    class OK1,OK2,OK3 tier
    class L,I,N decision
    class A,E error
```

- `build` writes the nested build of a present value and `null` for an empty one (a bean property is left unset). `parse` reads `null` as `Optional.empty()` and hands a present value to the nested `parse`, so failures locate under the component (`referrer.email: not an email address`).
- Works on record and bean wires, on full mappings and projections (the validated `patch`), and on one-directional beans, where the site's direction picks `asValidatedParse()` or `asValidatedBuild()`. Threaded and element-mapped generic specs, sealed pairs, flattened members, renames and cross-module specs were probed during review and compile under `-Xlint:all -Werror`.
- Tier arithmetic is unchanged: a bridged component was already non-identity, so no `asIso()`, and a bridged projection takes `patch`.
- Recursion terminates: a self-recursive `Node(String id, Optional<Node> parent)` delegates to its own `INSTANCE`.

## Diagnostics

- The bridge refusal now reads "neither a leaf nor a mapping spec converts them", names the spec route in its why, appends the hint for a registered spec that cannot be nested (a projection, a one-directional bean), and offers declaring a spec for a record pair only. The offer is withheld elsewhere because the commonest refusal, a value record against a `String`, is a pair no spec can map.
- A record component left unmarked offers the bare marker when exactly one spec serving the site maps the element pair. The candidate query (`servingCandidates`) is now shared with `resolveNestedSpec`, so the offer and the resolution cannot disagree.
- **Wording change beyond the bridge.** The ambiguity error, the shadowed-classpath note, the element-mapped composition error and the cycle error each offer a leaf in the spec's place. They now spell out the whole declaration (`'default ValidatedPrism<W, D> x()'`), and on a component carrying an `@OptionalBridge` marker they add the annotation and "as the component's only spec method": the marker and a same-named leaf are one method, so the old `'x()'` advice did not compile when followed. The plain composition text is unchanged; the ambiguity and note texts changed for every nested component (three test quotes updated).
- An `@OptionalBridge` leaf declared over an element-mapped generic spec's own element pair is supplied to that spec, as for an unbridged component. The `OptionalBridge` javadoc says so, and a test follows the diagnostic's fix through to a compiling spec.

## Tests

- `TheOptionalBridge`: nesting with a located failure, self-recursion, ambiguity with its fix text, leaf beats spec, identity beats a same-typed spec, the unusable-spec hint and record-pair offer, the unmarked marker offer, and the element-mapped leaf from diagnostic to compiling spec.
- `MappingTierMatrixTest` row "Optional bridge through a nested spec": record and bean wires, full and projection tiers, round trips and unset-as-empty.
- `MappingProcessorOneDirectionalTest`: parse-only and build-only bridges through one-directional nested specs.
- `GeneratedMappingLawsTest`: an absent and a present nested object, both law-checked.
- Golden `BridgedCustomerMappingImpl` gains a nested bridged component.
- `:hkj-processor:check` green (2679 tests, `MappingProcessor` at 100% line and branch); `:hkj-examples:bookVerify`, `:hkj-annotations:javadoc` and spotless green.

## Docs

A new "Optional nested objects" section in `mapping/structure.md` includes a compiled and run example from `RecordMappingBook`. The placement table in `basics.md` now asks "does the present value need a leaf?". `beans_patch.md`, the `hkj-mapping` skill, the `GenerateMapping` and `OptionalBridge` javadoc and the 0.4.11 release note are updated.

## Review

A five-lens panel (correctness, type safety, immutability, ergonomics, architecture) reviewed the implemented diff; the correctness and type-safety lenses reproduced dozens of scenarios with javac. No blockers. The one major, raised by three lenses (inherited fix lines unfollowable on a marked record component), is fixed as described above. The minors and nits (stale "through the leaf" wording, a duplicated registry query, a missing identity-beats-spec test, overlapping tests) are folded in. The immutability lens found nothing.

## Not in this PR

- A bridged container of mapped elements (`Optional<List<Customer>>` against a nullable `List<CustomerDto>`) is still refused; the book says "not supported yet".
- Pre-existing, found by the type-safety lens: `covers` treats an error type nested inside a type argument as a match, since javac's `isSameType` is true for `ErrorType`. javac's own "cannot find symbol" fails such a build today, so it only matters for a type generated in a later round.

## Summary by CodeRabbit

- **New Features**
  - Optional values can now map through their own nested mapping specifications when bridged to nullable record or bean fields.
  - Present and absent values are handled consistently during build and parse operations.
  - Nested validation failures now report precise paths, such as `referrer.email`.

- **Bug Fixes**
  - Improved guidance for ambiguous or unsupported optional mappings, including complete suggested leaf declarations.

- **Documentation**
  - Expanded guidance, examples, and release notes for optional nested-object mapping, supported scenarios, precedence rules, and limitations.


MagnusSmith:
@coderabbit full review

PR #866, opened 2026-09-17
A mapping over a raw type compiles under -Werror (#865)

### LINKED ISSUE #865: [BUG] @GenerateMapping: a marker over a raw type makes the generated Impl emit a raw override, failing -Werror
A spec that marks a component whose type is a raw container gets a generated Impl that fails a `-Werror` build. The Impl implements every marker method (`@MapField`, `@OptionalBridge`, `@Flatten`) with a stub that throws, and the stub restates the marker's return type exactly as declared. When that type is raw, javac reports `[rawtypes]` against generated code. The user cannot edit that code, and the `@SuppressWarnings("rawtypes")` they put on the spec does not reach it.

The rest of the Impl is already clean. A raw identity `List`, `Set` or `Map` component without a marker compiles warning-free: `MappingTierMatrixTest` holds its raw rows to `-Xlint:unchecked,rawtypes -Werror`. A marker is the one route that writes the raw type back out.

## Measured

On `main` (`5762e9839`), compiled with `-Xlint:all,-processing -Werror`:

```java
@SuppressWarnings("rawtypes")
public record Directory(String id, Optional<List> contacts, List tags) {}

@SuppressWarnings("rawtypes")
public record DirectoryDto(String id, List contacts, List labels) {}

@GenerateMapping
@SuppressWarnings("rawtypes")
public interface DirectoryMapping extends MappingSpec<Directory, DirectoryDto> {
  @OptionalBridge Optional<List> contacts();
  @MapField(to = "labels") List tags();
}
```

```
DirectoryMappingImpl.java:51: warning: [rawtypes] found raw type: List
  public Optional<List> contacts() {
                  ^
DirectoryMappingImpl.java:59: warning: [rawtypes] found raw type: List
  public List tags() {
         ^
```

Without `-Werror` the same sources compile, and the mapping is correct. The warning is the only defect.

| Shape | Result |
| --- | --- |
| raw identity `List`, `Set` or `Map` component, no marker | compiles without a warning |
| `@OptionalBridge Optional<List> contacts();` | `[rawtypes]` in the Impl |
| `@MapField(to = "labels") List tags();` | `[rawtypes]` in the Impl |

## Why

`MappingProcessor.addMarkerStubs` emits one override per marker name, returning `ProcessorUtils.typeNameOf(narrowest)`, the declared return type as written. Nothing asks whether that type mentions a raw type, so a raw one is copied into the Impl verbatim. The spec's own suppression is on the interface, and a generated Impl is a separate compilation unit.

## What "good" looks like

A spec that compiles warning-free under `-Xlint:all -Werror` generates an Impl that does too. A marker over a raw container is either honoured without a warning in generated code, or refused with a fix the author can follow.

## Things to explore

- **Suppress on the stub, or refuse the shape.** One option is `@SuppressWarnings("rawtypes")` on a stub whose return type mentions a raw type anywhere, since a raw type nested inside an argument (`Optional<List>`) cannot be rewritten away. A top-level raw `List` could become `List<?>`, which is return-type-substitutable, but the same fix cannot reach inside an invariant argument. The other option is to refuse the marker with the "declare the type arguments" fix that raw `Map` lifting already gives. Declaring the component with type arguments is the workaround today.
- **All three vocabularies share the stub.** `@MapField` and `@OptionalBridge` are measured above. A `@Flatten` over a raw generic record should be checked too.
- **Narrowest-return groups.** When unrelated mix-ins agree on a marker, the stub returns the narrowest declaration, so the check belongs on that chosen type.
- **A pin.** A tier-matrix row, or a golden, with a marker over a raw container keeps the Impl held to `-Xlint:rawtypes -Werror` like the existing raw rows.

## Priority

Proposed **P3**. It is warning noise in generated code on an uncommon shape, a marker over a raw container. It fails only under `-Werror` and never at runtime, and declaring the type arguments avoids it. This is the same class as #771, which was triaged P3.

## Dependencies / scope

- Depends on: nothing. Found by the type-safety review of #864 (#860), which does not change the stub.
- In: the marker stubs in the `MappingProcessor` Impl emitters, and a test pinning a marker over a raw container under `-Werror`.
- Out: raw containers without a marker, which already compile clean.

MagnusSmith:
Fixes #865.

## The defect

A `@GenerateMapping` spec over a generic type written without its type arguments compiled on javac's defaults and failed a build run with `-Xlint:rawtypes` and `-Werror`. The generated `Impl` repeated the raw type in a file the author cannot edit, and the `@SuppressWarnings("rawtypes")` on their own spec does not reach it.

The issue reported this for a marker's stub. Probing found the same failure well beyond that, several of them with no marker involved:

- marker stubs, in every tier (full, lens, patch, bean, sparse, parse-only, build-only, sealed, generic, element-mapped);
- an abstract leaf's field, accessor, constructor and factory (`ValidatedPrism<List, T> item()`);
- lambdas in `build`, `parse`, `patch` and `updateFrom`, including the `Optional<List>` PATCH property the processor itself recommends. javac reports a raw type nested in an inferred lambda parameter (`Optional<List>`) but stays silent on a top-level one (`List`), and a chunk's tuple nests everything it carries;
- `@GenerateMerge`, when its target is wider than sixteen components;
- a spec whose type parameter has a raw bound (`T extends List`), which fails on the `Impl`'s own type-parameter clause.

`@Flatten` over a raw type was already refused, and stays refused.

## The fix

`ProcessorUtils.rawTypesSuppression(types)` returns `@SuppressWarnings("rawtypes")` when `firstRawIn` finds a raw type among the types a member writes out or infers, and nothing otherwise, so call sites add it inline with no branch of their own.

- The marker stub asks about its own return type; an abstract leaf's field and accessor about that leaf, and the constructor and factory about every leaf they take.
- `build`, `parse`, `patch` and `updateFrom` ask about the pair they map. Which leg puts a raw type into a lambda is an emitter detail that changes, so the question is asked once per method rather than per leg. A method holding no such lambda, such as a record's `build`, carries it too.
- A raw type-parameter bound is answered on the `Impl` class, the one place no member annotation reaches.
- `@GenerateMerge` answers for its target's components the same way.

Only `rawtypes` is suppressed. An unchecked operation in generated code is a hole in that code and is still reported.

## Verification

- `:hkj-processor:check` green on the rebased branch: 2741 tests, 0 failures, and the 100% line, branch and instruction gate on `MappingProcessor*`, `MergeProcessor*` and `ProcessorUtils*`.
- New pins, each compiled under `-Xlint:unchecked,rawtypes -Werror` with `succeededWithoutWarnings()`: two `MappingTierMatrixTest` rows (a nested raw identity and a bridged raw container) plus a test that the suppression lands on members and never on the class; a rename stub, a raw abstract leaf and a raw type-parameter bound in `GenericMappingWireAxisTest`; a chunked parse and patch in `MappingProcessorTest`; a sparse `Optional<List>` property with an element leaf in `MappingProcessorUpdateTest`; and a wide fallible fill in `MergeProcessorTest`.
- Specs naming no raw type generate byte-identical output, checked across a 19-Impl probe set and all 60 mapping Impls in hkj-examples.

## Not in scope

The same defect reaches the optics and assembly generators: `@GenerateLenses`, `@GenerateGetters`, `@GenerateSetters` and `@GenerateAssembly` on any raw component, and `@GenerateFolds` and `@GenerateFocus` on a nested one. `@GenerateTraversals` and `@GeneratePrisms` are clean. That is tracked in #867, so this change stays with the mapper.

## Summary by CodeRabbit

- **Bug Fixes**
  - Generated mapping and merge code now avoids raw-type compiler warnings across methods, classes, fields, accessors, constructors, factories, and inferred parameters.
  - Raw types are suppressed only where needed, while unchecked warnings remain visible.
  - Wide mappings and merges with more than sixteen components now compile cleanly with warnings treated as errors.

- **Documentation**
  - Added release notes for version 0.4.11-SNAPSHOT.


MagnusSmith:
@coderabbit full review

MagnusSmith:
> Include flattened inner domain component types in `pairSuppression` ... can emit a raw domain element in `parseCall` as `$T[]::new` ... cause `-Xlint:rawtypes -Werror` compilation to fail.

Measured before changing anything: the coverage gap is real, the build failure is not.

A flattened inner `ARRAY` component does write the raw domain element out, and `pairSuppression` adds nothing there. Compiled with `-Xlint:all,-processing -Werror`:

```java
@SuppressWarnings("rawtypes")
public final class Raws {
  public record Group(List[] tags) {}
  public record Dom(String id, Group group) {}
  public record Dto(String id, String[] tags) {}

  @GenerateMapping
  public interface RawGroupMapping extends MappingSpec<Dom, Dto> {
    @Flatten Group group();
    default ValidatedPrism<String, List> tags() { ... }
  }
}
```

The generated `parse` carries no `@SuppressWarnings`, and contains exactly the construct the comment names:

```java
.field("tags", hkj$ifPresent(wire.tags(), v -> tags().parseAll(v, List[]::new)))
```

`JAVAC_EXIT=0`: javac does not report `[rawtypes]` for a raw array-constructor reference. Two controls: the same pair left unflattened is suppressed, because its top-level `List[]` component is scanned; and a flattened inner `List` emits `tags()::parseAll`, writing no element type at all.

Widening the scan would therefore add a suppression where no warning exists. Instead, in `c42f2156d`:

- `MappingProcessorFlattenTest.rawArrayElementInsideTheGroup` compiles this shape under `-Xlint:unchecked,rawtypes -Werror` with `succeededWithoutWarnings()` and asserts the emitted `List[]::new`, so the day javac reports that construct, or the emitter writes the element somewhere javac does report, this fails rather than a user's build.
- `pairSuppression`'s javadoc no longer gives the wrong reason. It said no lambda takes the inner types; in fact the inner legs take wire-typed parameters, which `wire.components()` already covers, and the domain side is written out only in that array-constructor reference.

If you would still prefer the scan widened as belt and braces, say so and I will, but it would be suppression without a warning behind it.

PR #870, opened 2026-09-17
Generated optics and companions answer for a raw type (#867)

### LINKED ISSUE #867: [BUG] Optics and assembly generators restate a raw component type, failing -Werror
A record whose component is declared with a raw type gets generated code that fails a `-Werror` build. `@GenerateLenses`, `@GenerateGetters`, `@GenerateSetters`, `@GenerateFolds`, `@GenerateFocus` and `@GenerateAssembly` put the component type into the class they generate, either written out in a signature or as the inferred type of a generated lambda's parameter. When that type is raw, javac reports `[rawtypes]` against a file the author cannot edit, and the `@SuppressWarnings("rawtypes")` on their record does not reach it.

#865 fixed the same defect for `@GenerateMapping` and `@GenerateMerge`. This issue covers the generators that fix did not touch.

## Measured

On `main` (`3e2b3084a`) with the #865 change applied, compiled with `-Xlint:all,-processing -Werror`:

```java
@GenerateLenses
@SuppressWarnings("rawtypes")
public record Doc(String id, Optional<List> contacts) {}
```

```
DocLenses.java:32: warning: [rawtypes] found raw type: List
  public static Lens<Doc, Optional<List>> contacts() {
DocLenses.java:33: warning: [rawtypes] found raw type: List
    return Lens.of(Doc::contacts, (source, newValue) -> new Doc(source.id(), newValue));
DocLenses.java:56: warning: [rawtypes] found raw type: List
  public static Doc withContacts(Doc source, Optional<List> newContacts) {
```

Every record below carries `@SuppressWarnings("rawtypes")`, so every warning is in generated code.

| Generator | `List tags` (bare raw) | `Optional<List> contacts` (nested raw) |
| --- | --- | --- |
| `@GenerateLenses` | `Lens<Doc, List> tags()`, `withTags(Doc, List)` | the signatures, plus the `Lens.of` lambda's `newValue` (above) |
| `@GenerateGetters` | `Getter<Doc, List> tags()`, `List getTags(Doc)` | not measured |
| `@GenerateSetters` | `Setter<Doc, List> tags()`, `withTags(Doc, List)` | not measured |
| `@GenerateFolds` | clean (the raw element widens to `Object`) | `Fold<Doc, Optional<List>> contacts()`, the fold class and its `foldMap` parameter |
| `@GenerateFocus` | clean (the raw element widens to `Object`) | the `Lens.of(...).some().each()` lambda's `newValue` |
| `@GenerateAssembly` | the stage method, field and constructor restate `Validated<NonEmptyList<FieldError>, List>` | the same, plus the curried `a1 -> a2 -> new Doc(...)` lambda |
| `@GenerateTraversals` | clean | clean |
| `@GeneratePrisms` | clean (a variant with a raw component) | not measured |

Without `-Werror` every case compiles; the warning is the only defect.

## Why

Two routes put a raw type into generated code:

1. **Written out.** A generated signature restates the component type: `Lens<Doc, List>`, a `withTags(Doc, List)` parameter, `Fold<Doc, Optional<List>>`, a staged builder's `Validated<..., List>` field.
2. **Inferred.** javac reports a raw type nested inside a type argument of an implicitly typed lambda parameter (`newValue` inferred as `Optional<List>`), while a top-level raw parameter (`List`) is silent. That is why `@GenerateFocus` is clean for `List tags` and not for `Optional<List> contacts`.

`@GenerateFolds` and `@GenerateFocus` are clean for a bare raw `List` because they widen its element to `Object`; neither route asks whether the type they do write names a raw type.

## What "good" looks like

A record that compiles warning-free under `-Xlint:all -Werror` generates optics and a companion that do too. An unchecked operation is still reported.

## Things to explore

- **Member suppression, as #865 did.** `ProcessorUtils.rawTypesSuppression(types)` returns `@SuppressWarnings("rawtypes")` for a member whose types name a raw type (`firstRawIn`), and nothing otherwise. Each generated member that writes out or infers a component type can add it, and the rest of the generated class stays held to `rawtypes`.
- **Public signatures.** Unlike a mapping `Impl`, these generators publish the raw type: `DocLenses.tags()` returns `Lens<Doc, List>`. A caller holding it in `var` or chaining the call sees no warning; one who writes the type out gets the warning in their own file, where they can suppress it. So suppressing inside the generated class hides nothing from callers. Widening to `Object`, as `@GenerateFocus` and `@GenerateFolds` already do for a bare raw element, is an alternative where the published type would otherwise be raw.
- **Refuse or honour.** `OpticsSpec<Box>` and `@PathVia` both refuse a raw type, because each is a signature written for the generator, where the type arguments are the author's to supply. A record component is different: it is the declared shape of the data, legal and working today, so refusing it would break builds that compile without `-Werror`.
- **A pin per generator.** Compile each measured shape under `-Xlint:unchecked,rawtypes -Werror` with `succeededWithoutWarnings()`.

## Priority

Proposed **P3**. It is warning noise in generated code, on raw record components. It fails only under `-Werror`, never at runtime, and declaring the type arguments avoids it. This is the same class as #865 and #771.

## Dependencies / scope

- Depends on: #865 (the shared `ProcessorUtils.rawTypesSuppression`).
- In: `LensProcessor`, `GetterProcessor`, `SetterProcessor`, `FoldProcessor`, `FocusProcessor`, `AssemblyProcessor`.
- To measure first: `IsoProcessor`, `ErrorEnvelopeProcessor`, `AccumulatorProcessor`, `ForComprehensionProcessor`, and the nested-raw cells marked "not measured" above.
- Out: `TraversalProcessor` and `PrismProcessor`, already clean; `ImportOpticsProcessor` and `PathProcessor`, which refuse a raw type; `EffectAlgebraProcessor`, `ComposeEffectsProcessor`, `PathSourceProcessor` and hkj-spring's `HkjHttpClientProcessor`, which generate from interfaces rather than record components and are not measured here.

MagnusSmith:
Fixes #867.

## The defect

A generic type written without its type arguments gave generated code that failed a build run with `-Xlint:rawtypes` and `-Werror`, in a file the `@SuppressWarnings("rawtypes")` on the author's own declaration cannot reach. #865 fixed this class for `@GenerateMapping` and `@GenerateMerge`; this is the same fix for the generators that one did not touch.

## Measured, on three axes

The issue listed two axes and several cells as "not measured". Measured before changing anything, with `-Xlint:all,-processing -Werror`:

| Generator | bare `List tags` | nested `Optional<List>` | bound `<T extends List>` |
| --- | --- | --- | --- |
| `@GenerateLenses`, `@GenerateGetters`, `@GenerateSetters` | fails | fails | fails |
| `@GenerateFolds`, `@GenerateFocus` | clean (element widens to `Object`) | fails | fails |
| `@GenerateTraversals` | note, no traversal | fails | fails |
| `@GeneratePrisms` | clean | clean | fails |
| `@GenerateAssembly` | fails | fails | refused (generic records) |
| `@GenerateIsos` | n/a | fails (the declared `Iso`) | refused (type variables) |
| `@GenerateErrorEnvelope` | fails (variant and context) | fails | refused (generic declarations) |

The **bound axis** is not in the issue at all: every member these generators emit redeclares the record's type parameters with their bounds, so a raw bound reaches members whose own types name nothing raw. A `@GenerateFocus` navigator leaks twice over: the navigator class's own methods write the target record's field types (a raw `Map`, `Function` or `Comparable`, including across modules), and the static method returning the navigator carries the bound.

`@GenerateAccumulators` and `@GenerateForComprehensions` are out of scope: both are `@Target(PACKAGE)` and generate HKJ's own helper families, never a user component type.

## The fix

`ProcessorUtils.rawTypesSuppression` gains a second overload taking the type a member is generated around plus the owner whose type parameters it redeclares, so one call covers both routes. Each generated member that writes out or infers such a type adds the annotation itself; a fold's and a traversal's implementation class takes it at class level, because no member annotation reaches a superinterface clause or a class's own type-parameter clause.

Only `rawtypes` is suppressed. An unchecked operation in generated code is a hole in that code and is still reported, which the pins prove by compiling with `-Xlint:unchecked,rawtypes`.

Two deliberate choices, both consistent with #865: Fold and Focus ask about the whole component rather than the element they widen to, which over-suppresses on a bare raw `List` where javac happens to stay silent; and the suppress-versus-refuse boundary is now recorded in the helper's javadoc and in `compiler_errors.md`, since `@ImportOptics` and `@PathVia` refuse a raw type for the very rationale these generators now answer for themselves.

## Verification

- `:hkj-processor:check` green: 2752 tests, 0 failures, coverage gates verified.
- Ten pins, one per generator, each compiling its raw shapes under `-Xlint:unchecked,rawtypes -Werror` with `succeededWithoutWarnings()`; six also assert the annotation's placement on the exact member.
- 27 javac-direct shapes clean, including the five that exposed the gap below.
- Specs and records naming no raw type generate byte-identical output.

## A trap worth recording

The first pass reported `@GenerateTraversals` as a clean control. That was an artefact: the probe classpath had no `hkj-processor-plugins` jar, so the generator found no `TraversableGenerator` through the SPI and emitted an empty companion, and the probes measured nothing. With the plugin present — the configuration every consumer uses, and already a `testImplementation` here — traversals fail on five axes. The harness now uses the plugin-inclusive classpath.

## Added after the first review

Both adjacent leaks of the same class are fixed here rather than deferred.

`@GeneratePathBridge` refused an interface whose type parameter was bounded by a raw type, citing the `[rawtypes]` warning an author's own suppression cannot reach, which is the reason this change answers for everywhere else. The bridge now repeats the bound and carries the suppression on its class, the one place a member annotation cannot reach. Its *signature* refusal stays, for a reason of its own that the diagnostic, the annotation's javadoc and the book now give: the bridge builds its Path from the effect's type arguments, and a raw effect has none. A second review confirmed a raw bound cannot by itself make bridge code unchecked, across twenty shapes and every bridged effect.

`@ImportOptics` restated a raw focus type verbatim in the optic method it generates. That method now answers for the focus the spec declared, for the lens focus a checked composition writes out as a local, and for the bounds of the type parameters it redeclares. Two reviewers independently caught that lens-focus position: a `Map` with a raw key and a clean value gave a clean focus and a raw local, so the generated file still failed `-Werror` while the sibling lens method was suppressed. Since a declaration takes one `@SuppressWarnings`, a method that also needs `unchecked` for a cast-path composition carries both tokens in one annotation, with its own pin.

Verification after this round: 2756 tests, 0 failures, coverage gates verified, and 31 javac-direct shapes clean.

## Not in scope

`@ImportOptics({Foo.class})`, the class-constant route, is served by a different generator (`ExternalLensGenerator`) with its own fixtures, and still restates a raw component. It is pre-existing and goes to a follow-up; the release note says "spec method", so this change claims nothing it does not do.

## Summary by CodeRabbit

* **Bug Fixes**
  * Generated optics and companion classes now compile cleanly with `-Werror` when raw types or raw type bounds are present.
  * Generated code applies targeted raw-type warning suppression while continuing to report unchecked-operation warnings.
  * Path bridges now accept raw bounds declared on their source interfaces, while still rejecting unsupported raw types in bridged signatures.

* **Documentation**
  * Clarified raw-type guidance and compiler diagnostics for optics specifications and path bridges.


MagnusSmith:
@coderabbit full review

PR #883, opened 2026-09-18
Traversals drop a redundant cast, and hkj's annotations draw no -Xlint:processing warning (#877)

### LINKED ISSUE #877: [BUG] Traversals over Optional, Maybe, Try, Validated and Either fields carry a redundant cast that fails -Xlint:all -Werror

A record with an `Optional`, `Maybe`, `Try`, `Validated` or `Either` component gets a generated traversal whose `modifyF` body casts a value to the type it already has. javac reports that under `-Xlint:cast`, which `-Xlint:all` includes, so a build run with `-Xlint:all -Werror` fails in a generated file. Nothing about the record needs to be unusual: `record Doc(String id, Optional<String> more)` is enough.

## Measured

At `a7427b834` (main), javac-direct on JDK 25.0.1 with `-Xlint:all,-processing -Werror`, the processor and `hkj-processor-plugins` both on the processor path. One warning per traversal, and nothing else in the file:

| Component of `record Doc(String id, ...)` | `@GenerateTraversals` | `@ImportOptics({Doc.class})` |
| --- | --- | --- |
| `Optional<String> more` | `redundant cast to Kind<F,String>` | same |
| `Optional<List<String>> more` | `redundant cast to Kind<F,List<String>>` | same |
| `Optional<? extends Number> more` | `redundant cast to Kind<F,Number>` | same |
| `Maybe<String> more` | `redundant cast to Kind<F,String>` | clean: no traversal generated |
| `Try<String> more` | `redundant cast to Kind<F,String>` | clean: no traversal generated |
| `Validated<String, Integer> more` | `redundant cast to Kind<F,Integer>` | clean: no traversal generated |
| `Either<String, Integer> more` | `redundant cast to Kind<F,Integer>` | clean: no traversal generated |

The class-constant route traverses only the JDK containers `TypeKindAnalyser.detectContainerType` recognises, so there it reaches only `OptionalGenerator`. A generic record, `Doc<T extends List<String>>(T value, Optional<String> more)`, fails the same way on both routes.

```
DocTraversals.java:42: warning: [cast] redundant cast to Kind<F,String>
        @SuppressWarnings("unchecked") final var g_of_b_casted = (Kind<F, String>) g_of_b;
```

Not measured: the `@GenerateFocus` navigator route, if it reaches `generateModifyF`.

## Why

Five plugin generators emit the same two statements:

```java
final var g_of_b = f.apply(optional.get());
@SuppressWarnings("unchecked") final var g_of_b_casted = (Kind<F, X>) g_of_b;
```

`f` is declared `Function<X, Kind<F, X>>` by the `modifyF` signature the processor writes, so `g_of_b` is already a `Kind<F, X>` and the cast changes nothing. The `unchecked` suppression on the local answers for nothing either: an identity cast is not unchecked.

- `basejdk/OptionalGenerator.generateModifyF`
- `hkj/MaybeGenerator.generateModifyF`
- `hkj/TryGenerator.generateModifyF`
- `hkj/ValidatedGenerator.generateModifyF`
- `hkj/EitherGenerator.generateModifyF`

The repository's own builds run `-Xlint:unchecked,rawtypes -Werror`, which does not include `cast`, so nothing in the suite sees it.

## What "good" looks like

A record that compiles warning-free under `-Xlint:all -Werror` gets traversals that do too, on either route.

## Things to explore

- **Drop the cast and its suppression**, passing `g_of_b` to `applicative.map` directly. Check the wildcard shapes first (`Optional<? extends Number>`, `Optional<?>`), where the focus is the resolved bound or `Object`, to confirm the cast is redundant there too rather than load-bearing.
- **A pin per generator**, compiled under `-Xlint:all -Werror` (or `-Xlint:cast,unchecked,rawtypes -Werror`) with `succeededWithoutWarnings()`.
- Whether the plugin test suite should compile its generated traversals under `-Xlint:all`, so the next lint category is caught where it is introduced.

## Priority

Proposed **P3**. Warning noise in generated code, under a lint category the repository does not enable. It fails only under `-Werror`, never at runtime, and there is no workaround short of dropping `cast` from the lint set.

## Dependencies / scope

- In: the five `generateModifyF` bodies in `hkj-processor-plugins`.
- Out: the `rawtypes` answer on the same traversals (#867 for `@GenerateTraversals`, #873 for `@ImportOptics`).

MagnusSmith:
## Summary

A consumer building with `-Xlint:all -Werror` failed in hkj's generated code for three separate reasons. This PR removes all three, and adds guards so the next one is caught where it is introduced.

1. **Redundant cast in five traversal generators (#877).** `OptionalGenerator`, `MaybeGenerator`, `TryGenerator`, `ValidatedGenerator` and `EitherGenerator` cast `f`'s result to `Kind<F, X>`, the type it already has, under an `@SuppressWarnings("unchecked")` that answered for nothing. javac reports it under `-Xlint:cast`. The generators now map the result as it stands, on the `@GenerateTraversals` route and the `@ImportOptics({Doc.class})` route alike.
2. **`-Xlint:processing` on every generated file and on user sources.** javac reports each annotation no processor *claims*, and every hkj processor named only the annotation it generates for. So `@Generated` on every generated file, and `@MapField`, `@Wither`, `@TraverseField`, `@PathVia`, `@OnStatus` and the rest in users' own sources, were reported, whatever else the build contained. A new `CompanionAnnotationProcessor` claims them. This part goes beyond #877's original scope, folded in at the maintainer's request.
3. **Two more generated shapes.** The functor `@EffectAlgebra` generates for an algebra without `mapK` cast `widen`'s result to its own return type (`-Xlint:cast`). The `@EffectAlgebra` interpreter and the `@HkjHttpClient` configuration declared no constructor (`-Xlint:missing-explicit-ctor`, in a named module that exports the package).

## How the claim works

```mermaid
flowchart LR
    R["a round's annotations"] --> G["each generating processor<br/>claims the one it generates for"]
    G --> C["CompanionAnnotationProcessor<br/>claims @Generated, @MapField, …"]
    C --> Q{"anything left unclaimed?"}
    Q -->|"hkj's: no"| OK(["-Xlint:processing quiet for hkj"])
    Q -->|"third-party: yes"| W(["reported, e.g. @NullMarked, Spring's"])
```

- `hkj-processor`'s companion claims 18 annotations: `@Generated`, `@MappingIndexEntry`, the coverage marker `org.higherkindedj.annotation.Generated`, and the 15 the generating processors read from the elements they generate for. `@PathConfig` stays unclaimed on purpose: no processor reads it.
- `hkj-spring-boot-client-processor` gets a same-named companion for `@Generated`, `@OnStatus` and `@OnStatuses`, because a Spring client build can run the client processor alone. That module does not depend on `hkj-processor`.
- Both generate nothing and are registered with Gradle as **isolating**. Per Gradle 9.4.0's bytecode (`javap`), `IsolatingProcessingStrategy` has no retention check and records no aggregated types. Aggregating would make Gradle reprocess every `@Generated` class on each incremental compile.
- Claiming changes nothing a generating processor sees: none of them lists a companion annotation, and they read these annotations from elements, not from the round's claim set. A guard test pins the two lists as disjoint.
- `HkjHttpClientProcessor` reports `latestSupported()` instead of pinning `RELEASE_25`. That pin would draw a source-version warning, which no lint flag can silence, on a newer `-source`.

## Measured

| | before | after |
|---|---|---|
| javac probe: 22 record shapes on both routes plus the navigator route (45 rows), `-Xlint:all,-processing -Werror` | 29 of 45 rows fail on `[cast]` | 45 clean |
| hkj annotation names in `[processing]` warnings, `-Xlint:all` over `hkj-examples` + both Spring examples | 14 | 0 |
| `[missing-explicit-ctor]` located in generated files, same corpus | 12 | 0 |
| Gradle incremental compile of `effect-example` after adding one file | 31 classes | 31 classes (`Incremental compilation`) |

The review's correctness lens compiled all 440 `hkj-examples` sources with and without the companion: the 415 generated files are byte-identical. With the cast restored, the traversals' bytecode differs only by one local-variable slot, and runtime results are the same.

## Decisions worth knowing

- **Claim every hkj annotation a processor reads or writes, not only `@Generated`.** Otherwise a mapping user's `-Xlint:all -Werror` build still fails on `@MapField`.
- **Leave third-party annotations alone.** jspecify's `@NullMarked` (written by `@EffectAlgebra`/`@ComposeEffects`) and Spring's (on every `@HkjHttpClient` client) are still reported. Claiming them could hide them from their own processors. The book says how to turn off that one lint.
- **`ValidatedGenerator.getGenericTypeName` stays.** It became unused, but it has shipped since 0.4.0, and a subclass calling it would silently get `E` instead of `A`. It is now pinned directly.
- **Processor order:** javac stops offering a round to further processors once every annotation in it is claimed. A catch-all processor such as Lombok's belongs ahead of `hkj-processor` on the processor path. This already applied to source sets using only generating annotations; the companion extends it to companion annotations. The book and the release note say so, and name the processor for builds that list theirs explicitly.

## Not in scope (follow-ups to file)

- `hkj-processor`'s `module-info` `provides` names 11 of the 20 processors. A consumer on `--processor-module-path` gets no Focus, Fold, Getter, Setter, Iso, Mapping, Merge, ErrorEnvelope or PathSource processing. This is pre-existing; the companion is added to the list.
- `@PathConfig` is documented as package-level defaults for `@PathSource`, but no processor reads it.
- `[exports]` and `[auxiliaryclass]` warnings in generated files come from the consumer's `requires` and file layout, as they do in hand-written code.

## Test plan

- [x] `gradle spotlessCheck`
- [x] clean `:hkj-processor:check`, `:hkj-processor-plugins:check`, `:hkj-spring:client-processor:check` (JaCoCo gates included; plugins at 100% on every counter)
- [x] `:hkj-processor:javadoc`, `:hkj-spring:client-processor:javadoc`
- [x] `:hkj-examples:check` (book snippet gate included), `:hkj-spring:client-example:compileJava`, `:hkj-spring:example:compileJava`
- [x] Every new pin mutation-tested: dropping `@Generated` from the claim, the interpreter constructor, or the companion from the plugin tests fails the pins that guard them; each generator pin fails on `[cast]` against main's generators
- [x] Five-lens review (correctness, type safety, immutability, ergonomics, architecture), run twice. Findings were verified and folded in.

Closes #877

## Summary by CodeRabbit

- **Bug Fixes**
  - Generated optics and effect code no longer produces redundant cast, unchecked, processing, or missing-constructor warnings.
  - Generated code now compiles cleanly with strict Java lint settings, including optional-like and wildcard-based components.
  - Processor support now follows the Java compiler version in use.

- **New Features**
  - Annotation processing automatically recognizes companion annotations, improving clean builds and incremental compilation.
  - Generated client configurations now include explicit public constructors.

- **Documentation**
  - Expanded guidance covers processor ordering, incremental compilation, lint-clean generation, and plugin testing.


MagnusSmith:
@coderabbit full review

Issue #871, opened 2026-09-17
[BUG] @GenerateMapping: a bean field initialiser defeats null-as-absent, so an omitted PATCH list clears the domain


MagnusSmith:
A bean property whose field is initialised, such as `private List<String> tags = new ArrayList<>()` or `private String status = "ACTIVE"`, never reads `null` when nothing set it. Both places that read `null` as absence then act on the default instead:

- **Sparse `UpdateSpec`:** an omitted list arrives as a present, empty list, and `updateFrom` replaces the domain's list with it.
- **Full mapping, automatic `Optional` bridge:** `build` skips the write for an empty `Optional`, so the bean keeps its default. `parse` then reads that default back as present.

Nothing warns, because the processor cannot see initialisers.

The sparse half is the data loss #841 fixed for getter-only lists, reached through the remedy #841's refusal now offers: "Give 'tags' a setTags setter, so an omitted field leaves it null."

Reproduced on `main` at `62b34a1ea`.

## Repro: sparse

```java
public record Profile(String name, List<String> tags) {}

public class ProfilePatch {
  private String name;
  private List<String> tags = new ArrayList<>();           // initialised
  public String getName() { return name; }
  public void setName(String v) { name = v; }
  public List<String> getTags() { return tags; }
  public void setTags(List<String> v) { tags = v; }
}

@GenerateMapping
public interface ProfilePatchMapping extends UpdateSpec<Profile, ProfilePatch> {}
```

```java
var body = new ProfilePatch();
body.setName("grace");                                     // the client omitted 'tags'

ProfilePatchMappingImpl.INSTANCE.updateFrom(body).apply(new Profile("ada", List.of("admin")));
// Valid(Profile[name=grace, tags=[]])                     'admin' is gone
```

The setter is present, so the #841 refusal does not fire. The initialiser alone is enough.

## Repro: dense bridge

```java
public record Profile(String name, Optional<String> status) {}

public class ProfileBean {
  private String name;
  private String status = "ACTIVE";                        // initialised
  // getters and setters for both
}

@GenerateMapping
public interface ProfileMapping extends MappingSpec<Profile, ProfileBean> {}
```

- `build(new Profile("ada", Optional.empty()))` leaves `status = "ACTIVE"`.
- `parse` of that bean is `Valid(Profile[name=ada, status=Optional[ACTIVE]])`.
- `MappingLaws.assertMappingLaws` fails parse-build on it:

```
[...]

PR #884, opened 2026-09-19
A bridged Optional is refused where its write site is declared non-null (#881)

### LINKED ISSUE #881: [BUG] @GenerateMapping: the Optional bridge writes null into a wire slot declared non-null, with no signal
For an empty domain `Optional`, the bridge's `build` writes `null`: into the record component on a record wire opted in with `@OptionalBridge`, and, since #871, through the setter or builder setter of every automatically bridged bean property. The processor never asks whether that slot may hold a `null`. When the slot is declared non-null, the mapping compiles without a word, even under `-Werror`, and then either:

- throws from `build`, when the setter enforces its declared type; or
- leaves a `null` in a field or component whose declared type says it cannot hold one, for a nullness checker or the next caller to trip over.

The book states the contract in both places, and the processor can see every signal it would need to check it.

Reproduced on the #871 fix branch at `747e539e4` (PR #880). The record half reproduces on `main` as well.

## Repro

A JSpecify `@NullMarked` package, where an unannotated reference type is non-null:

```java
@NullMarked
package p;
```

```java
record Profile(String name, Optional<String> nickname) {}

// 1. a bean setter declared non-null, not enforced
class ProfileBean {
  private @Nullable String nickname;
  public @Nullable String getNickname() { return nickname; }
  public void setNickname(String v) { nickname = v; }   // no @Nullable on the parameter
  // name getter/setter alike
}

// 2. the same setter, enforced
class CheckedBean {
  public void setNickname(String v) { nickname = Objects.requireNonNull(v, "nickname"); }
  // otherwise as ProfileBean
}

// 3. a record wire opting in, the component declared non-null
record ProfileDto(String name, String nickname) {}

@GenerateMapping interface ProfileBeanMapping extends MappingSpec<Profile, ProfileBean> {}
@GenerateMapping interface CheckedBeanMapping extends MappingSpec<Profile, CheckedBean> {}
@GenerateMapping interface ProfileDtoMapping extends MappingSpec<Profile, ProfileDto> {
  @OptionalBridge Optional<String> nickname();
}
```

All three compile cleanly under `-Xlint:unchecked,rawtypes -Werror`. Building `new Profile("ada", Optional.empty())`:

```
1 bean, unenforced: nickname=null
2 bean, enforced: build threw java.lang.NullPointerException: nickname
3 record: ProfileDto[name=ada, nickname=null]
```

The generated write in the bean cases is `wire.setNickname(domain.nickname().orElse(null));`.

## What the book says

- `basics.md`, Optional fields: "The bridged wire component is nullable by construction: `build` writes `null` into it for an absent value. In a JSpecify `@NullMarked` package, declare it `@Nullable String nickname` so the wire record says what the mapping does with it."
- `beans_patch.md`, bean-shaped wire targets (added by #871): "A bridged property's writer must take `null` ... in a JSpecify `@NullMarked` package its parameter is `@Nullable`, as a bridged record component's is."

Both are instructions to the reader. Nothing checks them.

## Cause

- `bridgeBuildValue` (`MappingProcessor.java:6554` on `747e539e4`) always renders `orElse(null)`, and `WireShape.buildStatements` (`WireShape.java:110`, `:165`) writes it through the constructor, setter or builder setter.
- Nothing in `MappingProcessor`, `BeanPropertyAnalyser` or `WireShape` reads nullness annotations. The processor's one nullness helper, `NullableAnnotations` (`NullableAnnotations.java:33`, `:74`), recognises `@Nullable` names on record components only, for `WideningAnalysis`'s null-safe lenses. It has no notion of a non-null signal or of a `@NullMarked` scope.

## Suggested fix

Classify a bridge whose write site is declared non-null as a refusal, with a fix line that can be followed.

```mermaid
flowchart TD
    B["a bridged component's write site:<br/>record component, setter parameter,<br/>or builder setter parameter"] --> N{"declared nullable?<br/>(a recognised @Nullable)"}
    N -->|yes| OK["accepted, as today"]
    N -->|no| S{"declared non-null?"}
    S -->|"an explicit non-null annotation"| R["refused: build would write null here"]
    S -->|"inside @NullMarked,<br/>no @NullUnmarked between"| R
    S -->|"no signal either way"| OK

    class B wire
    class OK ok
    class R bad
    class N,S decision
```

- **Signals.**
  - An explicit non-null annotation on the write site: JSpecify `@NonNull`, JSR-305 and Jakarta `@Nonnull`, JetBrains `@NotNull`, AndroidX `@NonNull`, Lombok `@NonNull`, Checker Framework `@NonNull`.
  - JSpecify's scope rule: the write site sits inside a `@NullMarked` class, package or module, with no `@NullUnmarked` in between, and carries no recognised `@Nullable`.
  - Read the same three placements `hasNullableAnnotation` already reads (the element, its type, and for a record the accessor), so a `TYPE_USE` and a declaration annotation both count.
- **Fix line.** Offer, in order:
  - annotate the parameter or component `@Nullable`;
  - declare the domain component without the `Optional`;
  - give it a leaf over the whole `Optional` that encodes absence the wire's way.

  On a class the author cannot edit, only the last two apply.
- **Only the write side.** The parse side reads through `Optional.ofNullable`, so a getter declared non-null is harmless there and needs no check.

## Decision points

- **Error or note.** The repo rule is errors for wrong, notes for redundant, and this is wrong under the declared contract. A warning would fail `-Werror` builds anyway. The cost: a `@NullMarked` DTO package that bridges without `@Nullable` compiled until now, so the release note must say so.
- **Whether an implicit `@NullMarked` scope counts,** or only an explicit non-null annotation. JSpecify's semantics say it counts. Counting only explicit annotations would miss the repro's first two cases, which are the common JSpecify shape.
- **Class-file beans.** A bean from a jar keeps its type-use annotations in the class file, and javac only exposes them on members read from a class file since JDK 22. Confirm on JDK 25 that a `@NullMarked` jar bean is seen, and decide what an unannotated jar bean means (no signal, so accepted).

## Tests the fix should add

- The three repro shapes are refused, each naming the write site and offering `@Nullable`.
- The same shapes with `@Nullable` on the parameter or component compile, and `MappingLaws` passes on an empty `Optional`.
- `@NullUnmarked` on the bean class lifts the refusal.
- Each recognised non-null annotation is refused outside any `@NullMarked` scope.
- A builder setter declared non-null is refused, as a setter is.

## Workaround

Annotate the bridged setter parameter, builder setter parameter or record component `@Nullable`, as the book says, or drop the `Optional` from the domain component.

## Related

- #871 (PR #880), which made every automatic bean bridge write `null` and so widened this from opted-in record components to every bridged bean property.
- #750, type-use annotations in generated code.
- #767, nullable optic foci: a different axis of the same `@NullMarked` story.

MagnusSmith:
Fixes #881.

The `Optional` bridge writes an empty domain `Optional` as `null`: into a record component a spec bridges with `@OptionalBridge`, and, since #871, through the setter or builder setter of every bridged bean property. Nothing asked whether that site could take a `null`. A site declared non-null compiled cleanly, even under `-Werror`, and then either threw from `build` (a setter that checks its argument) or kept a `null` its type rules out.

Reproduced on `d6d791bc8`, before the change; the branch is rebased onto `44203b449` (#888) with the issue's three shapes: an unenforced setter kept `nickname=null`, an enforced one threw `NullPointerException: nickname`, and the record wire built `ProfileDto[name=ada, nickname=null]`.

## What changes

`resolveBridge`, which every bridge passes through, now reads the site the bridge writes into, beside the existing refusal for a getter-only `List`: the record component, or the parameter of the setter or builder setter.

```mermaid
flowchart TD
    S["the site build writes a bridged value into:<br/>record component, setter or builder-setter parameter"] --> N{"reads as nullable?<br/>any @Nullable or @CheckForNull,<br/>or a when other than ALWAYS"}
    N -->|yes| OK["bridges, as before"]
    N -->|no| A{"a recognised<br/>non-null annotation?"}
    A -->|yes| R["refused: build would write<br/>the null its declaration rules out"]
    A -->|no| V{"typed by a type variable<br/>that may stand for a nullable type?"}
    V -->|yes| OK
    V -->|no| M{"inside @NullMarked, with<br/>no @NullUnmarked nearer?"}
    M -->|yes| R
    M -->|no| OK

    class S wire
    class OK ok
    class R bad
    class N,A,V,M decision
```

- **The rule** lives in `NullableAnnotations.nonNullReason`, beside the nullable names the optics generators already use:
  - a site is **nullable** when any annotation on it is named `Nullable` or `CheckForNull`, or carries a `when` other than `ALWAYS` (JSR-305's `@Nonnull(when = MAYBE)`);
  - it is **non-null** when it carries a recognised non-null annotation (JSpecify, JSR-305, Jakarta, JetBrains, AndroidX, SpotBugs, Lombok, Checker Framework), or sits inside a `@NullMarked` module, package, class or method with no `@NullUnmarked` nearer;
  - a **type-variable** site follows its bounds: a plain `<T>` in a marked scope has the non-null bound `Object`, and `<T extends @Nullable Object>` leaves nullness to the argument, so it bridges.
- **Placement** is read the way JSpecify reads it: a declaration annotation on the parameter or component, a `TYPE_USE` annotation on its type, the accessor, and the record's backing field, where a field- and method-targeted `@Nullable` lands when the accessor is declared. `@Nullable String[]` marks the elements and leaves the array non-null.
- **The diagnostic** names the site and why it is non-null, and offers three fixes: mark it `@Nullable` (spelt `String @Nullable []` for an array), declare the domain component without the `Optional`, or add a leaf over the whole `Optional` that encodes absence the way the wire does. A marker or a bridged leaf is named where each fix replaces it, with the mix-in it is inherited from. For a site read from a class file, the `@Nullable` fix says where the class is declared.
- **The refusals that offer a bridge** as their fix, for a pair with no source and for a bridge whose element types differ, now add the `@Nullable` a non-null site needs, so following them no longer leads straight into this one. So does the wrapper a bridge onto a primitive is offered (from #885): inside `@NullMarked`, `Integer` is as non-null as `int`. The element-types refusal stops calling a non-null site nullable.
- `WireShape.WriteSite.Setter` carries the setter's `ExecutableElement` rather than its name, so the parameter can be read. Generated code is unchanged (byte-identical across 28 Impls, probed in review).
- The book's own bridged records sat in the `@NullMarked` examples module without `@Nullable` and were refused. They are marked, and the book pages that include them follow.

## Decisions

- **An error, not a note.** A bridge onto a site declared non-null is wrong under the declared contract, which is the repo's line between an error and a note. It is source-incompatible for a `@NullMarked` mapping whose bridged site lacks `@Nullable`; the release note's _Notes for existing builds_ covers it, including Lombok's `@NonNull` and Kotlin-compiled DTOs, with a path for classes that cannot be edited.
- **Lenient about nullable, exact about non-null.** A refusal needs certainty, so any annotation that reads as nullable counts, whichever library it comes from, while non-null is read only from a fixed list. Jakarta Validation's `@NotNull` is left out: it says what a valid value is, not what the declaration can hold.
- **A class file keeps the `@Nullable` fix.** Compiled is not the same as uneditable: a sibling module's DTO arrives as class files, and a Lombok setter is source the author never wrote. The fix line says where the change is made instead of withholding it.
- **Only declarations are read.** A setter that refuses `null` without declaring it (`requireNonNull`, a protobuf builder outside `@NullMarked`) is not detectable and still throws from `build`; the book's limits table says so.

## Review

Five-lens panel (correctness, type safety, immutability, ergonomics, architecture), with the correctness, type-safety and ergonomics findings reproduced against the built processor. No blockers. Folded in: the two bridge-offering fix lines that led into the refusal (the one major, raised by two lenses and confirmed by probe); type-variable bounds; the backing-field read; `CheckForNull` and `when`; keeping the `@Nullable` fix for class files; the array spelling, nested type names, inherited setter and bridged-leaf wording; helper reuse (`compiledHere`, `componentNamed`, the accessor role labels); the vendor test rebuilt on real `@Target`s; a law test; the #871 release-note entry reconciled with this one; the skill and javadoc examples. Probed and holding: every bridge route reaches the check with one error (marker and marked leaf, bean and builder, build-only, projection, `@Flatten`, nested spec, bridged `List`/`Map` with `@MapKey`, generic spec), nothing is refused where nothing is written (parse-only and private-constructor beans), scope is lexical for inherited setters both ways, module-level `@NullMarked` is seen, a missing annotation class on the classpath does not crash, and every followed fix compiles under `-Werror` and passes `MappingLaws`.

Declined: a leaf fix on a component renamed with `@MapField` cannot be followed, since `@MapField` needs an abstract method. The baseline no-usable-source diagnostic has the same gap, so it is not this change's.

## Behaviour changes

Listed in the release notes: a bridged record component, setter or builder setter declared non-null, by a non-null annotation or by a `@NullMarked` scope without `@Nullable`, compiled before and is now refused.

## Tests

- `MappingProcessorNonNullBridgeTest` (new, 6 tests, about 1.3 s, several specs per compilation):
  - refused under `@NullMarked`: setter, builder setter, build-only bean, record, projection, `@Nullable String[]`, a plain `<T>` record, an inherited generic setter, a bridged leaf; the full message pinned for the bean and the record;
  - class and method scopes, and an explicit `@NonNull`;
  - every recognised non-null annotation at its real `@Target`, on a setter and on a record component, with the list asserted equal to `NON_NULL_ANNOTATION_NAMES`;
  - accepted: `@Nullable` from any library, `@CheckForNull`, `@Nonnull(when = MAYBE)`, an explicit accessor with a field-targeted `@Nullable`, `String @Nullable []`, `@NullUnmarked`, `<T extends @Nullable Object>`, an unmarked `<T>`, a parse-only bean;
  - the two bridge-offering refusals, and the wrapper offered for a primitive, ask for `@Nullable`;
  - a jar bean's refusal says where to mark it.
- `GeneratedMappingLawsTest`: `@Nullable` setter and record component under `@NullMarked` pass `MappingLaws`, absent and present.
- The first cut was mutation-checked: with the check disabled, every refusal test failed.

`:hkj-processor:check` (100% line and branch gate), `:hkj-processor:javadoc`, `:hkj-annotations:check`, `:hkj-annotations:javadoc`, `:hkj-examples:check` (including `bookVerify`) and `spotlessCheck` are green.

## Summary by CodeRabbit

* **New Features**
  * `@OptionalBridge` now validates nullability for record components, setters, and builder setters.
  * Bridges are rejected when empty `Optional` values would be written to declared non-null targets, with diagnostics suggesting `@Nullable` or alternative mappings.
  * Nullable wire targets are supported and documented under JSpecify null-marked scopes.

* **Documentation**
  * Updated mapping guidance, examples, and release notes to clarify empty `Optional` handling and nullability requirements.

* **Tests**
  * Added coverage across records, beans, builders, arrays, annotations, and JSpecify scopes.


MagnusSmith:
@coderabbit review

MagnusSmith:
@coderabbit full review

MagnusSmith:
@coderabbit full review

Issue #893, opened 2026-09-19
[BUG] @GenerateMapping: a sparse UpdateSpec rebuilds the record once per field, so a valid PATCH trips a cross-field invariant


MagnusSmith:
_Names should be confirmed for best ergonomic fit during implementation._

A sparse `UpdateSpec` applies each present PATCH field on its own, rebuilding the domain record through its canonical constructor once per field. A record that enforces a cross-field invariant in its compact constructor therefore sees every intermediate value. A PATCH whose **final** value is valid throws, and the exception escapes `Edits.Accumulated.apply`, which is declared to return a `Validated`.

Found while fixing #872. That fix reports a domain constructor's exception as a located `FieldError` on every surface that builds the record whole from parsed parts: `parse`, the validated `patch`, a flattened group, the fallible merge and `@GenerateAssembly`'s `assemble()`. The sparse tier is left out deliberately, since a plain catch there would turn a valid PATCH into an `Invalid`. Its docs call an invariant spanning the fields a PATCH sets "not supported yet".

Reproduced on `main` at `3ef2b5341`.

## Repro

```java
public record Range(int lo, int hi) {
  public Range {
    if (lo > hi) throw new IllegalArgumentException("lo > hi");
  }
}

public class RangePatch {
  private Integer lo;
  private Integer hi;
  // getters and setters
}

@GenerateMapping
public interface RangePatchMapping extends UpdateSpec<Range, RangePatch> {}
```

```java
RangePatch widen = new RangePatch();
widen.setLo(5);
widen.setHi(10);
RangePatchMappingImpl.INSTANCE.updateFrom(widen).apply(new Range(1, 3));
// throws IllegalArgumentException: lo > hi
// expected: Valid(Range[lo=5, hi=10])
```

Setting `lo` first builds `Range(5, 3)`, which the constructor refuses before `hi` is ever set. A PATCH whose final value is invalid (`lo` 5 alone onto `Range(1, 3)`) throws the same exception, where `Invalid` carrying `lo > hi` is expected.

## Cause

`updateFrom` folds one `Edit.setIfPresent(Setter.fromGetSet(Range::lo, (d, v) -> new Range(v, d.hi())), wire.getLo())` per property (`MappingProcessor.setterExpr`), and `Edits.Accumulated.apply` runs them left to right through `folded.map(update -> update.apply(source))`. Each setter calls the constructor, and nothing catches what it throws.

## Suggested fix

This needs a design decision:

[...]

PR #895, opened 2026-09-19
An element-mapped spec's of(...) takes its own leaves first (#876)

### LINKED ISSUE #876: [BUG] @GenerateMapping: an element-mapped of(...) takes inherited abstract leaves before the spec's own
An element-mapped spec whose abstract leaves come partly from a mix-in generates an `of(...)` whose parameter order puts the inherited leaves first. The book and the processor's own javadoc both say the order is declaration order, with the spec's own members first.

When the leaf types coincide, a positional call compiles and silently swaps the codecs.

## Repro

```java
public record Tagged<N, V>(N name, V value) {}
public record TaggedDto<ND, VD>(ND name, VD value) {}

interface NameLeaf<N, ND> {
  ValidatedPrism<ND, N> name();
}

@GenerateMapping
public interface TaggedMapping<N, ND, V, VD>
    extends NameLeaf<N, ND>, MappingSpec<Tagged<N, V>, TaggedDto<ND, VD>> {
  ValidatedPrism<VD, V> value();
}
```

The generated factory signature is `of(ValidatedPrism<ND, N> name, ValidatedPrism<VD, V> value)`.

Called in the documented order (the spec's own `value`, then the inherited `name`), with every type argument `String`, on `a7427b834`:

```java
TaggedMappingImpl.<String, String, String, String>of(lowerCase, upperCase)
    .parse(new TaggedDto<>("NAME", "value"));
// Invalid(NonEmptyList[name: lower expected, got NAME, value: upper expected, got value])
```

Each codec ran against the other component. With three leaves (one inherited between two own), the signature comes out as `of(second, first, third)`.

## What the book says

- `generics.md`, *Element-mapped specs*: "one `ValidatedPrism` per leaf in declaration order".
- `abstractLeaves` javadoc (`MappingProcessor.java:1374`): "The spec's abstract leaves in declaration order (own members before inherited)".

## Cause

`abstractLeaves` (`MappingProcessor.java:1382`) iterates `specMembers(spec)`, which comes from `Elements.getAllMembers` (`:643`), and that returns inherited members before the spec's own.

## Suggested fix

Order the leaves explicitly:

- the spec's own members, in source order;
- then each inherited interface's members, walking the declared supertypes in order.

That is the order the javadoc already states. The generated `of(...)` javadoc should list the parameter order it emits, so a caller can read it off the Impl.

## Release note

Element-mapped specs shipped in 0.4.10, so this changes a generated signature that is already public. A spec whose leaves are all its own is unaffected; one with an inherited abstract leaf changes parameter order, and a positional call that compiles today may compile afterwards with the arguments in the other order. The release entry should say so plainly.

## Tests the fix should add

- A two-leaf spec with one inherited leaf, pinned by signature and by a runtime call whose leaf types coincide.
- The three-leaf interleaved case.

## Workaround

Declare every abstract leaf on the spec itself, or call `of(...)` with arguments typed so a swap cannot compile.

MagnusSmith:
Closes #876.

## The bug

An element-mapped spec's generated `of(...)` took one `ValidatedPrism` per abstract leaf in the order javac's `Elements.getAllMembers` listed them. That order is a compiler internal: it can put an inherited leaf before the spec's own, and it orders the leaves of several mix-ins arbitrarily. On main, `extends CLeaf, MappingSpec<...>, ALeaf` (where `CLeaf extends DLeaf`) with an own leaf `b()` gave `of(d, a, c, b)`. Where the leaf types coincide, a call written in the documented order compiled and ran each prism against the other component:

```
Invalid(NonEmptyList[name: lower expected, got NAME, value: upper expected, got value])
```

So putting the spec's own members first, as `PathProcessor` does for `@PathVia`, would not have been enough.

## The fix

`specMembers`, the one place every mapping rule reads a spec's members from, now returns them in declaration order:

```mermaid
flowchart LR
    S["DepthFirstMapping<br/>own: b()"] -->|"1st clause"| C["CLeaf<br/>c()"]
    C --> D["DLeaf<br/>d()"]
    S -->|"2nd clause"| M["MappingSpec<br/>(no members)"]
    S -->|"3rd clause"| A["ALeaf<br/>a()"]
    R["of(b, c, d, a)"]
```

- The spec's own members come first, in source order. Then each superinterface follows, in the order the `extends` clause names it, depth first. An interface reached twice counts where it is first reached.
- A leaf that several mix-ins declare is placed, and labelled, by its first declaration. Its type is still the narrowest declaration's.
- The generated `of(...)` now documents each parameter, for example `@param name the {@code name()} leaf, inherited from {@code NameLeaf}`, so the order can be read off the Impl.
- The whole existing suite passed unchanged under the new order, golden files included.
- The classpath index now passes over an entry that names anything but an interface. This closes the one route by which a class could reach the member walk. Without it, a hand-written entry naming a class could crash the comparator.

## Two related defects, fixed here too

**Type-use annotations on type-variable uses were dropped from restated members.**
- `Types.asMemberOf` drops an annotation written on a type variable's use. `ValidatedPrism<String, @Nullable T>` comes back as `ValidatedPrism<String, T>`, even under the declaring type itself. This applied to the spec's own leaves as well as inherited ones.
- The Impl restates leaves (as a field, constructor and `of(...)` parameter, and accessor) and marker stubs. So inside a `@NullMarked` scope, its copy said non-null where the spec said nullable.
- `ProcessorUtils.typeNameOf(type, declared)` now walks the substituted type beside its declaration. It writes such annotations onto whatever replaced the variable: `@Nullable N` under `extends NameLeaf<String>` becomes `@Nullable String`.
- It doesn't write an annotation twice where the replacement already carries it.
- It puts back only an annotation the generated file can write cleanly. One missing from the consumer's compile classpath (a `compileOnly` annotation on a published mix-in), one private in its nesting, one package-private to another package, or a deprecated one is left out as on main, where writing it would fail the build or draw a warning no one can suppress.
- It follows the owner's superinterface clauses as written, so an annotation an intermediate mix-in's clause wrote (`Mid<M> extends NameLeaf<@Nullable M>`) is kept too.
- The one-argument `typeNameOf` delegates to it, so the 30-odd existing call sites produce the same output.
- Swept to the siblings that restate a type read under an instantiation: `@GeneratePathBridge` bridges (parameters, return-type arguments, thrown types, type-variable bounds), `@Flatten` groups, the type witness of a nested generic spec (`PageMappingImpl.<@Nullable String>instance()`, named from the using record's component), and the lens local of a checked `@ImportOptics` traversal.

**A dependency's spec missing part of its hierarchy was called anyway.**
- A spec read from the classpath index is completed only as the processor reads it. If a mix-in (or anything a mix-in extends) was missing from the consumer's classpath, the spec arrived without that mix-in's members.
- The processor then generated a call that javac refused inside the generated source with `cannot access`. An element-mapped spec was even called with too few prisms.
- Such a spec now registers as `Origin.CLASSPATH_MISSING_SUPERTYPE`, beside `CLASSPATH_MISSING_IMPL`, and is never chosen. The use site's error names the missing type:

```
'com.partial.Specs.TaggedMapping (classpath)' maps this pair, but it extends 'com.base.NameLeaf',
which is not on this module's compile classpath, so its members cannot all be read and nothing
can call its Impl: put the module declaring 'com.base.NameLeaf' on this module's compile
classpath.
```

- Reached through a mix-in, the error names that clause instead: `'com.partial.Specs.UserVocabulary' extends 'com.base.MissingBase'`.
- The usual cause is a dependency declaring the mix-in's module with `implementation` rather than `api`; the book's Multi-module builds section now says so.

- The hint that explains an unusable spec used to match by exact type. It now matches through `covers()`, as candidate resolution does, so a generic spec (threaded or element-mapped) is named too. That also improves the existing missing-Impl hint for generic specs.

## Behaviour changes, in the release notes

- **Parameter order.** A spec whose abstract leaves are declared in more than one interface (its own and a mix-in's, or two mix-ins') can get a new `of(...)` parameter order. This was verified with a probe:
  - `of(...)` infers its type arguments, so an old-order call stops compiling only where its result meets a declared type.
  - Through `var`, or where the leaf types coincide, it still compiles and runs each prism against the wrong component.
  - The release notes say to check each call against the parameter list the Impl now documents.
- **Mixed processor versions.** A module nesting such a spec from a dependency works out the leaf order from the spec, so both need the same version. The notes say so. I didn't add code to read the order from the dependency's generated Impl: that only helps mixed versions, and parameter names aren't reliably in class files.
- **Diagnostic order.** Where a spec has several problems, one it declares itself is now reported before an inherited one.

## Tests

- **`of(...)` order:**
  - A runtime test where both leaves are `String` and the two prisms accept disjoint input, so only the right pairing parses.
  - One javac run pins four hierarchies: interleaved, depth-first, diamond, and a leaf two mix-ins declare (position and label).
  - A cross-module test: the spec is read back from its class file, with element types that would not compile in the wrong order.
- **Index:** an entry naming a class is passed over, and the use-site hint does not name it.
- **Type-use annotations:**
  - `ProcessorUtilsTest` pins the member form against a variable bound to itself, to a concrete type, to an already annotated type, to an array and to a wildcard, and against a raw read. It pins which annotations each form keeps: unresolved, private, package-private at home and elsewhere, and public inside a package-private class.
  - The hkj-spring client-processor test compiles a client over a base interface whose annotation class is off its classpath.
  - `TypeUseAnnotationSurvivalTest` gains `@GenerateMapping` cases (own and inherited leaves, a mix-in whose variables the spec binds, a rename stub, a flattened group, an intermediate clause, the nested type witness), an `@ImportOptics` checked-lens case, and a group for annotations a generated file cannot name (kept at home, left off in another package, private). Its `@GeneratePathBridge` case now covers `@Nullable T` and an inherited method.
  - A classpath test compiles a spec whose published mix-in carries an annotation off the consumer's classpath: it builds, and the Impl leaves the annotation out.
- **Missing supertype:** one downstream compile checks a directly missing mix-in on an element-mapped spec and a missing supertype of a mix-in on a concrete spec. Both are named, and neither produces a `cannot access` error.
- **Mutation check:** each new test fails when the fix it pins is reverted.
- **Gate:** `gradle :hkj-processor:clean :hkj-processor:check :hkj-processor-plugins:check :hkj-examples:check :hkj-spring:client-processor:check`. That includes the 100% JaCoCo gate on `MappingProcessor`, `MappingIndexes`, `ProcessorUtils`, `PathProcessor` and the other gated generators, and `bookVerify`.

## Review

A five-lens review panel (correctness, type safety, immutability, ergonomics, architecture) ran over the #876 part. It found no blockers. Its one major was a false claim in the draft release note: that a reordered call "stops compiling", when `var` lets it compile. That was verified and fixed. Its minors are folded in:
- the label of a leaf two mix-ins declare;
- the classpath filter in place of a superclass walk;
- an ordered loop in place of `Stream.forEach`;
- a stale test comment;
- a sharper book paragraph.

The two related defects above came out of the same review, and a second five-lens panel then reviewed them. It found no blockers, and confirmed that no shape with a missing supertype compiled on main. Its majors, all fixed here:
- re-applied annotations that the generated file could not write (missing from the classpath, or not public) broke builds main compiled;
- `@GeneratePathBridge` had the same annotation loss and was unswept;
- the book said the index caveats never touch a vocabulary, and the build-setup page didn't name the `api`/`implementation` cause.

Its minors are folded in too: diagnostic wording, the `@Flatten` sweep, a single read per member, one compile for the `ProcessorUtils` probe, and tighter release notes.

## Follow-ups from the second panel, fixed here too

The branch is rebased onto the latest main as two commits: the PR's change, and one that unifies it with #893. That PR independently added `ProcessorUtils.typeNameOf(type, keep)` and `writableFrom(elements, targetPackage)` for the sparse `Components` record, and also left off deprecated annotations. There is now one rule, and it has #893's deprecation check. The package forms carry it, so the predicate pair is gone and the `Components` record names its types with `typeNameOf(type, implPackage)`.

- **Unnameable annotations anywhere.** An annotation a generated file cannot name used to be copied onto an ordinary type anyway (present on main), failing the build: `Tag is not public in a`, or `package ann does not exist` for a `compileOnly` annotation. `ProcessorUtils` gains `typeNameOf(type, targetPackage)` and `typeVariableOf(parameter, targetPackage)`, which keep an annotation only where that package can name it, and every in-tree generator now passes the package it writes into. The package-free forms remain for the SPI traversal plugins, which are not told the package: they leave off unresolved, private and deprecated annotations and keep a package-private one. The hkj-spring client's copy mirrors that rule.
- **Intermediate clauses.** `typeNameOf(type, declared, owner, targetPackage)` follows the owner's superinterface clauses as written.
- **`@ImportOptics` lens local.** `TraversalHintInfo` carries the lens method's resolved and declared types, and the local is named from the declaration: `Lens<Bag<U>, List<@Nullable U>> lens`.
- **Nested generic type witness.** The witness is named from the using record's component, at the place the use site sits in it (the whole component, a `List` element or a `Map` value), including through an array pattern (`MappingSpec<Bin<T[]>, …>`).

Each fix has a test that fails when the fix is reverted.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved generated code for generic types, nested packages, traversals, lenses, prisms, and other optics.
  - Preserved type-use annotations when they can be safely referenced, while omitting unavailable annotations to prevent compilation failures.
  - Dependency mappings with unavailable supertypes are now skipped with clearer diagnostics identifying the missing type.
  - Invalid classpath entries that do not represent interfaces are ignored.

- **Documentation**
  - Clarified module visibility requirements and generated mapping factory parameter ordering.
  - Added release notes covering these mapping and annotation improvements.


MagnusSmith:
@coderabbit full review

MagnusSmith:
@coderabbit full review

Issue #874, opened 2026-09-18
[BUG] @GenerateMapping: parse constructs the domain through a same-arity overload instead of the canonical constructor


MagnusSmith:
When a domain record has a primitive component and also declares another constructor with the same number of parameters, `parse` can call that other constructor instead of the canonical one. The same happens in the validated `patch`, and in the chunked ladder a domain wider than sixteen components assembles through.

The domain value comes out silently wrong, and `asIso().reverseGet` disagrees with `parse`.

## Repro

```java
public record Money(long cents, String currency) {
  public Money(Number major, String currency) {
    this(Math.round(major.doubleValue() * 100), currency);
  }
}
public record MoneyDto(long cents, String currency) {}

@GenerateMapping
public interface MoneyMapping extends MappingSpec<Money, MoneyDto> {}
```

```java
var dto = new MoneyDto(1234, "GBP");
MoneyMappingImpl.INSTANCE.parse(dto);                // Valid(Money[cents=123400, currency=GBP])
MoneyMappingImpl.INSTANCE.asIso().reverseGet(dto);   // Money[cents=1234, currency=GBP]
```

## Which legs are affected

Verified on `a7427b834`:

| Leg | Result |
|---|---|
| `parse` (up to sixteen components) | `Money[cents=123400]`, wrong |
| `parse` through the chunked ladder (seventeen components) | wrong the same way |
| the validated `patch` on a projection | `Ledger[cents=123400]`, wrong |
| `asIso().reverseGet` | `Money[cents=1234]`, correct |
| a projection's `asLens().set` | `Ledger[cents=1234]`, correct |

The split is the whole diagnosis. The two correct legs read the wire's accessor straight into the constructor call, so a `long` argument keeps the canonical constructor most specific. The fallible legs carry each component through the `Validated` ladder, where it is boxed.

## What the book says

- `beans_patch.md`: "`parse` assembles the domain through its canonical constructor".
- `tiers.md`: the lossless tier's parse-iso coherence law.

## Cause

The ladder's values are boxed (`Long`), and the domain is constructed by overload resolution:

- `.apply($T::new)` — `MappingProcessor.java:7101`;
- `new $T(...)` for a wider domain, through `ChunkedAssembly.java:89` — `MappingProcessor.java:7110` (parse) and `:7351` (patch);
- `new $T(...)` in the patch lambda — `MappingProcessor.java:7365`.

[...]

Issue #872, opened 2026-09-17
[BUG] @GenerateMapping: an exception from the domain constructor escapes `parse`, discarding every other error


MagnusSmith:
A domain record that enforces an invariant in its compact constructor makes `parse` throw instead of returning `Invalid`. The compact constructor is the idiomatic place for a record invariant.

Every error already accumulated is lost. At the Spring boundary the request becomes a 500, not the 422 the chapter builds towards.

`@GenerateMerge`'s fallible merge constructs its target the same way, and behaves the same.

Reproduced on `main` at `62b34a1ea`.

## Repro: parse

```java
public record Range(int lo, int hi) {
  public Range {
    if (lo > hi) throw new IllegalArgumentException("lo > hi");
  }
}
public record RangeDto(int lo, int hi) {}

public record Schedule(String name, List<Range> ranges) {}
public record ScheduleDto(String name, List<RangeDto> ranges) {}

@GenerateMapping public interface RangeMapping extends MappingSpec<Range, RangeDto> {}
@GenerateMapping public interface ScheduleMapping extends MappingSpec<Schedule, ScheduleDto> {}
```

```java
ScheduleMappingImpl.INSTANCE.parse(
    new ScheduleDto(null, List.of(new RangeDto(1, 2), new RangeDto(5, 1))));
// throws IllegalArgumentException: lo > hi
```

Expected instead: `Invalid` carrying `name: must not be null`, and a failure located at `ranges.1`. The `name` error is accumulated before the exception is thrown, and is lost with it.

## Repro: merge

```java
public record Src1(int lo, String id) {}
public record Src2(int hi) {}
public record Target(int lo, int hi, UUID id) {
  public Target {
    if (lo > hi) throw new IllegalArgumentException("lo > hi");
  }
}

@GenerateMerge
public interface Assemble {
  Validated<NonEmptyList<FieldError>, Target> merge(Src1 first, Src2 second);

  default ValidatedPrism<String, UUID> id() { return StandardCodecs.uuid(); }
}
```

```java
AssembleImpl.INSTANCE.merge(new Src1(5, "123e4567-…"), new Src2(1));
// throws IllegalArgumentException: lo > hi
```

## What the book says

- `basics.md`: "a `parse` that reports every bad field at once", and "Note what the failure looks like: a value, not an exception".
- `ch_intro.md`: "the fallible one reports **every** bad field at once".

[...]

PR #896, opened 2026-09-19
A sparse UpdateSpec constructs the domain record once (#893)

### LINKED ISSUE #893: [BUG] @GenerateMapping: a sparse UpdateSpec rebuilds the record once per field, so a valid PATCH trips a cross-field invariant

MagnusSmith:
Fixes #893.

A sparse `UpdateSpec` folded one `Setter` per present field, and each one called the domain record's canonical constructor. A record whose compact constructor checks its fields against each other (`if (lo > hi) throw ...`) therefore saw every value in between: a PATCH moving both ends of a range built `Range(5, 3)` on the way and threw, although the range it ends on is valid, and the exception escaped `apply`, which is declared to return a `Validated`.

Reproduced on `3ef2b5341` before the change: the valid widening (`lo` 5, `hi` 10 onto `Range(1, 3)`) and the invalid `lo` 5 alone both threw `lo > hi`.

## What changes

The issue's option 1, construct once:

```java
RangePatchMappingImpl.INSTANCE.updateFrom(widen).apply(new Range(1, 3));
// before: throws IllegalArgumentException: lo > hi
// after:  Valid(Range[lo=5, hi=10])

RangePatchMappingImpl.INSTANCE.updateFrom(loOnly).apply(new Range(1, 3));
// before: throws IllegalArgumentException: lo > hi
// after:  Invalid(NonEmptyList[lo > hi])
```

```mermaid
flowchart LR
    subgraph before["before: one Setter per field"]
        direction LR
        A1(["Range(1, 3)"]) -->|"lo = 5"| A2(["Range(5, 3)<br/>the constructor throws"])
    end
    subgraph after["after: onto Components, constructed once"]
        direction LR
        B1(["Components(1, 3)"]) -->|"lo = 5"| B2(["Components(5, 3)"]) -->|"hi = 10"| B3(["Components(5, 10)"]) -->|"set once"| B4(["Range(5, 10)"])
    end

    class A1,B1,B2,B3,B4 ok
    class A2 bad
```

- **Core, `Edits.accumulate(Lens<S, A> focus, edits...)`** (and a `List` overload): the edits validate exactly as `accumulate` does, write onto `focus.get(source)`, and `focus.set` runs once. A `RuntimeException` from that set is the constructor refusing the final values and becomes an unlabelled `FieldError` carrying its message; one without a message, or with a blank one, reads `not a valid Range`, as #872's guard does. Only the set is caught: an edit's own function and `focus.get` still throw. When every edit is absent the source comes back as it is, with no set. A hand-written PATCH can use it directly.
- **Generated `updateFrom`:** writes onto a private `@Generated record Components` holding only the components the PATCH can set, and the lens's setter reads every other component from the current value:
  ```java
  return Edits.<Range, Components>accumulate(
          Lens.of(
              d -> new Components(d.lo(), d.hi()),
              (d, c) -> new Range(c.lo(), c.hi())),
          Edit.setIfPresent(Setter.fromGetSet(Components::lo, (c, v) -> new Components(v, c.hi())), wire.getLo()),
          Edit.setIfPresent(Setter.fromGetSet(Components::hi, (c, v) -> new Components(c.lo(), v)), wire.getHi()));
  ```
- **`toValidated()`** hands back the construct-once `Update`. It has no error channel, so a refused final value throws from it, as it did before; a valid one no longer does. `basics.md` now lists it beside `asIso().reverseGet` and `asLens().set` as the surfaces that propagate.
- **Side effects:**
  - **Compile time:** the explicit `<Range, Components>` witness gives every edit its target type up front. A 64-property PATCH compiled in about 15 seconds on main, and now in under one; 200 properties take 1.3 seconds.
  - **#874 on this tier:** the constructor now receives `c.x()` with the component's own type, so the canonical constructor is chosen, where a same-arity overload taking a wrapper's supertype won before (pinned by a test).

## Design notes

Two shapes were tried and dropped, both by probe:
- **An `Iso` over every component.** Writing out a component the PATCH does not set is new: main only inferred it. A deprecated or `forRemoval` component type then fails under `-Werror`, and a package-private one from another package does not compile. Carrying only the covered components, with the rest read from the domain, is what makes the relationship a `Lens`.
- **A generic `Components<T1..Tn>`.** javac checks inferred diamond and lambda types for access and deprecation too, so it avoided nothing, and its inference had not finished after minutes at 64 fields.

A covered component's type was already inferred by main's setters, so naming it adds nothing that main did not already warn on or refuse (checked for deprecated and inaccessible types).

The API name, `Edits.accumulate(focus, ...)`, is new in this SNAPSHOT; the ergonomics lens recommended keeping it as an `accumulate` overload over a separate name or an instance method on `Accumulated`.

## Review

Five-lens panel (correctness, type safety, immutability, ergonomics, architecture); correctness and type safety reproduced every claim against the built processor. All findings were verified.

**Folded in:**
- **Major, correctness and immutability:** a PATCH that sends nothing ran the constructor, so it returned a new instance, broke the sparse identity law for a constructor that clones an array, and could change a stamping constructor's value. Edits that hand back the focus they read now return the source untouched, which restores main exactly.
- **Major, type safety:** the `Components` record copied the covered components' type-use annotations, so one missing from the consumer's classpath, package-private in another package, or deprecated broke a build main compiled. `ProcessorUtils.typeNameOf` gained an overload with an annotation filter, and `writableFrom` keeps only what the Impl's package can write cleanly, at every depth; a `@Nullable` survives.
- **Major, ergonomics:** `MappingLaws`' sparse validation law requires located errors, and a constructor's refusal is unlabelled, so a domain with no leaf cannot pass an invalid wire. The javadoc, book and skills say so, and the book shows `assertSparseIdentity` and `assertSparseIdempotent` used on their own.
- **Minor:**
  - the guard wrapped `Validated.validNel(...)` too, so it now catches the set alone;
  - a spec nesting a type named `Lens` shadowed the new import (the skeleton's `avoidClashesWithNestedClasses`, from #894, covers it; pinned by a test);
  - an anonymous source class's fallback names its binary name;
  - the setter lambda's parameter is `c`, as the lens's is;
  - stale `Edits` javadoc, a Coupled Fields table row, the multi-edit page's links into the new section, and fuller release notes.
- **#872 reconciliation:** the notes #872 added saying the sparse tier does not support an invariant spanning the fields it sets are gone, from basics.md, beans_patch.md, the `UpdateSpec` and `GenerateMapping` javadoc, the hkj-mapping skill and the #872 release entry.

**Declined:**
- `Edits.accumulate(null, edit)` is now ambiguous: every such call threw a `NullPointerException` before.
- A wire mutated after `updateFrom` validated it now surfaces as a root `Invalid` rather than an NPE. The aliasing is unchanged from main, and catching any `RuntimeException` is #872's policy.

**Not in scope, unchanged from main:** a spec nesting a type named `Edits` does not compile, because javapoet imports the top level of `Edits.Accumulated` by the nested name.

## Behaviour change

For code built against 0.4.10, where the domain constructor refuses the values a PATCH ends on:
- `apply` and `applyPath` return an `Invalid` rather than throwing;
- any `RuntimeException` counts, so a constructor bug reaches the client as its message, and at the Spring boundary a 500 becomes a 400 or 422.

A PATCH that sends nothing still returns the same instance, without running the constructor.

## Testing

- `EditsTest$AccumulateOntoFocusTests` (14): construct once through an invalid midpoint, the refusal at the root, the fallback names (a record, an anonymous class), absent edits handing back the source with no set, errors alone when an edit fails, only the set caught (an edit bug, a `get` bug, a `null` from the set), many sources, `applyPath`, `toValidated`, the list overload, nulls.
- `MappingProcessorUpdateTest$ConstructOnce` (10), one shared `-Xlint:all -Werror` compilation probed through a `Probes` class: the valid final value, the refused one, the fallback, a record with no check, an uncovered component whose type the Impl cannot name, the canonical constructor over a same-arity overload, the writable annotations, spec-nested `Lens`/`Components`/`Setter`, and a two-stage compile with an annotation missing from the classpath.
- A `RangePatchMappingImpl$Components` row in `GeneratedMarkerReachesNestedTypesTest`, two regenerated goldens, and seven fragment assertions.
- The book's `PriceBand` example and its law check in `RecordMappingBookLawsTest`, and the `Range`/`Ends` example on the multi-edit page, both compiled.
- `gradle spotlessCheck :hkj-core:test :hkj-core:jacocoTestReport :hkj-annotations:check :hkj-test:check :hkj-processor:check :hkj-examples:check :hkj-spring:example:compileJava`, clean, on the rebased branch. JaCoCo is 100% on `Edits`; the only hkj-core misses, `MonadError.recoverWith`, are the same on `origin/main`.

## Summary by CodeRabbit

- **New Features**
  - Added focus-based accumulation for applying multiple validated edits and setting the result back once.
  - Sparse PATCH updates now construct domain records once from final values, allowing cross-field invariants to validate complete updates.
  - Constructor refusals are returned as validation errors, while empty PATCHes preserve the current value without reconstruction.

- **Documentation**
  - Expanded guidance, examples, API references, and release notes for invariant-safe updates and focus-based edits.
  - Clarified validation laws and error-handling behavior for sparse updates.


MagnusSmith:
@coderabbit full review

MagnusSmith:
@CodeRabbit full review

PR #898, opened 2026-09-19
The fields() builders end in construct for a constructor that may refuse (#897)

### LINKED ISSUE #897: [DESIGN] fields(): a guarded construct terminal, decided before the 0.4.11 cut

**Decide before the 0.4.11 cut, or drop.** This proposes new public API on unreleased (SNAPSHOT) stages; once 0.4.11 ships, it is permanent.

## The gap

#872 (PR #894) made the generated `parse`, validated `patch`, fallible merge and `@GenerateAssembly`'s `assemble()` report an exception from the record's canonical constructor as a `FieldError` at the record's path, instead of letting it escape and discard every accumulated error. #893 (PR #896) did the same for the sparse tier through a public API, `Edits.accumulate(Lens focus, edits...)`.

The hand-written builder is now the one assembly surface without that guard. `apply(...)` runs whatever function it is handed, so a compact constructor that refuses the combination still throws:

```java
Validated<NonEmptyList<FieldError>, Range> range =
    Validated.fields()
        .field("lo", parseInt(dto.lo()))
        .field("hi", parseInt(dto.hi()))
        .apply(Range::new);                 // throws IllegalArgumentException: lo > hi
```

The book documents the difference in `validated_assembly.md`: the builder and a `@GenerateAssembly` companion agree on every input the constructor accepts, and part where it throws.

## Proposal

A guarded terminal on the labelled `fields()` stages, beside `apply`:

```java
Validated.fields()
    .field("lo", parseInt(dto.lo()))
    .field("hi", parseInt(dto.hi()))
    .construct(Range::new, "not a valid Range");
// Invalid(NonEmptyList[lo > hi]), unlabelled, so an outer field(label, ...) locates it
```

It would have the same contract as the generated guard:
- the function runs only once every field is valid;
- a `RuntimeException` it throws becomes an unlabelled `FieldError` carrying the exception's message, or the fallback when the message is missing or blank;
- only the function sits inside the catch.

The stages are generated by `AccumulatorStepGenerator` for arities 1 to 16, so this is one generator change: a method per arity on each labelled family.

## What it would buy

- **Hand-written builders get the guard** as an opt-in, and the documented builder-versus-companion difference closes for anyone who uses it.
- **Simpler generated code.** The mapping processor could emit `.construct(Range::new, "not a valid Range")` where it now emits a constructor thunk wrapped in a generated `hkj$construct` helper. That drops the thunk and the `.<Group>field(...)` type witness a `@Flatten` group needs, and it drops the lambda-parameter naming that `parse`, the `@Flatten` group and the fallible merge need. That naming caused the one major regression #894's review found: a group member named `c1` collided with a chunk local.
- **Not everything goes.** The validated `patch` still needs lambdas, because it reads the unprojected components outside the guard. A ladder wider than 16 legs combines chunks with `ap`, which has no stage terminal to call, so it would keep the generated helper or need a public static counterpart.

## Decisions needed

1. **Go or drop.** Dropping keeps today's behaviour and the documented difference; nothing else changes.
2. **Name.** `construct` is a placeholder; `applyGuarded` and `applyChecked` are alternatives. It has to read well beside `apply`.
3. **The fallback message.** A method reference cannot name its type, so either the caller passes it (as above), the API takes the class (`construct(Range.class, Range::new)`), or a fixed default is used (`"not a valid value"`). `Edits.accumulate(Lens, ...)` names the source's runtime class, which a failed construction does not have.
4. **Which families.** Only the labelled ones carry `FieldError`: `ValidatedFieldsN`, `ValidationPathFieldsN` and `EitherOrBothFieldsN`. The generic `accumulate()` families would need an error-mapping function instead of a message, or stay out. For `EitherOrBoth`, what a refusal does to accumulated warnings (`Both`) needs its own answer.
5. **Whether the generator adopts it.** The generator can adopt it for the single-ladder surfaces. The chunked form then needs either the generated helper or a public static beside the terminal.

## If it goes ahead

- Tests: every arity refuses with the message, and falls back on a missing or blank one; an exception thrown by a leg's own `Validated` is not caught; a nested refusal is located by the outer label.
- The book's `validated_assembly.md`, the hkj-guide and hkj-mapping skills, and the release notes.
- If the generator adopts it, regenerate the goldens; `ConstructorInvariantTest` pins the behaviour, which must not change.

MagnusSmith:
Fixes #897.

`apply` runs whatever function it is handed, so a record whose compact constructor checks its fields against each other made the hand-written builder throw:

```java
Validated.fields()
    .field("opens", parseDate(raw.opens()))
    .field("closes", parseDate(raw.closes()))
    .apply(Window::new);          // throws IllegalArgumentException: closes must be after opens
```

That was the one assembly surface without the guard the generated mappings got in #872 (PR #894) and the sparse PATCH tier in #893 (PR #896), which is the difference `validated_assembly.md` had to document.

## What changes

Every labelled builder also ends in `construct(f, fallbackMessage)`:

```java
Validated.fields()
    .field("opens", parseDate(raw.opens()))
    .field("closes", parseDate(raw.closes()))
    .construct(Window::new, "not a valid Window");
// Invalid(NonEmptyList[closes must be after opens])
```

- **The contract**, the generated mappings' own: `f` runs only once every field is valid, so a record reports either its fields' errors or its refusal, never both. A `RuntimeException` it throws becomes an unlabelled `FieldError` carrying the message, or `fallbackMessage` when the message is missing or blank, so an enclosing `field(label, ...)` locates it (`window: closes must be after opens`).
- **Only `f` is guarded.** A `null` it returns throws, exactly as it does from `apply`, and a blank `fallbackMessage` is rejected.
- **All three carriers**, arities 1 to 16, from one template in `AccumulatorStepGenerator`: `Validated.fields()`, `Path.fields()` and `EitherOrBoth.fields()`. On the tolerant carrier a refusal comes back as a `Left` keeping the warnings collected so far, since `flatMap(NonEmptyList.semigroup(), ...)` already combines them.
- **Out of scope by decision:** the generic `accumulate()` builders have no `FieldError` to build, so they end in `apply` only and their generated code is byte-identical; the mapping processor keeps its own `hkj$construct` helper, since #874's fix will need exactly-typed constructor arguments.

## Decisions

Settled on the issue before implementation: go ahead; the name `construct`; a required `String` fallback, since a constructor reference cannot name its type and the message is client-facing; the three labelled families only; and the generator keeps its helper.

## Review

Five-lens panel (correctness, type safety, immutability, ergonomics, architecture), each reproducing its claims against the built jars.

**Folded in:**
- **Major, found independently by three lenses and reproduced:** the carrier's wrapping call sat inside the guard, so a function returning `null` was caught and reported as a refusal carrying a library-internal message (`Validated.construction value cannot be null`), where `apply` throws; at a Spring boundary that string reached the client. The call is now hoisted out of the `try`, which also makes the emitted "Only `f` runs inside the guard" true. `Edits.setOnce` from #896 already kept its wrap outside; the mapping processor's helper is safe only because its thunk is a constructor call that cannot return `null`.
- **Minor:** a blank `fallbackMessage` was accepted and produced an unreadable empty error at the one point where nothing else can; it now throws `IllegalArgumentException`.
- **Docs:** the motivating sentence was false on the strict carriers, where `apply` never runs the function once a field has failed, so nothing is lost; it now says an enclosing assembly loses what it accumulated around it, and the tolerant carrier its warnings. Also the mapping skill's "`ValidatedFields16` offers only `apply(...)`", the `@return` wording, a warning that a bug in `f` reaches the caller as its message, `apply`'s own javadoc pointing at `construct`, the validated package overview, and three sibling pages.
- **Style:** the generated `ValidationPath` body now indents under its bind like the other two carriers, derived from the bind so they cannot drift; test helpers moved to each file's helper block.

**Checked, no change:** stage immutability and repeated completion; `f` evaluated exactly once and never while a field is invalid; `Error` and sneaky-thrown checked exceptions propagate; declaration order at arity 16 on all three carriers; warning order on `EitherOrBoth`; `apply`'s signature and behaviour unchanged, and no existing call site affected; inference from constructor references at every arity, nested in `field(...)` with no witness, with overloaded constructors and generic records; the generated code clean under `-Xlint:all -Werror`; no ArchUnit rule affected.

## Tests

- 21 cases across the three carriers (`ValidatedAssemblyTest`, `ValidationPathAssemblyTest`, `EitherOrBothAssemblyTest`): the arity shapes including the 16 ceiling, a refusal and an accepted value, the missing and blank message, a field that stops the function, an `Error` passing through, nesting located by the outer label, a function returning `null`, and the null and blank argument guards. On `EitherOrBoth`, warnings kept on refusal and carried beside an accepted value.
- `AssemblyPropertyTest.constructAgreesWithApply`: `construct` equals `apply` on every input the function accepts, and reports the refusal on the rest.
- `AccumulatorGeneratorTest`: the structural shape per carrier, and that the `Accum` families do not gain it.
- `ValidatedAssemblyBookTest` pins the errors behind the result the book's new example prints.

Green: `:hkj-processor:check` (100% gate) and `:javadoc`, `:hkj-core:check` and `:javadoc`, `:hkj-examples:check` including `bookVerify`, the Spring example modules, and `spotlessCheck`.

## Notes

Purely additive on final generated classes: no existing signature changes, so nothing needs a note for existing builds.

## Summary by CodeRabbit

* **New Features**
  * Added `construct(...)` to labelled field builders for safely assembling records whose constructors may reject valid fields.
  * Constructor runtime exceptions are returned as unlabelled field errors, using the exception message or a fallback message.
  * Accumulated warnings are preserved for tolerant validation results.
  * Existing `apply(...)` behavior remains unchanged.

* **Documentation**
  * Updated guides, API documentation, examples, and release notes with usage details and comparisons between `apply` and `construct`.


MagnusSmith:
@coderabbit full review

PR #901, opened 2026-09-20
A @Wither lens is checked against the method its call binds, and every name a strategy carries with it (#887)

### LINKED ISSUE #887: [BUG] A generated wither call can bind to an overload the checks never looked at

The lens `@ImportOptics` generates for a wither calls it as `source.withX(newValue)`, and javac chooses among the `withX` overloads by the type of `newValue`. That need not be the overload that was checked. #878 held a wither's return to the source type on both routes, and the pairing rule already holds its parameter to the getter. But each check reads one method, or any method of that name. Neither reads the one the call binds to, so an overload that returns something else can still be the one called. On the spec route, a `@Wither` naming no applicable method at all is reported only by javac, inside the generated file.

## Measured

On `fix-878-wither-admission` (#878), javac-direct on JDK 25.0.1 with `-Xlint:all,-processing,-cast -Werror`. Each fixture is compiled from source beside its import.

| Route | Class | Result |
| --- | --- | --- |
| class literal | `WS<T>`: `int n()`, `WS<T> withN(int)`, and a raw `WS withN(Integer)` | 2 `[unchecked]` in `WSLenses` |
| class literal | `WX<T>`: `int n()`, `WX<T> withN(int)`, and `Object withN(Integer)` | `incompatible types` in `WXLenses` |
| spec, `OpticsSpec<WR<T>>` | a raw `WR withId(String)` beside `WR<T> withId(Object)` | 2 `[unchecked]` in `WROptics` |
| spec, `OpticsSpec<WO<T>>` | `WO<String> withId(String)` beside `WO<T> withId(Integer)` | `incompatible types` in `WOOptics` |
| spec, `OpticsSpec<WE>` | `@Wither("withIdentifier")`, which `WE` does not declare | `cannot find symbol` in `WEOptics` |

```
WSLenses.java:23: warning: [unchecked] unchecked method invocation: method of in interface Lens is applied to given types
    return Lens.of(WS<T>::n, (source, newValue) -> source.withN(newValue));
WSLenses.java:23: warning: [unchecked] unchecked conversion
WXLenses.java:23: error: incompatible types: cannot infer type-variable(s) S,A
    return Lens.of(WX<T>::n, (source, newValue) -> source.withN(newValue));
WEOptics.java:27: error: cannot find symbol
      (source, newValue) -> source.withIdentifier(newValue));
```

In each row the checked method is sound: `withN(int)` pairs with `int n()` and returns the class, and one `withId` in each spec row returns the source type. The call binds to the other overload. The first four rows behave the same on `main`, before #878. There the fifth also reports only javac's error.

## Why

- **Class-literal route.** `TypeKindAnalyser.detectWitherMethods` pairs `withN(int)` with `int n()`. The lens field type is boxed (`Lens<S, Integer>`), so `newValue` is an `Integer`, and phase-1 overload resolution prefers `withN(Integer)`, which needs no unboxing.
- **Spec route.** `SpecInterfaceAnalyser.rebuildsThroughWitherOfAnotherType` accepts when any reachable one-parameter instance method of the name returns the source type. It does not ask which of them `newValue`, typed by the lens focus, would bind to. It leaves a name with no such method to javac. That was a decision in #878, taken to keep that change to the return.

## What "good" looks like

The generated call binds to a method that was checked. Where no method applies, or the one that applies hands back another type, the refusal comes at the declaration, naming the method, not as a javac error inside a generated file.

## Things to explore

- **Resolve the overload the call binds to.** Pick among the applicable methods for the focus type, `Types.isAssignable(focus, parameter)`, and check that one's return. On the class-literal route, either pass the unboxed value (a cast to the primitive in the generated lambda), or skip the pairing when a boxed overload applies too.
- **Existence and parameter on the spec route.** Refuse a `@Wither` whose name no one-parameter instance method carries, or none of whose methods takes the focus type, offering a near name where there is one.
- Keep the retag rule of #878 (`<U> Draft<U> withId(String)` is inferred back to the source type) under whatever resolution is chosen.

## Priority

Proposed **P3**. It needs overloaded withers whose returns differ, or a misspelt `@Wither` name. It fails at build time, in a generated file, never at runtime.

## Dependencies / scope

- In: `TypeKindAnalyser.detectWitherMethods`, `ExternalLensGenerator.createWitherLensMethod` (the generated call), and `SpecInterfaceAnalyser.rebuildsThroughWitherOfAnotherType`.
- Depends on: #878, whose return check this extends.

MagnusSmith:
Fixes #887.

A generated wither lens sets through `source.withX(newValue)`, and where `withX` is overloaded javac picks the method by the type of `newValue`, the lens's focus. Neither route's checks read that method, so an overload none of them looked at could be the one called. Two gaps the same work left are folded in: a field whose accessor is spelt more than one way, and the other method names a copy strategy carries.

## The wither call binds the method that was checked

**Imported by class literal.** `withN(int)` pairs with `int n()`, but the lens hands its setter an `Integer`, and javac's first phase prefers any overload that takes it without unboxing: `withN(Integer)`, an inherited `withN(Number)`, a `withN(Object)`, a generic `<V> withN(V)`. Where that overload returned something else, the generated file did not compile, or warned `[unchecked]`; where it returned the class, it compiled cleanly and the lens rebuilt through the wrong method, keeping none of the fields that overload was not passed. Where the name is overloaded, the lens now passes a primitive parameter unboxed, `source.withN((int) newValue)`, which binds the method that paired. A wither with no such overload is generated exactly as before.

**Imported by spec interface.** `@Wither` admitted any reachable one-parameter method of the name that returned the source type, which need not be the one the call binds. A new `WitherBinding` repeats javac's choice for a one-argument call: the strict phase, then the loose one, then the most specific parameter, with one signature inherited twice resolved to its most specific return. Where the choice rests on inference it is left undecided and read leniently, so nothing javac would call is refused: a parameter naming the method's own type variable, the variable-arity phase, and a focus written as a wildcard, which such a lens infers from its getter as much as from the wildcard.

The method the call binds is then checked, and these are refused at the spec method rather than inside a generated file:

| The declaration | What it now says |
| --- | --- |
| a name the source type does not have | `'Ticket' has no method 'withIdentifier' for the generated lens to call`, listing the withers it does have with the value each takes, and offering the nearest |
| a name it declares out of reach | the same, saying it is declared where the generated class cannot call it |
| a focus no overload takes | `No method 'withCount' of 'Tally' takes the lens's focus type 'String'` |
| more than one overload taking it | `The generated call to 'withId' cannot choose between 'withId(Serializable)' and 'withId(CharSequence)'` |
| a static method | `'withId(CharSequence)' is static, so the generated lens cannot rebuild a 'Stamp' through it` |
| a return that is not the source type | unchanged, but now naming the overload: `'withId(String)' returns 'Draft', not the source type 'Draft<T>'` |

## One lens per field name

A class can spell one field's accessor more than one way and carry a wither for each: `int n()` with `withN(int)`, and `String getN()` with `withN(String)`. Both paired with the field `n`, the companion declared the lens method twice, and it did not compile. The pairing rule now tries the spellings in the order it states, the field's own name, then `getXxx`, then `isXxx`, rather than in the order the class declares its members, so the same class always yields the same lens. The pair left out is reported in a note naming both, since the class is someone else's to spell, and a note does not fail a `-Werror` build.

## Every method name a strategy carries

A misspelt name on a spec's copy strategy reached javac as `cannot find symbol` inside the generated file. Each is now checked at the spec method: the getter every strategy reads through, which for `@ViaCopyAndSet` and `@ViaConstructor` is the lens method's own name; `@ViaBuilder`'s `toBuilder`, `setter` and `build`, each read on what the step before it hands back; and each accessor a `@ViaConstructor` `parameterOrder` names, which has to include the lens's own, or the rebuild would set nothing.

An accessor that reads a type the lens cannot hand back as its focus is refused with both types named, which is the `withMonth(int)` beside `getMonth()` pairing `LocalDate` is known for. A builder's setter is resolved as a wither's is, so a name none of its overloads takes, one no overload of which takes a single argument, an ambiguous call and a static method are each refused there too, as is a builder, or a type its setter hands back, that the generated class cannot name.

Every refusal offers a remedy that can be followed: one that names only the strategies other than the one being refused, the boxed type where an accessor reads a primitive, both types in full where their simple names collide, and no focus at all where the accessor reads nothing, or a type only a wildcard stands for.

## Behaviour changes

A spec `@Wither`, or a `setter` a strategy names, whose call binds a static method compiled before, with a `[static]` warning under `-Xlint`, and is now refused: a static method never reads the value it is called on. On a class imported by class literal, a lens for an overloaded primitive wither now calls the primitive overload, so a type whose boxed overload did something else behaves differently, and setting `null` through such a lens throws where the boxed overload used to take it.

The unit tests' own fixtures declared `@ViaBuilder` on a record with no builder, and a `@Wither` getter that was a public field; both are shapes the processor now refuses, so the fixtures declare what they name. The two annotation javadocs whose examples named a lens method after no accessor are corrected the same way.

## Docs

Nine new entries in [Compiler Errors](hkj-book/src/optics/compiler_errors.md), eight of them gated by `bookVerify` against the real diagnostic, with the triage table and the `@Wither` entries' reasons updated. The pairing rule in Importing Optics, the `@Wither` section of Copy Strategies, the `hkj-optics` skill and the release history follow.

## Verification

`:hkj-processor:check` green, including the per-class 100% JaCoCo gates, which `WitherBinding` now joins. `:hkj-examples:bookVerify` green, with both snippet floors raised. Every shape above was also measured javac-direct on JDK 25.0.1 under `-Xlint:all -Werror`, before and after.

## Known limits

A raw accessor return still lands an `[unchecked]` warning in generated code; `parameterOrder` is not held to the constructor's arity; a field with three spellings draws one note per pair left out rather than one for the field. None is a regression, and each would risk refusing something javac compiles.


MagnusSmith:
@coderabbit  full review

